PR #142, opened 2016-11-20
Like Feature

### LINKED ISSUE #139: Like button required features
- should be refactored into redux
- should show the list of users who liked after clicking on the number of likes
- should optimistically update like numbers after clicking on like
- should re-fetch the data of post to update earnings

Mock up by @adcpm 
pasted_image_at_2016_11_17_16_29

p0o:
should fix #139


adcpm:
Sometime when i click on like icon the button act like the vote hasn't been saved (but i was)
like

adcpm:
Can we know about downvote on the Like list?

p0o:
@adcpm Please check if the reported issues still exist or not. Also it should work fine on single page without list and extra info (that's for another PR)

adcpm:
The previous issue has been solved. There is a new issue with unlike. When i unlike a post and refresh the page, i still see the like button active. 
unlike

adcpm:
Also when 0 like or dislike the tab should not be clickable, comments too.

adcpm:
Oops. When doing "unlike" its actually making a downvote
image

adcpm:
There is an issue for me, the vote now are just incremented and not keep status see vote

p0o:
@adcpm reported errors should be fixed now.

adcpm:
Like button work fine now :). I saw this issue about the tabs:
- If i click on 0 Likes, look like the tab will open empty (it creating 2pixel difference on UI).
- Comment box should not be openable when 0 comment.
- When there is active vote/downvote the tab should be visible (atm if i click on -1 vote its not opening and i cant see the downvote)

p0o:
The problem here is with posts older than payout. Blockchain API is not returning `active_votes` for these posts so that's the reason of inconsistency between the number of likes and its list. We need to check and if the `active_votes` are not loaded we should load them manually. I'm adding that implementation to this PR soon.

adcpm:
PR look's good using Steemit WS. Well done!

PR #146, opened 2016-11-23
Add search in sidebar tab categories and users

adcpm:


src/app/Sidebar/SidebarUsers.js
```diff
@@ -54,8 +67,12 @@ export default class SidebarMessages extends Component {
       );
     });
 
-    let contacts = this.props.contacts.slice(0, 20);
-    contacts = map(filter(contacts, (contact) => {
+    const search = this.state.search;
+    let users = filter(this.props.contacts, (user) => {
```
p0o: filter is available in javascript why requiring additional lib? `this.props.contacts.filter(cb)`

src/app/Sidebar/SidebarUsers.js
```diff
@@ -3,30 +3,43 @@ import React, { Component, PropTypes } from 'react';
...
+    this.state = {
+      search: '',
+    };
+  }
+
+  search = (e) => {
+    this.setState({ search: e.target.value });
+  };
+
   render() {
```
p0o: So many calculations are happening in the render() function. It's better to isolate calculations into different pure functions (could be defined inside the scope of react class or in the top of file but better not in a separate file) with meaningful names so with a quick look into render functions we understand the stream of actions taking place.

Functions could be defined with verbose names like `usersFilteredBySearchQuery()` or in the similar ways to make it clean and understandable.

src/app/Sidebar/SidebarUsers.js
```diff
@@ -54,8 +67,12 @@ export default class SidebarMessages extends Component {
...
     });
 
-    let contacts = this.props.contacts.slice(0, 20);
-    contacts = map(filter(contacts, (contact) => {
+    const search = this.state.search;
+    let users = filter(this.props.contacts, (user) => {
+      return startsWith(user, search);
+    });
+    users = users.slice(0, 20);
+    users = map(filter(users, (contact) => {
```
p0o: map exist in javascript arrays too

src/app/Sidebar/SidebarUsers.js
```diff
@@ -74,13 +91,25 @@ export default class SidebarMessages extends Component {
     ));
 
     return (
-      <div className="SidebarMessages">
+      <div className="SidebarUsers">
         <ul className="tags">
+          <li className="search">
+            <div className="input-group">
```
p0o: classNames are not based on BEM

src/app/Sidebar/SidebarUsers.js
```diff
@@ -74,13 +91,25 @@ export default class SidebarMessages extends Component {
...
+                onChange={this.search}
+              />
+            </div>
+          </li>
           {unreadMessages.length ? (
             unreadMessages
           ) : null}
-          {contacts.length ? (
-            contacts
+          {users.length ? (
```
p0o: Much cleaner to have the braces in the same row number to improve readability, also I prefer this style for single line if statements:
```
{ contacts.length
    ? ( contacts )
    : null
}

PR #148, opened 2016-11-25
Login/Splash page

adcpm:


src/wrapper.js
```diff
@@ -1,57 +1,63 @@
...
       <IntlProvider locale={app.locale} messages={messages[app.locale]}>
-        <div className={className}>
-          <Sidebar />
-          <Notification />
-          {React.cloneElement(
-            this.props.children,
-            { auth, notify }
-          )}
-        </div>
+        { auth.isFetching ?
```
p0o: I would prefer to write this part like this:

```
{ auth.isFetching &&
  <Modal>
    <Loading />
  </Modal>
}

{ auth.isAuthenticated ?
  <div className={className}>
     <Sidebar />
     <Notification />
     { React.cloneElement(
        this.props.children,
        { auth, notify }
       )
     }
   </div>
   :
     <Splash />
}
```

src/app/Sidebar.sass
```diff
@@ -19,6 +19,19 @@ nav.sidebar
   left: 0
   color: #fff
 
+  .form-control, .input-group-addon
```
adcpm: I prefer use sass and i use BEM protocol when its not Bootstrap class.

src/wrapper.js
adcpm: It would not work as expected. What if i'm fetching auth and i'm not yet authentificated. I will see Modal Loading and also <Splash>. We dont want to show Splash before knowing if user is auth or not.

PR #150, opened 2016-11-27
Comments enhancement

p0o:


adcpm:
Great work! I've just find a small issue with like retry, the like counter is still increasing on every try look the last Like button:
spamlike

p0o:
@adcpm thanks for the report. should be fixed!

adcpm:
Work perfect now 👍

PR #154, opened 2016-11-28
Fix bug with user Profile data

p0o:


adcpm:
When i click from a profile to another profile the interface will have strange behavior, it's changing from the old the new data and reverse without reason few times.
shake

adcpm:
Work perfect now well done :)

PR #157, opened 2016-12-01
Comment and singlepost enhancements

adcpm:
I've mostly done this:

- Replaced TriggerPost to a footer navigation:
image

- Improve nested comment render
image

### DIFF
--- a/src/app/Header.scss
+++ b/src/app/Header.scss
@@ -159,7 +159,7 @@ header {
       &.active {
         box-shadow: inset 0 -1px 0 $primary-color;
 
-        i {
+        .icon {
           color: $primary-color;
         }
       }
--- a/src/app/Header.scss
+++ b/src/app/Header.scss
@@ -148,7 +148,7 @@ header {
     line-height: 53px;
 
     a {
-      display: block;
+      display: inline-block;
       height: 57px;
       line-height: 54px;
 
@@ -158,6 +158,10 @@ header {
 
       &.active {
         box-shadow: inset 0 -1px 0 $primary-color;
+
+        i {
+          color: $primary-color;
+        }
       }
     }
   }
--- a/src/app/Menu/MenuPost.scss
+++ b/src/app/Menu/MenuPost.scss
@@ -1,0 +1,21 @@
+.MenuPost {
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  display: block;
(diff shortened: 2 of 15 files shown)


src/app/Header.scss
```diff
@@ -158,6 +158,10 @@ header {
 
       &.active {
         box-shadow: inset 0 -1px 0 $primary-color;
+
+        i {
```
p0o: Please style `.icon` instead of html `i` tag. It helps to make it more clear and avoid css conflicts.

src/app/Menu/MenuPost.scss
```diff
@@ -0,0 +1,21 @@
+.MenuPost {
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  display: block;
+  box-shadow: inset 0 1px 0 #E7E7EC
+}
+
+.full-width .MenuPost {
```
p0o: is `.full-width` a modifier? It should be `.MenuPost--fullWidth` based on BEM

src/comments/CommentItem.scss
```diff
@@ -1,22 +1,55 @@
...
       color: $primary-color;
     }
   }
 }
 
+.CommentItem__content {
+  margin: 0.5em 0;
+  display: flex;
+
+  &.CommentItem__content--level-1 {
```
p0o: No need to nest modifiers inside elements, define them below the elements separately. The whole point of writing the full name in modifiers is to keep the CSS structure flat. This way the final CSS file would also be shorter since selectors are not long and nested.

src/styles/modules/post/post.scss
```diff
@@ -107,3 +107,8 @@
     border-width: 1px 0 1px 0 !important;
   }
 }
+
+.grid-row > .Comments > .CommentsList {
```
p0o: Try to avoid this pattern as much as possible. In this case if someone change `.grid-view` in the future it will break the style in other parts and is kinda hard to find and debug.

adcpm:
@mynameisek now with separator between buttons
image

src/app/Menu/MenuPost.scss
adcpm: Actually .full-width is already been used, it's a modifier for .app-wrapper

adcpm:
I've made the requested changes

PR #171, opened 2016-12-09
Comment reply feature

p0o:
This PR will restructure the comment state, add the feature of replying to comments and sort the comments based on votes. Still some refinements remained.

Should also fix the sorting of comments, and show the whole comments with replies open in single page views.

Should be merged after #170

### DIFF
--- a/src/comments/CommentsList.js
+++ b/src/comments/CommentsList.js
@@ -15,9 +15,11 @@ const CommentsList = ({
   }
 
   const { show, list } = comments.listByPostId[postId];
+  const showLimit = isSinglePage ? list.length : show;
+
   const visibleList = list.sort((item1, item2) => {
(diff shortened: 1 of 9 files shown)


src/comments/CommentsList.js
```diff
@@ -7,7 +7,7 @@ const CommentsList = ({ postId, comments, likeComment, unlikeComment, auth, open
   }
 
   const { show, list } = comments.listByPostId[postId];
-  const visibleList = list.slice(0, show);
+  const visibleList = list.sort((item1, item2) => item1.net_votes > item2.net_votes).slice(0, show);
```
adcpm: It should be based on reward, not net_votes.

adcpm:
When i try to comment while comment tab is close, pressing Enter will result to create a new line. When the comments are visible, pressing Enter would submit the form and work.

p0o:
@adcpm please take a look. Both should be fixed.

PR #174, opened 2016-12-10
Render category search within a link on sidebar

adcpm:
This update for add the search as a link on sidebar. While typing for search a category, what you are typing will also appear in a link. To be able to browse category even if its not on the list.
image


adcpm:
I've also fix the page /tags, it was not working anymore probably due on the Hard fork, the WS response change the naming from "categories" to "tags". And removed the triggers.

src/user/UserProfile.js
```diff
@@ -134,7 +125,17 @@ class Profile extends Component {
                   : () => this.props.addUserFavorite(username)
                 }
               />
-              { has(jsonMetadata, 'profile.name') ? jsonMetadata.profile.name : `@${username} ` }
+              { has(jsonMetadata, 'profile.name') ? jsonMetadata.profile.name : `@${username}` }
+              { ' ' }
+              <Follow
```
nil1511: Follow does not uses username, params and edit properties.

src/app/Sidebar.js
```diff
@@ -124,6 +124,24 @@ export default class Sidebar extends Component {
               #{category}
             </Link>
           </li>
+        );
+    }
+    return [];
+  }
+
+  renderSearchAsTag() {
+    const { search, categories } = this.state;
```
nil1511: `this.state.search` should be clear on onClickMenu.

src/user/UserProfile.js
adcpm: Right, well see thank you!

PR #176, opened 2016-12-11
Follower/Followed Search and paginate

nil1511:


src/user/UserList.js
```diff
@@ -1,18 +1,50 @@
 import React from 'react';
 import { Link } from 'react-router';
+import ReduxInfiniteScroll from 'redux-infinite-scroll';
+import _ from 'lodash';
```
adcpm: Avoid using root better to use `import { isNumber, filter, take } from 'lodash';`

nil1511: Udpated in d787625

src/user/UserList.js
```diff
@@ -1,18 +1,50 @@
...
-      <div className="UserList pvm">
-        <Avatar md username={this.props.username} className="mrm" />
-        <Link to={`/@${this.props.username}`}>@{this.props.username}</Link>
-      </div>
-    );
+    const filterText = this.state.filterText;
+    const defaultPageItems = 10;
+    const noOfItemsToShow = defaultPageItems * this.state.page;
+    const users = this.state.filterText ?
+      filter(this.props.users, user => user.indexOf(filterText) >= 0) : this.props.users;
```
p0o: why not using the Array.filter() in-built function of javascript?

src/user/UserList.js
```diff
@@ -1,18 +1,50 @@
...
+
+    return (<div>
```
p0o: It's nicer to start jsx the next line after paranthesis:

```
return (
  <div>
    ..
  </div>
);

src/user/UserList.js
nil1511: Array.filter might not be present in some older browsers.

PR #181, opened 2016-12-14
Donors

adcpm:
Create donors page and add donor tag on user profile :

image

image


src/donors/Donors.js
```diff
@@ -0,0 +1,30 @@
...
+      <ul>
+        { Object.keys(donors).map(donor =>
+          <h3>
+            <Link to={`/@${donor}`}>
+              <Avatar username={donor} sm /> @{donor}
+            </Link>
+            { ' ' }
+            <Donor rank={donors[donor]} />
+            { ' ' }
+            <small></small>
```
nil1511: for empty tags use it as `<small />` and why it this needed?

src/helpers/donors.js
```diff
@@ -0,0 +1,10 @@
+
+export default {
+  'steemit': 'gold',
```
nil1511: follow eslint rule. remove single quotes when not needed. this should be as steemit:gold

adcpm: I used the single quote because it would not work for username with - or . character

nil1511:
@adcpm not merge parallel branches.

src/helpers/donors.js
nil1511: use qoute only in name with - or . character.

PR #183, opened 2016-12-15
Dislike button on single page

p0o:
This PR should add dislike button on single page with showing the number of Dislikes


src/post/PostSingle.js
```diff
@@ -29,8 +29,9 @@ import * as bookmarkActions from '../bookmarks/bookmarksActions';
     ),
     openCommentingDraft: commentsActions.openCommentingDraft,
     closeCommentingDraft: commentsActions.closeCommentingDraft,
-    likePost: (id) => postActions.likePost(id),
-    unlikePost: (id) => postActions.likePost(id, 0),
+    likePost: (id) => postActions.votePost(id),
```
nil1511: Follow Eslint rules. single parameter should be as `id => console.log(id);` instead of `(id) => console.log(id)`;

PR #185, opened 2016-12-15
Bug Fixes

### LINKED ISSUE #175: Comments tab should not be openable if 0 comments
On the feed, the comment tab should not be openable/clickable if there is 0 comments available. See this demo:
0comment

nil1511:
- [x] Fix #175  Comments tab should not be openable if 0 comments

### DIFF
--- a/src/comments/Comments.js
+++ b/src/comments/Comments.js
@@ -26,7 +26,7 @@ export default class Comments extends Component {
   }
 
   static propTypes = {
-    postId: PropTypes.string.isRequired,
+    postId: PropTypes.number.isRequired,
     comments: PropTypes.object,
     getComments: PropTypes.func,
     className: PropTypes.string,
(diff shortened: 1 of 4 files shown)


nil1511:
@adcpm Loading on click should be there, as comments are not update realtime. Every time user click the loading button it will fetch the comment and will show them if there are any and if not the user can try again after sometimes by clicking it instead of refreshing the page.

adcpm:
@nil1511 we know already before fetching comment if there is any or not. We dont need to show a link and a loading when we know already there is 0 comment available.

nil1511:
@adcpm disabled click if no comments.

src/comments/Comments.js
```diff
@@ -26,7 +26,7 @@ export default class Comments extends Component {
   }
 
   static propTypes = {
-    postId: PropTypes.string.isRequired,
+    postId: PropTypes.number.isRequired,
```
p0o: postId is not a number but a string because we use it as an important object key in our state. API is returning number in some cases and that's why there is a react warning, however in some other cases postId is like this: `111.11.22`.

If you want to remove that warning you should make sure postId is always converted to string before passing as props. We should change the data to fit our requirement not the other way.

PR #186, opened 2016-12-15
Fix modal issue

p0o:
Should fix the issue with modals when you click on the same link that you opened the modal from.

### DIFF
--- a/src/post/PostSingleModal.js
+++ b/src/post/PostSingleModal.js
@@ -14,7 +14,7 @@ export default class PostSingleModal extends Component {
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


nil1511:
Why are we pushing to window.history and not react-router's browserHistroy. May be there won't be need to listen to history. 

https://github.com/ReactTraining/react-router/blob/master/docs/guides/Histories.md#examples

adcpm:
It's now working! Well done.

p0o:
@nil1511 pushing to browser history will trigger a react router's route change to the respective component. In this case we don't want to transition to the component which is defined for that path but we only want to open a modal and show the path in address bar for users to copy. Makes sense?

PR #187, opened 2016-12-15
Markdown fixes

nil1511:


src/post/body.js
```diff
@@ -1,32 +1,53 @@
...
+marked.setOptions({
+  renderer: new marked.Renderer(),
+  sanitize: true,
+  smartypants: true
+});
+
+export default (props) => {
+  const embeds = steemembed.getAll(props.body);
+  let body = props.body;
+  // body = striptags(remarkable.render(body), ['a', 'p', 'h1', 'h2', 'h3', 'img', 'code', 'pre']);
```
p0o: If a code is not necessary always remove rather than commenting. If we need it later we can always use git history for that.

PR #189, opened 2016-12-17
Transfer History

p0o:
This PR should add transaction history to `@user/transactions` 
Results are integrated with how Steemit is showing them and is implemented to support infinite scrolling but results older than 1 month is not returned from the API

### DIFF
--- a/src/user/TransferHistory.js
+++ b/src/user/TransferHistory.js
@@ -30,21 +30,20 @@ const renderReportFromOp = (op, username) => {
   }
 
   /*  all transfers involve up to 2 accounts, context and 1 other. */
-  let descriptionStart = "";
+  let descriptionStart = '';
   let otherAccount = null;
-  let descriptionEnd = "";
+  let descriptionEnd = '';
 
   if (type === 'transfer_to_vesting') {
     if (data.from === username) {
       if (data.to === '') {
         descriptionStart += `Transfer ${data.amount.split(' ')[0]} to STEEM POWER`;
-      }
-      else {
+      } else {
         descriptionStart += `Transfer ${data.amount.split(' ')[0]} STEEM POWER to `;
         otherAccount = data.to;
       }
     }
-    else if( data.to === username ) {
+    else if (data.to === username) {
       descriptionStart += `Receive ${data.amount.split(' ')[0]} STEEM POWER from `;
       otherAccount = data.from;
     } else {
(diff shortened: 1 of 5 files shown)


src/user/TransferHistory.js
```diff
@@ -0,0 +1,160 @@
...
+
+    return (
+      <table className="table">
+        <ReduxInfiniteScroll
+          loadMore={() => this.handleNextPage()}
+          elementIsScrollable={false}
+          hasMore={list.length > visibleItems}
+        >
+          { getOnlyViableTransfers(list).reverse().slice(0, visibleItems).map(op =>
+            <tr>
```
nil1511: set key props in tr.
set holderType to tbody in ReduxInfiniteScroll.

src/wallet/walletActions.js
```diff
@@ -0,0 +1,20 @@
+import Promise from 'bluebird';
+
+export const GET_WALLET = '@wallet/GET_WALLET';
+export const GET_WALLET_START = '@wallet/GET_WALLET_START';
+export const GET_WALLET_SUCCESS = '@wallet/GET_WALLET_SUCCESS';
+export const GET_WALLET_ERROR = '@wallet/GET_WALLET_ERROR';
+
+export const getWallet = (username) => {
```
nil1511: block is not needed if function just returns.

p0o: @nil1511 It looks cleaner for redux thunks since the function is returning another function. The other way it would be like this:

```
export const getWallet = (username) => (dispatch, getState, { steemAPI }) => {
  ...
}
```
which is not that easy to read.

nil1511: How about as this?
export const getWallet = (username) => 
  (dispatch, getState, { steemAPI }) => {
    ...
  }

p0o: @nil1511 the problem is the first scope is lost in your example. We can not easily see if the first one is the parent function since we can not nest the second one. I think this rule is much better for single line results or jsx (which naturally has clear scope) but not for multiple lines of JS.

I don't have a strong opinion on this but I just think it's not wrong in both cases and since I already wrote a lot of actions in this style I prefer to keep it this way (unless you believe it's wrong?)

nil1511: Ok. We will update the eslint rule for this.

src/user/TransferHistory.js
p0o: @nil1511 Good points! I made necessary changes please take a look.

PR #198, opened 2016-12-22
Allow Max 5 Categories

nil1511:


nil1511:
@adcpm Now support adding category on space.

nil1511:
@adcpm now converting to lowercase and added close button using <Icon />

PR #199, opened 2016-12-22
Modularized Lodash

nil1511:
For lodash to be modularized we should use it as
`import map from 'lodash/map'` not as `import {map} from 'lodash'`. 

This PR adds lodash plugin for webpack and babel that do the heavy lifting. 
Now you can use `import _ from 'lodash'` and it will automatically extract only those functions which  are used. 

TLDR  reduce lodash build size from 950kb to 330kb
https://dev.busy.org/js/statistics.html
https://busy6-pr-199.herokuapp.com/js/statistics.html


src/helpers/localStorageHelpers.js
```diff
@@ -1,5 +1,5 @@
 import store from 'store';
-import _ from 'lodash';
```
adcpm: Why you are not using "_" now?

nil1511: just using one function in that file and in whole project only this place was different.

PR #200, opened 2016-12-22
Multiple layouts

adcpm:


src/actions.js
```diff
@@ -16,45 +17,63 @@ export const ACCOUNT_SUCCESS = '@app/ACCOUNT_SUCCESS';
...
-
 export const openPostModal = createAction(OPEN_POST_MODAL);
 export const closePostModal = createAction(CLOSE_POST_MODAL);
 
-export const getConfig = () => {
-  return (dispatch, getState) => {
-    var req = {type: CONFIG_REQUEST};
-    dispatch(req);
+export const getConfig = () =>
+  (dispatch) => {
```
p0o: do not import `api`, get it from the third params in action thunk like this:
```
(dispatch, getState, { steemAPI }) => ...
```

src/helpers/localStorageHelpers.js
```diff
@@ -81,17 +81,16 @@ export const removeFavoriteCategory = (category) => {
   return true;
 };
 
-export const getDrafts = () => {
-  return store.get('drafts') || {};
-};
+export const getLayout = () =>
+  store.get('layout') || 'card';
```
p0o: I'm against this whole idea of using `store` package to directly work with localStorage. The reason is that we are changing the state and then change the localStorage and it makes a duplication of source of truth and makes it prone to errors in the logic. That's why I used persistence methods for `favorites` feature like here:

https://github.com/adcpm/busy/blob/dev/src/store.js#L63

Just make sure the redux state is saving the values in a separate key and add it to the line above like `favorites` and it stay persistent without storing and reloading manually.

adcpm: Nested path (app.locale, app.layout) are not working using persistState.
image
I'm not sure it's relevant to have a root state object for every features, i prefer having all the settings into 1 state (app).

adcpm: Maybe a "settings" state root object would make more sense. We could include favorites + bookmarks + locale + layout + drafts data on that. What your throughs about this?

nil1511: For nested path you need to implement slicer function. 

https://github.com/elgerlambert/redux-localstorage/blob/master/README.md#configslicer

p0o: right `slicer` function in config or settings key is the way to go.

PR #208, opened 2016-12-27
Comments sort

p0o:
should add the sorting ability to comments in single post page. needs style refinement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "react-redux": "^4.4.5",
     "react-responsive-utils": "^1.0.2",
     "react-router": "^3.0.0",
+    "react-select": "^1.0.0-rc.2",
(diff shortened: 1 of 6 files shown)


adcpm:
Trending sorting not seem to work, top reward should be shown first:
image

p0o:
@adcpm should work fine now.

adcpm:
Using trending sorting the result is not same on steemit or busy see:
http://localhost:3000/busy/@surpassinggoogle/what-exactly-will-happen-when-busy-org-starts-are-we-steemians-going-to-leave-steemit
https://steemit.com/busy/@surpassinggoogle/what-exactly-will-happen-when-busy-org-starts-are-we-steemians-going-to-leave-steemit
You probably have to deal with vote too when reward is 0 to sort trending.

PR #213, opened 2016-12-31
show a message for empty feeds

p0o:
This PR should show a message for empty feeds

### DIFF
--- a/src/feed/Feed.js
+++ b/src/feed/Feed.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 import ReduxInfiniteScroll from 'redux-infinite-scroll';
@@ -12,7 +11,6 @@ import * as commentsActions from '../comments/commentsActions';
 import * as bookmarkActions from '../bookmarks/bookmarksActions';
 import * as reblogActions from '../app/Reblog/reblogActions';
 import PostSingle from '../post/PostSingle';
-import Icon from '../widgets/Icon';
 
 import './Feed.scss';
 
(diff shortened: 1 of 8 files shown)


adcpm:
Actually the "empty feed message" will appear on Bookmark and user profile too. We need a way to handle custom message for these pages.

p0o:
@adcpm shouldn't be a problem. What kind of messages we need for these pages?

adcpm:
There is already an empty message on bookmark page, you can keep same text. About user profile i would see a different message if you its your profile you are visiting or if its another user profile. My profile the message could be : "You didn't publish any story yet. Start now" and another user "This user don't have any story published yet."

p0o:
@adcpm should work as expected

src/user/UserFeed.js
```diff
@@ -30,14 +31,20 @@ export default class UserProfileFeed extends Component {
...
     return (
-      <Feed
-        content={content}
-        isFetching={isFetching}
-        hasMore={hasMore}
-        loadContent={loadContentAction}
-        loadMoreContent={loadMoreContentAction}
-        route={this.props.route}
-      />
+      <di>
```
nil1511: I think It should be div. 😄

src/feed/Feed.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
```
nil1511: This file changes can be cleared.

src/user/UserFeed.js
p0o: @nil1511 Hahaha, right 😄

p0o:
@nil1511 thank you, should be okay now.

adcpm:
@p0o Empty message is visible on user profile while posts are fetching.

p0o:
@adcpm thank you. It's fixed now.

PR #214, opened 2016-12-31
Adding Favorite button to /tags page

p0o:
also fixed the bug on the page since it wasn't working.

### DIFF
--- a/src/favorites/FavoriteCategoryButton.js
+++ b/src/favorites/FavoriteCategoryButton.js
@@ -3,7 +3,10 @@ import Icon from '../widgets/Icon';
 
 const FavoriteCategoryButton = ({ name, onClick, isFavorited, noTitle }) => {
   return (
-    <h2 className="mt-1 mb-0 text-xs-center">
+    <h2
+      className="mt-1 mb-0 text-xs-center"
+      style={noTitle ? { display: 'inline-block' } : {}}
+    >
       <a onClick={onClick}>
         <Icon name={isFavorited ? 'star' : 'star_border'} sm />{ ' ' }
         { noTitle ? '' : `#${name}` }
--- a/src/favorites/FavoriteCategoryButton.js
+++ b/src/favorites/FavoriteCategoryButton.js
@@ -1,13 +1,10 @@
 import React from 'react';
 import Icon from '../widgets/Icon';
 
-const FavoriteCategoryButton = ({ name, onClick, isFavorited, noTitle }) => (
+const FavoriteCategoryButton = ({ children, onClick, isFavorited }) =>
   <a onClick={onClick}>
     <Icon name={isFavorited ? 'star' : 'star_border'} sm />{ ' ' }
-    <h2 style={{ display: 'inline' }}>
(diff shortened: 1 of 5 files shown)


src/favorites/FavoriteCategoryButton.js
```diff
@@ -1,12 +1,15 @@
 import React from 'react';
 import Icon from '../widgets/Icon';
 
-const FavoriteCategoryButton = ({ name, onClick, isFavorited }) => {
+const FavoriteCategoryButton = ({ name, onClick, isFavorited, noTitle }) => {
   return (
-    <h2 className="mt-1 mb-0 text-xs-center">
```
adcpm: Actually we have some h2 inside h1 tag, IMO FavoriteCategoryButton should not have styling by default and title. You could rewrite it like this:
```
import React from 'react';
import Icon from '../widgets/Icon';

const FavoriteCategoryButton = ({ onClick, isFavorited }) =>
  <a onClick={onClick}>
    <Icon name={isFavorited ? 'star' : 'star_border'} sm />
  </a>;

export default FavoriteCategoryButton;

```

src/favorites/FavoriteCategoryButton.js
```diff
@@ -1,12 +1,15 @@
 import React from 'react';
 import Icon from '../widgets/Icon';
 
-const FavoriteCategoryButton = ({ name, onClick, isFavorited }) => {
+const FavoriteCategoryButton = ({ name, onClick, isFavorited, noTitle }) => {
```
adcpm: You can use short syntax without bracket and "return".

src/favorites/FavoriteCategoryButton.js
adcpm: Abviously this would need some small change here too: https://github.com/adcpm/busy/blob/38fa1adb8fd4f7e13cd9a3dbfaad3092466d7ff6/src/feed/Page.js#L92

p0o: I think the matter is in category page we want to be able to favorite categories even by clicking on the category title (and not star only) By this change we will lose that possibility.

adcpm: Right, then we could keep the title, but remove h2 and styling.

p0o:
@adcpm made some changes. shouldn't have the previous issues.

adcpm:
@p0o the FavoriteCategoryButton comp should not have h2 tag in.

p0o:
@adcpm I'm not following. Why not? then how should we style it if we are not doing it inside the component?

adcpm:
Having h2 inside the component will restrict the style to be only H2. What if i want add the button inside h1 or h3 or p tag? Button comp should have only the <a tag then the <h2> tag should be added on the page /tags directly, not inside button.

nil1511:
store.favorites is not saved on localStorage may be it broke while adding slicer. 
You could just set `__persist` for favourite like here https://github.com/adcpm/busy/blob/dev/src/app/appReducers.js#L6

PR #221, opened 2017-01-06
checks if permlink already exists

### LINKED ISSUE #219: I'm not able to post story using same title
We need to check if the user have already a post using the title and if yes add a prexif (same as drafts) on the permlink.

nil1511:
Close #219

### DIFF
--- a/src/post/Write/EditorActions.js
+++ b/src/post/Write/EditorActions.js
@@ -5,6 +5,7 @@
 import Promise from 'bluebird';
 import assert from 'assert';
 import request from 'superagent';
+import steem from 'steem';
 import SteemConnect from 'steemconnect';
 import { browserHistory } from 'react-router';
 import { createAction } from 'redux-actions';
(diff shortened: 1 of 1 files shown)


nil1511:
steemit adds it on beginning. it would be better if we follow same pattern.

adcpm:
Any through on this @roadscape?

adcpm:
On Steemit random string prefix is used for duplicate title while suffix is used for comment. That make url of comment having both prefix and suffix with random string see https://steemit.com/test/@siol/re-siol-u9soq-hey-yo-20170106t054151946z

src/post/Write/EditorActions.js
```diff
@@ -19,23 +20,36 @@ export const CREATE_POST_ERROR = '@editor/CREATE_POST_ERROR';
 
 const requiredFields = 'parentAuthor,parentPermlink,author,permlink,title,body,jsonMetadata'.split(',');
 
+const getPermlink = (author, permlink) =>
```
p0o: Did you get the algorithm from steemit.com rep? If not please do so and if you did add a comment to the link of the source code you borrow the algorithm from so we can have it for future reference (if they changed it etc)

Also, since it's common better to add it to `src/helpers/steemitHelpers.js` like the way I did with commentPermlink.

src/post/Write/EditorActions.js
```diff
@@ -19,23 +20,36 @@ export const CREATE_POST_ERROR = '@editor/CREATE_POST_ERROR';
...
-          .then(({ result }) => {
-            browserHistory.push(`/${parentPermlink}/@${author}/${permlink}`);
-            return result;
-          }).catch(err => err)
+        promise: getPermlink(author, postData.permlink)
+          .then(permlink =>
+            SteemConnect
+              .commentAsync(parentAuthor, parentPermlink, author, permlink, title, body, jsonMetadata)
+              .then(({ result }) => {
+                browserHistory.push(`/${parentPermlink}/@${author}/${permlink}`);
```
p0o: It's an anti-pattern to manipulate the UI from an action. The action is supposed to changed the state and UI components should respond to state change.

nil1511:
This was merged by mistake and was reverted in https://github.com/adcpm/busy/commit/18f5c9ed504848c02f78b3a2cec5e02484f14a18

PR #224, opened 2017-01-07
Transfer

adcpm:
- Create page /transfer
- Add Coinmarketcap API to show Steem rate


src/app/Sidebar.js
```diff
@@ -3,19 +3,18 @@ import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 import { Link } from 'react-router';
 import { FormattedMessage } from 'react-intl';
+import fetch from 'isomorphic-fetch';
```
nil1511: missing in package.json

src/app/Sidebar.js
```diff
@@ -58,10 +57,17 @@ export default class Sidebar extends Component {
...
         categories: categories,
         props: result.props,
-        feedPrice: result.feed_price
       });
       this.getFollowing();
     });
+
+    fetch('https://api.coinmarketcap.com/v1/ticker/steem/')
+      .then(res =>
+        res.json().then((json) => {
```
nil1511: Writing as would be better.
```js
fetch('https://api.coinmarketcap.com/v1/ticker/steem/')
      .then(res => res.json())
      .then((json) => {
        const priceUsd = json[0].price_usd;
        this.setState({ price: priceUsd });
      });
```
Also it would better to save price in redux save.

src/app/Sidebar.js
```diff
@@ -160,14 +166,22 @@ export default class Sidebar extends Component {
...
+    const { search, price, props, menu } = this.state;
     const user = this.props.auth.user;
-    const search = this.state.search;
 
-    if (_.has(this.state.feedPrice, 'base')) {
-      var power = formatter.vestToSteem(user.vesting_shares, this.state.props.total_vesting_shares, this.state.props.total_vesting_fund_steem);
-      var base = (this.state.feedPrice.base).replace(' SBD', '').replace(',', '');
-      var dollar = (parseFloat(base) * (parseFloat(user.balance) + parseFloat(power))) + parseFloat(user.sbd_balance);
-    }
+    const power = props
```
nil1511: (props) this.state.props is confusing would be nice to name as categoriesProp(if it is).

src/wallet/Transfer.js
```diff
@@ -0,0 +1,120 @@
...
+import numeral from 'numeral';
+import Icon from '../widgets/Icon';
+import Header from '../app/Header';
+
+@connect(
+  state => ({
+    auth: state.auth,
+  })
+)
+
```
nil1511: Remove the space.

src/wallet/Transfer.js
```diff
@@ -0,0 +1,120 @@
...
+                  placeholder="Memo"
+                  type="text"
+                  className="form-control form-control-lg"
+                />
+              </blockquote>
+            </div>
+            <div className="form-group">
+              <a
+                href={url}
+                target="_blank"
```
nil1511: > security tip 

always add `rel="noopener noreferrer"`  with target="_blank" if url is leaving the website.

src/wallet/Transfer.js
adcpm: Is there any reason to not having a space between connect and the class?

p0o: @adcpm `connect` is a Higher Order Component which means is containing the whole class like a function. As we don't leave a space in this example:
```
function something(

{ foo: 'bar' }
) { ...
```
IMHO we also shouldn't leave there because empty lines are used for separating conceptually independent parts but `connect` and the class it's containing are tightly coupled.

src/wallet/Transfer.js
p0o: @nil1511 why is that? hiding track of staging URL?

nil1511: by default new url expose `window.opener` pointing to previous page. noopener will set is to null and noreferrer removes referrer information.

adcpm: That's a linter rule

p0o: @nil1511 I just read about it. `window.opener` is a crazy API! thank you

Issue #218, opened 2017-01-05
Favorites does not persist anymore on localStorage


adcpm:
We probably should use the _persist param for favorite on redux state.

p0o:
@adcpm no we decided to deprecate it. I will merge the fix soon as we will just persist the three main state keys `editor`, `favorite` and `app` for now (or maybe as a finalized solution if we don't face any issues later)

PR #228, opened 2017-01-08
Fix setup issues

p0o:
some fixes with codebase initial setup and dev tools. commits are self explanatory.


nil1511:
Closes #218

PR #234, opened 2017-01-10
Fix 231

### LINKED ISSUE #231: Empty Drafts dont need to be Saved
emptydraft

When we edit new article and delete content (make it Empty), it still save it.
It's a small bug (not high priority) but I think it's necessary to have the feature fixed. Otherwise It will spam the draft manager will tons of empty edits.

nil1511:
Closes #231


nil1511:
@adcpm Article is not being saved on clicking "Write your story...". It is saved when we write something in the body.

PR #245, opened 2017-01-14
update editor for comments

nil1511:


nil1511:
Currently this pr is blocked. As [EmojiPlugin](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-emoji-plugin/README.md) only show suggestion list below the text and it get partially hidden as comment form always sticks bottom of page. Should we disable suggestion for comments?

adcpm:
Sorry, I've been reconsidering this mission and actually we dont need editor on comment.

PR #253, opened 2017-01-17
Chat issues fix

p0o:
This PR should fix the send and receive issues in chat and bring back the unread counter


p0o:
@adcpm all issues should be fixed now.

adcpm:
@p0o about category chat its still not working properly for me http://localhost:3000/messages/busy when sending a message i would no see it on another side unless i refresh the page.

PR #255, opened 2017-01-17
Update style

### LINKED ISSUE #230: Image style in post
image

----
https://steemit.com/health/@saramiller/how-to-yoga-a-chart-of-asanas

image

nil1511:
Will closes #230 
@adcpm some post created on steemit may break on busy.
like  https://steemit.com/health/@saramiller/how-to-yoga-a-chart-of-asanas
Images are placed with `&nbsp` exactly to fit well on steemit. This break on busy as container class limit max-width to 800px and steemit uses 50rem width which for me is ~840px.

### DIFF
--- a/src/post/PostSingleContent.scss
+++ b/src/post/PostSingleContent.scss
@@ -11,17 +11,21 @@
 
 .PostSingleContent__content,
 .PostSingleContent__replies .CommentBody {
-  font-size: 20px;
+  width: 820px;
+  font-size: 1em;
   line-height: 1.6em;
   overflow: hidden;
 
   img {
-    max-width: 100%;
     height: auto;
     margin: 2em 0;
     border: 1px solid rgba(0, 0, 0, 0.1);
-    display: block;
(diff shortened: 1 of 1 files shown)


src/post/PostSingleContent.scss
```diff
@@ -11,17 +11,21 @@
 
 .PostSingleContent__content,
 .PostSingleContent__replies .CommentBody {
-  font-size: 20px;
+  width: 820px;
```
adcpm: Size should come from "container" and not be defined here

nil1511: container class is defined by bootstrap. can we have another class for this width?

adcpm: The point is that i dont want change the width, Medium use 700 width and doing fine, it doesnt matter is some article failed when it was made for Steemit width.

nil1511: Ok. Closing the pr and #230 is not solvable.

PR #258, opened 2017-01-18
Design

adcpm:


src/post/PostSingleContent.js
```diff
@@ -36,38 +50,68 @@ const PostSingleContent = ({ content, toggleBookmark, bookmarks }) => {
...
+            ))}
+          </div>
+        }
       </div>
 
-      { content.children > 0 && (
-        <div className="PostSingleContent__replies py-3">
-          <div className="container">
-            <h1><Icon name="reply" lg /> Comments ({content.children})</h1>
+      <MenuPost
```
p0o: MenuPost is not supposed to be here. The goal of PostSingleContent is just to render the Post content and handling extra behaviors of the page better to be handled by top containers like the way it used to be. Is there a reason for moving it?

src/post/PostSingleModal.js
```diff
@@ -59,41 +57,64 @@ export default class PostSingleModal extends Component {
   }
 
   render() {
-    const { content, sidebarIsVisible, reblog, isReblogged, openCommentingDraft } = this.props;
+    const {
```
p0o: I don't like this pattern. If variables are meant to be used only once or twice or are just meant to be passed to another component better to use the full form of `this.props.name` instead of a huge extraction which is actually making the size of the whole code in the file bigger than smaller.

src/post/actionButtons/PayoutLabel.js
```diff
@@ -1,12 +1,22 @@
...
-    </span>
-    {value}
-  </a>;
+const PayoutLabel = ({ post, onClick }) => {
+  const payout = parseFloat(post.total_payout_value)
+    + parseFloat(post.total_pending_payout_value)
+    + parseFloat(post.curator_payout_value);
+  const maxAcceptedPayout = parseFloat(post.max_accepted_payout);
+  return (
+    <a onClick={onClick && onClick}>
```
p0o: there is no need for this. If you don't send `onClick` from the main component react will automatically remove this listener so it's safe to use: `onClick={onClick}`

src/post/PostSingleContent.js
adcpm: It's because MenuPost is now not fixed on screen but appear between the end of article and Author bio

PR #275, opened 2017-01-24
Chat fix

p0o:
* should fix issues with chat previous messages loading
* should dispatch chat actions after authentication
* should add emoji picker to chat editor
* should render emojis in chat content


adcpm:
Can we merge this PR?

p0o:
@adcpm yes sure

PR #294, opened 2017-01-31
New Tooltip/Popover/Modal Implementation

p0o:
This PR will add the implementation for new Tooltip, Popover and Modal to be used for busy.


adcpm:
@p0o Can you add react-gateway on this?

p0o:
@adcpm yes sure should refactor to portal method.

src/widgets/tooltip/ProfileTooltip.scss
```diff
@@ -0,0 +1,11 @@
+.ProfileTooltip {
+  font-size: 16px;
+  text-align: center;
+  width: 250px;
+  height: 260px;
```
adcpm: The content of the profile tooltip is not visible entirely, it's crop on the user location, can you make it more height like 280px?

src/widgets/tooltip/Tooltip.js
```diff
@@ -0,0 +1,109 @@
+import React, { Component, PropTypes } from 'react';
+import { Gateway } from 'react-gateway';
+import { getElementPosition } from './tooltipHelpers';
+import SimpleTooltip from './SimpleTooltip';
+
+const DELAY = 500;
```
adcpm: Isn't too slow? I would see little bit lower delay.

src/wrapper.js
```diff
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 import { IntlProvider } from 'react-intl';
+import { GatewayProvider, GatewayDest } from 'react-gateway';
 import { Tooltip, actions as tooltipActions } from 'redux-tooltip';
```
adcpm: Why we still have the dependancy `redux-tooltip`?

src/widgets/tooltip/Tooltip.js
p0o: @adcpm Yes thanks. Delay value works accurately. The problem was with profile tooltip that was taking longer because we couldn't see loading time. Fixed!

p0o:
@adcpm all issues are fixed but I can't simply fix this issue:

> When i have a tooltip open in header and i scroll the page, the tooltip will follow the scroll instead. It should not move as the icon is not moving see

The thing is it only happens for header because it's fixed positioned and tooltip system works by getting the position of one element and absolutely place it somewhere. Because tooltip in DOM is out of any element it cannot understand the element is fixed positioned and by scrolling the element is still there.

I can solve it by disappearing tooltips with the event of mouse scroll (looks logical to me) or any other suggestion you may have?

adcpm:
@p0o Ok yes, removing the tooltip on scroll is seem to be common behavior you can go for it.
- Also on the header the tooltip is very far from the icon see:
image
Can we get it more near same it was before?

PR #312, opened 2017-02-03
Don't Stringify JsonMetaData. Already handled by steem-js

### LINKED ISSUE #311: Comment json_metadata not properly set
json_metadata on comment is not properly set. See: 
https://steemd.com/tx/9289d9e9cf0cb2afde0914f71b27f5edf5f9613d
It should be an object with `app` and `format` param too.
Reported by @abit

nil1511:
Fix #311 

Ref https://github.com/steemit/steem-js/blob/6ce1a43afa34a5c82509c3f15cf3cbb8642c0f3a/src/broadcast/index.js#L91


adcpm:
Can you add the `app`?

nil1511:
Added `app`

PR #340, opened 2017-02-07
remove reblog button on users own posts

### LINKED ISSUE #338: Do not show reblog button on own blog
It's unnecessary since we cannot reblog our own content.
If you want to keep it, I recommend graying it out, or do not change the design at on-hover event. A message could also help. New users don't know why nothing happens, when there is a button and they pressed it.

p0o:
should fix #338


adcpm:
I agree about `canReblog`, it make more sense. Can you change it @p0o?

p0o:
@nil1511 Thanks, good point! Changed.

PR #352, opened 2017-02-09
Fix the style and position of comment after replying

### LINKED ISSUE #325: Comment format issues right after submission
[](http://www.steemimg.com/image/G0v2Z)

Note that both of my comments are on the same level. 
Every freshly submitted comment appears like comment#2: smaller profile picture and missing action row. 
Comment#1 only appears normal, because I'd reloaded the article before I've sent comment#2. So everything goes to normal after a reload. However, the problem exist on post modal and page view as well. Tested with FF.

p0o:
should fix #325 

screen shot 2017-02-09 at 3 14 52 pm


p0o:
@adcpm thanks please try again

PR #354, opened 2017-02-10
Thumb first on youtube embeds

p0o:
This PR uses new changes in [embedjs](https://github.com/adcpm/embedjs) to show the thumbnail on youtube embeds on PostFeed view. there is a play button over thumbs and will transition to youtube iframe embed on click.

### DIFF
--- a/src/post/Feed/PostFeedCard.js
+++ b/src/post/Feed/PostFeedCard.js
@@ -129,7 +129,7 @@ const PostFeedCard = ({
       <BodyShort body={post.body} />
     </div>
 
-    { (embeds && embeds[0]) &&
+    {(embeds && embeds[0]) &&
       <PostFeedEmbed post={post} />
     }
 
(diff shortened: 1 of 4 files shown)


src/post/Feed/PostFeedCard.js
```diff
@@ -128,8 +129,8 @@ const PostFeedCard = ({
       <BodyShort body={post.body} />
     </div>
 
-    {_.has(embeds, '[0].embed') &&
-      <div className="PostFeedCard__thumbs" dangerouslySetInnerHTML={{ __html: embeds[0].embed }} />
+    { (embeds && embeds[0]) &&
```
adcpm: We should no use space here `{ (`

p0o:
@adcpm Thank you great point. I added `autoplay` as an option to embedjs but its default is true so no need to specify it here. Also The API is documented in [embedjs page](https://github.com/adcpm/embedjs/blob/dev/README.md)

adcpm:
Awesome, thank you for the documentation!

PR #356, opened 2017-02-10
Separate messages by day groups

yamadapc:
- Uses `react-intl` and a new `dateHelpers` module to separate messages by
  date group
- Renames messages components so they make more sense
- Documents the grouping functions
- Simplifies message sorting
- Offsets `startOfDay`s so `react-intl` doesn't need to handle timezones
  propertly (which it isn't)

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "lodash": "^4.0.0",
     "lodash-webpack-plugin": "^0.10.6",
     "matchmedia": "^0.1.2",
-    "moment": "^2.17.1",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "lodash": "^4.0.0",
     "lodash-webpack-plugin": "^0.10.6",
     "matchmedia": "^0.1.2",
+    "moment": "^2.17.1",
```
adcpm: We dont want moment.

PR #369, opened 2017-02-12
Editor Fixes

### LINKED ISSUE #249: Copy paste markdown on Editor won't convert directly
pastemarkdown
When we copy paste markdown to our Editor is there a way to convert directly the markdown, like it's done when we type it live 

live

most writers like to write markdown first then paste to their editor.

cc @adcpm @nil1511

### LINKED ISSUE #344: Editor should handle trailing space
**About Issue** 
Making a text bold or italic with the editor revert to plain text after publish, if the last character of the selected text is a space.

**Url(If any).**
Happened in test content (changing), see screenshots instead.

**Expected behavior and actual behavior**
In the editor, it changes visually when I format as normal, but after publishing, there are only Markdown `**` & `_` 
It happens inline too, so my first thought is wrong (auto-delete trailing space). Markdown sign should jump back to the last character that is not a space (or any other that it cannot handle).

**Steps to reproduce the problem.**
1. Select a text with a trailing space
2. Make it bold or italic
3. Publish

**Browser and version.**
51.0.1 (64-bit) Linux 

**Published bold text**

---

**Published italic inline text**

---

### LINKED ISSUE #353: Cannot escape from code block in the editor
#### Expected behavior

The user should be able to continue their work without reverting their last action. Adding a new line after the code block automatically, if there is no one already, could solve this issue. 

#### Actual behavior

When the last line in the editor is a part of our article that we want to be a code block, and we set it with the editor, we cannot escape (make a new paragraph) with enter, because it will make a new line in the code block only.
The only way adding a new line is reverting the code block format, adding a new line to be able to continue the document, and then format the previous paragraph to code block again. 

#### How to reproduce

1. Write something in the editor
1. Make the last line a code block with the editor
1. Try to continue the document after the code block

* Browser: Firefox 51.0.1
* Operating system: Ubuntu 16.10 (64bit)

<sup>PS.: I would change the order of the template from `1. Expected 2. Actual 3. How to` to `1. Actual 2. Expected 3. How to`. It's true that when the problem happens, the first thing is 'what we want' and the second is 'what happens'. However, when we complain about something, we start with `what happened`. The `what we wanted` part is an explanation that we add to become more clear about what is our issue. It would be even better to use in the future, when most of the issues will be feature requests.</sup>

### LINKED ISSUE #367: Unable to post Markdown image with the editor
#### Expected behavior

Posting images via Markdown should be allowed.

#### Actual behavior

I cannot post images via Markdown. This is strange, because I was able to do it before. (I'm using the dev branch for tests.) It appears upon editing, but disappears [after submission](https://busy.org/teszt/@magyarul/teszt-2-markdown-kep-friss-bejegyzesben).

#### How to reproduce

Try to post an embed image via Markdown.

* Browser: Firefox 51.0.1
* Operating system: Ubuntu 16.10

nil1511:
Using image block from plugin
Added Delete btn on image. 
Fix #344 
Fix #249 
Fix #353 
Possible Fix #367


p0o:
Great work on all these libraries but may I ask why instead of forking these packages you cloned them and added your commits from the scratch? If these issues are bugs in these libraries using github handles in `package.json` are supposed to be a temporarily solution and we need to convince the package owners to merge your contributions. Having packages without storing past commits prevent you to send them a PR and merge your changes.

nil1511:
@p0o I have created pr for those repo. Waiting for there response it is breaking changes for the plugins. When they will merge I will replace them with npm package.

p0o:
@nil1511 yes I was mistaken by this [rep](https://github.com/nil1511/draft-js-delete-img-btn-plugin) but now that I look it seems your wrote it yourself. Great job and thanks.

PR #378, opened 2017-02-16
Translate

adcpm:
- [x] Change translation system to use the Steem blockchain comment to store translations instead of static file. This article is used for the translations: https://steemit.com/test/@siol/translations
- [x] Add Spanish language
- [x] Created a page for see untranslated text:
https://api.busy.org/v1/translate/missing?parentPermlink=test&author=siol&permlink=translations or
https://api.busy.org/v1/translate/missing?parentPermlink=test&author=siol&permlink=translations&locale=fr
- [x] Removed translations statics files from the repo


src/translations/translationHelper.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const getMessages = (replies) => {
+  const localesMap = {};
+  const messages = {};
+  const sortedReplies = _.sortBy(replies, 'net_rshares');
+
+  Object.keys(sortedReplies).forEach((key) => {
+    const reply = sortedReplies[key];
+    if (reply.depth === 1 && reply.body.length === 4 && reply.body.substring(0, 2) === '# ') {
```
p0o: for conditions that are long and confusing like this, it's better to give them a name and use the name inside the if for easier readability:
```javascript
// not sure if name fits the action or not just example
const isTranslationCommentAvailable = (reply.depth === 1 && reply.body.length === 4 && reply.body.substring(0, 2) === '# ');

if (isTranslationCommentAvailable) {
  ...
}
```

src/wrapper.js
```diff
@@ -40,13 +45,22 @@ export default class Wrapper extends Component {
     this.props.getStoredBookmarks();
     this.props.getRebloggedList();
     this.props.getRate();
+    this.loadMessages();
   }
 
+  loadMessages = () => {
+    steemAPI.getState('/test/@siol/4psulv-test', (err, result) => {
```
p0o: Could be a good idea to instead of hardcoding put the string in a config file. We should do it for many other things too. It's easier to have it there in case of changing. Something like `src/config/index.js` and import as `import { translationsURL } from '../config';`

p0o:
@adcpm there are some heavy conventions in this PR that needs to be addressed as comments to be understandable for others. May you please add some lines of detailed explanation in `wrapper.js` where you are loading the translations and include a direct link to the post that translations exist in it?

adcpm:
Thank you i've made the requested changes.

src/translations/translationHelper.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+  return object;
+};
+
+export const getMessages = (replies) => {
+  const localesMap = {};
+  const messages = {};
+  const sortedReplies = _.sortBy(replies, 'net_rshares');
+
+  Object.keys(sortedReplies).forEach((key) => {
```
p0o: I guess instead of two `forEach` with O(2n) complexity you can just use one like this:

```javascript
Object.keys(sortedReplies).forEach((key) => {
     const reply = sortedReplies[key];
     if (isRootLocaleComment(reply)) {
       locale = reply.body.replace('#', '').replace(' ', '').toLowerCase();
       if (locale) {
         const object = toObject(reply.body);
         messages[locale] = { ...messages[locale], ...object }; // es7 way
       }
    }
});
```

src/translations/translationHelper.js
```diff
@@ -0,0 +1,49 @@
...
+  lines.forEach((line) => {
+    const message = line.split(':');
+    if (message.length === 2) {
+      object[message[0].trim()] = message[1].trim();
+    }
+  });
+  return object;
+};
+
+export const getMessages = (replies) => {
```
p0o: I have security concerns over this function. There is no checks for the username or number of votes so the latest comment will override the translations in that locale. For example if I add a comment after you with the root comment of `FR` and write some french curses inside it it will be the primary language of the platform for french language which is an obvious vulnerability to attacks.

The first way to prevent it is to pick the root comment with the most upvotes between similar other locales but even doing so doesn't guarantee people using bots to promote a no-acceptable translation into the system. For now I suggest using a whitelist of users so if the username is not in that whitelist we just ignore it. It's easy to update the whitelist (could be a simple array in config.json) when we see a community voted translation is introduced.

src/translations/translationHelper.js
```diff
@@ -0,0 +1,49 @@
+import _ from 'lodash';
+
+const isRootLocaleComment =
+  reply => (reply.depth === 1 && reply.body.length === 4 && reply.body.substring(0, 2) === '# ');
+
+export const toObject = (body) => {
+  const object = {};
+  const lines = body.split(/\n/);
+  lines.forEach((line) => {
+    const message = line.split(':');
```
p0o: This code will have unwanted results if the translation strings includes `:` inside like this :
```json
busy_team: 'We are proud to have: Fabien, Pooria, etc'
```

src/translations/translationHelper.js
adcpm: It's already been sorted by `net_rshares`. Number of vote or voter dont matter and should not. The latest comment will not overwrite the translation, unless it has more `net_rshares` than others translations.. If there is 3 comments root with `# FR` all nested comments will be merged together and only the best net_rshares translations would be chosen. I don't see how it's vulnerable to attacks.

p0o: @adcpm sorry didn't see the sorting part but even so, are we going to load the data into our language system without any approval and only based on `net_rshares`? What if the data were incomplete and in wrong format and the author didn't bother to edit?

src/translations/translationHelper.js
adcpm: This code will not work. The condition "isRootLocaleComment" remove all translations replies. And the result will be { fr: {}, en: {}, de: {} } (empty objects).

src/translations/translationHelper.js
adcpm: Ok thx, i've updated the code to disallow content with negative `net_rshares`. So we can flag the content if it's incomplete / wrong format.

PR #389, opened 2017-02-20
Add loading state for messages (#388)

### LINKED ISSUE #388: Add loading state to the messages page

yamadapc:
This closes #388

### DIFF
--- a/src/messages/MessageList.js
+++ b/src/messages/MessageList.js
@@ -7,6 +7,7 @@ import { connect } from 'react-redux';
 
 import './MessageList.scss';
 import InfiniteScroll from '../widgets/InfiniteScroll';
+import Loading from '../widgets/Loading';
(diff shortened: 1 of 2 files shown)


adcpm:
Can you remove this issue and the console.log with the object?
image

adcpm:
There should be a parameter `isLoaded` to avoid load multiple time the same request. Actually if i load all the messages on the chat history, everytime if i up, the loading will come up and try to load messages while there is no messages to load.

yamadapc:
There's a `hasMore` parameter, but I've to change it indeed. Also shouldn't be offset based, but based on the last known `uuid`s or timestamps

adcpm:
Yes using uuid would avoid loading duplicate message. It's pretty much what Steem doing on the blockchain (pagination with username + permlink).

PR #394, opened 2017-02-26
Add Premlink to comments

nil1511:


src/comments/CommentItem.js
```diff
@@ -60,6 +60,11 @@ export default class CommentItem extends Component {
     }
   }
 
+  updateCommentPostion = () => {
+    // should happen after react-router updated the route
+    setTimeout(() => this.checkHashLink());
```
p0o: Not a necessary hack. I think you can just use `componentDidMount` to call `checkHashLink()` and use `@withRouter` from `react-router` package to have access to `this.props.location` in case.

nil1511: componentDidMount is not called is just hash is changed. It is for clicking on different comment link for same post eg
From `http://localhost:3000/category/@user/title#hash1` to `http://localhost:3000/category/@user/title#hash2` 

ComponentDidMount is not called in this case. 

Also now checkHashLink is in CommentItem's ComponentDidMount. Shouldn't it would be better to have it in CommentsList. Generally now If user would go to comment's link checkHash would be called as many times as no. of comments and so will `document.getElementById` and `targetElm.scrollIntoView()`

p0o: @nil1511 if you have @withRouter in your component when the hash is changed `this.props.location` should change so at least you should have `componentWillReceiveProps` as an update right?

Well yes I think we can move it there but `elm.scrollIntoView()` will be called only once since there is an [early exist](https://github.com/adcpm/busy/blob/3d80ee9815c10ff74a10bc21c3e0fe95d3b64479/src/comments/CommentItem.js#L72) for other comments.

nil1511: @withRouter is already in the component and first i tried with `componentDidUpdate` new props it was like this 
```js 
componentDidUpdate(prevProps) {
    const prevHash = _.get(prevProps, 'location.hash');
    const currentHash = _.get(this.props, 'location.hash');
    if (prevHash !== currentHash) {
      this.checkHashLink();
    }
  }
```
But router was not taking new position I think that has to do with `react-router-scroll/lib/useScroll`

p0o: @nil1511 I can access it in `componentWillReceiveProps` but in order to trigger an update from it we should use state which leads to more unnecessary verbosity. I approve the PR for now but I'm afraid it stops working based on future updates of RR or anything similar.

nil1511: location is accessible in componentWillReceiveProps. But checkHashList won't work there react-router resets the scroll of the page when the route is changed.

PR #395, opened 2017-02-26
Handle username which does not exist and show on profile

nil1511:


nil1511:
@adcpm Updated. can you check again?

adcpm:
We should have same styling for every error component (with div > h3) see  https://github.com/adcpm/busy/blob/dev/src/statics/EmptyUserProfile.js

nil1511:
Updated to match styling with EmptyUserProfile.

nil1511:
@p0o  Added DefaultMessage to all components.

PR #397, opened 2017-02-28
Translations

p0o:


src/statics/About.js
```diff
@@ -11,42 +11,90 @@ export default () =>
...
+              id="@statics/ensure_compensation"
+              defaultMessage="Ensuring compensation for the creators of value"
+            />
+          </h1>
+          <p>
+            <FormattedMessage
+              id="@statics/busy_mission_2"
+              defaultMessage="Artists and User-generated content create billions dollars worth of value absorbed by the shareholders of these intermediaries and centralized services"
+            />
+            .
```
adcpm: Why we have a dot here? It should be on the defaultMessage sentence isnt?

adcpm: Here too, i think you have to include all of them into the defaultMessage. Thai language dont have dot for example.

src/statics/About.js
```diff
@@ -11,42 +11,90 @@ export default () =>
...
+            {' '}
+            <FormattedMessage
+              id="@statics/busy_mission3"
+              defaultMessage="As a non-profit and decentralized organization, Busy provides the power of cryptocurrency and blockchain innovations within a platform that returns much of its value to the creators ensuring fair compensation"
+            />
+            .
```
adcpm: Here too, i think you have to include all of them into the defaultMessage. Thai language dont have dot for example.

src/statics/About.js
```diff
@@ -11,42 +11,90 @@ export default () =>
...
-            fee that allows people to transact safely and freely.</p>
+          <h1>
+            <FormattedMessage
+              id="@statics/sharing_economy"
+              defaultMessage="Creating a true sharing economy"
+            />
+          </h1>
+          <p>
+            <FormattedMessage
+              id="@statics/sharing_economy_2"
```
adcpm: Why separate a text into multiple translations text? This will probably change and will require more work to maintain with lot of translation isnt?

src/statics/About.js
```diff
@@ -11,42 +11,90 @@ export default () =>
...
+              defaultMessage="Uber or Airbnb are often described as the sharing economy, a nice notion that peers create and share in value. But these businesses have little to do with sharing, they in fact simply aggregate as middleman through their centralized, proprietary platforms and do resell by taking huge commissions"
+            />
+            .
+            {' '}
+            <FormattedMessage
+              id="@statics/sharing_economy_3"
+              defaultMessage="Today's big disrupters are about to get disrupted"
+            />
+            !
+            {' '}
```
adcpm: IMO spaces also should be part of the translation.

p0o:
@adcpm Please check if it's ready to merge. I moved dots inside translations but in most of the cases the reason for separating one translation into separate parts is to format them better (for example link the word `contact us` in the middle of sentence.

PR #401, opened 2017-03-06
Add Yarn 🚀

nil1511:


adcpm:
Great! We probably also need to remove npm scripts from package.json file so Heroku can switch to yarn.

nil1511:
heroku will use yarn if yarn.lock file is present. npm script should not be an issue.

PR #407, opened 2017-03-12
Simple tipping

adcpm:
These are small changes to make tipping and transfers more easy.

- Added a simple button "Tip the author" in the end of post.
image

- And a button "transfer" on user profile.
image

Waiting for your review or suggestions.

### DIFF
--- a/src/app/Menu/MenuPost.js
+++ b/src/app/Menu/MenuPost.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import numeral from 'numeral';
 import { FormattedMessage } from 'react-intl';
-import _ from 'lodash';
+import { Link } from 'react-router';
 import { SimpleTooltipOrigin } from '../../widgets/tooltip/SimpleTooltip';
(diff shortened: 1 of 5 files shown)


tibonova:
_Sorry for being late with the response._

Oh yeah, I could delete my workaround for the tipping from my "footer" in my posts templates. Every Steem UI really need this, since some authors posting timeless contents, like a part from their books, and the payment period is (will be) only a week. 

I think there are more tipping there where is easy to do it. I'm not saying that tipping could substitute the value that Steem distributes to authors. That never worked. But it's useful for additional/optional income on long term.

Since Steem is the best blockchain I know for microtransactions, I wonder why we don't use it, advertise it... this would be long. 
Just look at my experiments with @pipes where, depending on the random number, I could advertise a content with 0.001 SBD, and users aren't offended (I wouldn't be, too) because the tiny amount, I get good responses.

I would like this feature would make available to _power up_ the tipped amount directly as an option.

PR #411, opened 2017-03-15
Webpack 2 & new Build system

p0o:


webpack/makeConfig.js
```diff
@@ -51,13 +50,11 @@ function makePlugins(options) {
 
   if (isDevelopment) {
     plugins = plugins.concat([
-      new webpack.optimize.OccurenceOrderPlugin(),
```
nil1511: OccurenceOrderPlugin is renamed to OccurrenceOrderPlugin.

p0o: > OccurrenceOrderPlugin is now on by default
It's no longer necessary to specify it in the configuration.

nil1511: What is current build size?

adcpm: Deployment failing on Heroku

p0o:
Many changes in build pipeline we have here. The dev environment is still a little buggy (font loading etc) but the production should work as expected. @nil1511 and @adcpm please take a look and review the implementation until I fix small remaining issues.

* webpack is upgraded to version 2
* all static files on the server are now being cached for `one week`
* The production JS is splitted into three parts, `main=720kb`, `vendor=2mb` and `manifest=15kb`
* All files from public are removed and static ones are now in `/assets`, the build script automatically copy them into public
* All production assets (except fonts and images) has content identified hashes. The hashes only change if the content of the files change so in most of the time the biggest js file `vendor` is remaining unchanged and cache for users
* The HTML for production is automatically generated to include scripts with hashes.

adcpm:
Can you make sure it works on Heroku? I think some dep are missing. On local i have an error on yarn install: `Error: Cannot find module 'co'`. Also on Heroku build there is this error: `Error: Cannot find module 'json-stable-stringify'`. Even if i add these dep i've not been able to make the site working.

adcpm:
It still not working for me or on Heroku, please add "waiting-review" label when it's ready.

p0o:
@adcpm I know I was working on it the first problem solved due to a yarn bug issue and second one with upgrading node-sass. You can review the production but I still need to fix some issues for dev environment.

By the way since the changes in github is not clear since I moved a lot of files please use the filter in top to view `.js` and `.scss` files separately (not sure if there is any better way?)

adcpm:
Cool Heroku deployment it's working fine now. On local i still have this error `Error: Failed to lookup view "error" in views directory "/Users/fabien/WebstormProjects/busy/views"`.

adcpm:
What is the reason of moving from handlebar to ejs for view engine?

p0o:
@adcpm because we don't need templating and most of the backend is basically loading simple html files. In the other hand `html-webpack-plugin` works with simple html and doesn't work with handlebar templates. So `ejs` is used as default to render html and answer the future needs of templating.

adcpm:
It seem to not work when open website not from homepage see here: https://busy6-pr-411.herokuapp.com/hot

p0o:
@adcpm should be fixed for both dev and production env

adcpm:
Cool i confirm that the last issue is solved. Now there is another issue about the style. I think it's the import order which is wrong. App styles should overwrite dependancies styles. 
image

PR #417, opened 2017-03-18
Fix issue with Youtube thumbs

adcpm:
Some Youtube thumb was not working on Busy, i've changed the thumb link to a better one on Embed.js and updated to v0.0.4


adcpm:
There was still an issue to solve, will do in another PR.

PR #418, opened 2017-03-19
Sort Items in post feed card

nil1511:


adcpm:
Great work @nil1511, it's working well for me. Do you have concern @p0o about the implementation? Maybe we could have the logic on a separated js file?

adcpm:
Something i notice: when there is a title then a preview text, the space between these 2 element will be much bigger than usual. 

Before: 
image

After:
image

I think there should not have 2 times the `PostFeedCard__cell--body` element in this case.

nil1511:
@adcpm Updated the styles
@p0o Thank you for the suggestion. I have updated the code. let me know if more improvement is needed.

src/helpers/postFeedHelpers.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
```
p0o: It's better to keep this file in the related directory. If it's a helper that is going to be used only for PostFeedCard it's better to be named `PostFeedCardHelper.js` and be saved alongside the component. This folder better to be kept only for general helpers.

src/helpers/postFeedHelpers.js
```diff
@@ -0,0 +1,97 @@
...
+
+const postWithAnEmbed = (tagPositions, searchPostion) => {
+  const result = (tagPositions.embed && tagPositions.embed < searchPostion);
+  if (tagPositions.image !== undefined) {
+    return (tagPositions.image > tagPositions.embed) && result;
+  }
+  return result;
+};
+
+const postStartsWithAPicture = tagPositions => postWithPicture(tagPositions, STARTWITHPERCENT);
```
p0o: sorry for wrong hints before in my review, I think since these functions are returning boolean should be named like: `isPostStartsWith...`

src/helpers/postFeedHelpers.js
```diff
@@ -0,0 +1,97 @@
...
+const postStartsWithAnEmbed = tagPositions => postWithAnEmbed(tagPositions, STARTWITHPERCENT);
+const postWithPictureBeforeFirstHalf = tagPositions => postWithPicture(tagPositions, 50);
+const postWithEmbedBeforeFirstHalf = tagPositions => postWithAnEmbed(tagPositions, 50);
+
+// Case 1: Post start with a picture
```
p0o: I think there is no need for comments anymore

src/helpers/postFeedHelpers.js
```diff
@@ -0,0 +1,97 @@
...
+    image: () => imagePath && <div key="image" className="PostFeedCard__thumbs">
+      <PostModalLink post={post} onClick={() => openPostModal(post.id)}>
+        <img alt="post" key={imagePath} src={imagePath} />
+      </PostModalLink>
+    </div>
+  };
+  const tagPositions = getPositions(htmlBody);
+  const feedBody = [];
+
+  if (postStartsWithAPicture(tagPositions)) { // case 1
```
p0o: IMO this implementation is templating and should be done in the component not helper. Helper should only check the conditions and if conditions met all the elements of post should be assembled in the component.

PR #420, opened 2017-03-19
Json parse

### LINKED ISSUE #419: Json metadata with twice stringify make post feed links fail
#### Actual behavior

I can't open a post modal if the json_metadata of the post has been parsed twice.

#### How to reproduce

Go here: https://busy.org/hot/steempoll and click on the first story.

adcpm:
Init formatter.jsonParse method and allow string that have been parsed twice.

Close #419

### DIFF
--- a/src/helpers/apiHelpers.js
+++ b/src/helpers/apiHelpers.js
@@ -1,6 +1,6 @@
 import Promise from 'bluebird';
-
 import steemAPI from '../steemAPI';
+import { jsonParse } from '../helpers/formatter';
 
 /** *
  * Get the path from URL and the API object of steem and return the correct API call based on path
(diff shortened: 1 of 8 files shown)


adcpm:
Thx @nil1511 it's better you're way. I've done the changes.

p0o:
@adcpm did you consider using something [like this](https://www.npmjs.com/package/parse-json), I'm not exactly sure but I think having the `try` in another file will throw error from the wrong place which is making debugging harder.

nil1511:
@p0o Does it handle over stringified data?

p0o:
@nil1511 No I don't think so it was just a suggestion for throwing error better but I think practically it works just like `JSON/parse()`, I don't have a strong preference over this so if you are okay with this approach just go ahead and merge it.

PR #449, opened 2017-04-03
Server side rendering

nil1511:
Currently, It is doing server-side rendering for post only. For feeds and users page actions need to changes to be promisable and work with route. For other pages, this can be broken into different PR.


webpack/webpack.config.js
```diff
@@ -0,0 +1,5 @@
+
+module.exports = require('./makeConfig')({
+  isDevelopmen: true
```
adcpm: `isDevelopmen` miss the `t` in the end

nil1511:
@adcpm can you check now? 
Sidebar is not loading because it is not using redux state. 
Server rendering is working only post as of now So we should not remove title and description meta.

adcpm:
I try to share on Facebook and findout it doesnt link to busy but to steemit. Issue is `og:url`, it should always be busy.org, it's not same as canonical.

adcpm:
Also can you add " - Busy" on the end of article title?

adcpm:
Hum. For some reason, sharing on Facebook would not show image & description anymore. 
image
 Strange because that was not the case on last deployment. Something that may cause this is that we have actually 2 <title> tag:
image

adcpm:
Maybe Facebook create image and description based on the URL of og:url.

nil1511:
yes that can be.
should we try with pr app url?

adcpm:
You can make that `og:url` use the current url origin.

nil1511:
Updated. Now it is working.

nil1511:
The method of setting origin now(using global)  is not good. we should eventually switch to env variable in future.

PR #451, opened 2017-04-04
Popover enhancement

p0o:
add right and bottom-left view features to popover


p0o:
@adcpm what is up with these repetitive build failures which happen after irrelevant changes? Yarn maybe?

adcpm:
@p0o I think builds failing are on the PR for server side.

PR #460, opened 2017-04-11
Reply button

### LINKED ISSUE #443: Single Post: Add scroll and auto focus when click on reply icon
#### Actual behavior

Actually when we click on reply icon on post single page or modal, nothing will happen.

#### Expected behavior

When click on reply icon there should be a scroll down to comment form, and an auto focus. Similar to medium: autofocus

#### How to reproduce

Open an article, click reply icon button.

p0o:
Should fix #443 and also auto focus comment forms after clicking on reply to comments.

### DIFF
--- a/src/app/Menu/MenuPost.js
+++ b/src/app/Menu/MenuPost.js
@@ -2,6 +2,7 @@ import React from 'react';
 import numeral from 'numeral';
 import { FormattedMessage } from 'react-intl';
 import { Link } from 'react-router';
+import { animateScroll } from 'react-scroll';
 import { SimpleTooltipOrigin } from '../../widgets/tooltip/SimpleTooltip';
 import { getUpvotes, getDownvotes, sortVotes } from '../../helpers/voteHelpers';
 import Icon from '../../widgets/Icon';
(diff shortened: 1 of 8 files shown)


src/app/Menu/MenuPost.js
```diff
@@ -11,15 +12,30 @@ const MenuPost = ({
...
   unlikePost,
   dislikePost,
   content,
   onEdit,
+  modal
 }) => {
+  const handleReplyButtonClick = (e) => {
+    e.stopPropagation();
+    if (modal) {
+      animateScroll.scrollMore(400, { containerId: 'busyModal' });
```
adcpm: The scroll position should rely on comment form position and not a defined value. Actually when i click comment multiple time the scroll will keep increase to 400. Are we sure we want to add a new dependancy and complexity in the code for a smooth effect? We could use hashtag link instead and no worry about this.

p0o: @adcpm right I can calculate the exact position but we cannot have smooth scroll with simple anchor (hashtag) links. It's a good library and can do many things for us including go back to top/bottom features in timeline.

PR #477, opened 2017-05-06
Added DefaultMessage for missing Messages

nil1511:


p0o:
@nil1511 There is a small conflict for the latest merges I guess.

PR #479, opened 2017-05-08
Activity Notification UI

p0o:
This PR is adding the new full featured UI notification on top of Busy platform. This feature is a complementary feature to our browser notification feature.


adcpm:
- If i click on a notification, i got redirected to the related comment but then, if i click to another notification, i will stay on the same page (the previous notification).
- Counter should not be visible when "0"
- When click notification icon the popover is open, when i click again on notification icon the popover should close. Actually its closing then opening another time if i click.
- On the page `/notifications` :
  - Content frame should not be full-size, there should be a `container` div.
  - The notification description should not be limited to `250px` width

PR #480, opened 2017-05-08
do not subscribe to push if already subscribed

p0o:


src/auth/pushpadHelper.js
```diff
@@ -8,62 +8,70 @@ export const initPushpad = (uid, username, token) => {
...
-            username,
-            token,
-          }),
-        })
+        const { signatureId } = res;
+
+        pushpad('init', PUSHPAD_PROJECT_ID);
+        pushpad('uid', `${uid}`, signatureId);
+        pushpad('subscribe', (isSubscribed) => {
+          if (!isSubscribed) {
```
nil1511: Can this condition be true?

p0o: @nil1511 what you mean?

nil1511: Signature is fetched only when user is not subscribed( inferred from line 25). So I thought isSubscribed should be always false here.
If isSubscribed is always false how will it go to line 57.

variable should be named differently If there purpose is different.

p0o: I understand what you mean but it doesn't work like that. signature is fetched if user is not subscribed so after that pushpad will ask user to accept the permission, if user accept this time, the callback will return true so the code will be executed in line 57.

nil1511: Cool. Can you change the variable name to like permissionAccepted or something like that. It is confusing using same variable name.

p0o: It's the default name from pushpad API and the variable is the same and not different meaning. The thing is the first time we want to make sure to don't repeat the process for already subscribed users and second time if user subscribed we want to send something to our server.

PR #486, opened 2017-05-10
Using Steemconnect v2 [Do not merge]

nil1511:
- [x] Login
- [x] Vote
- [x] Comment
- [x] Reblog
- [x] Post
- [x] Follow/Unfollow


adcpm:
Can you make that the server check for authentication? 
Also `https://v2.steemconnect.com` should be the default `STEEMCONNECT_HOST`
and `https://busy.org/callback` default `STEEMCONNECT_REDIRECT_URL`

PR #508, opened 2017-06-01
Customize antd

Sekhmet:
This PR will focus only on customizing Ant Design just by overriding less variables. Inconsistences between antd and Busy design will be resolved later with custom classes.

### DIFF
--- a/src/styles/custom.less
+++ b/src/styles/custom.less
@@ -21,7 +21,10 @@
 
 // Buttons
 @btn-height-base: 30px;
+@btn-height-sm: 25px;
+
 @btn-font-weight: 600;
 
 @btn-primary-color: @white-darker;
 @btn-primary-bg: @purple;
+
--- a/src/styles/custom.less
+++ b/src/styles/custom.less
@@ -24,7 +24,8 @@
 
 @body-background: @white-darker;
 @component-background: @white-darker;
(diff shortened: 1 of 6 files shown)


src/styles/custom.less
```diff
@@ -12,6 +13,25 @@
...
 
+@primary-color: @purple;
+
 @font-family-sans-serif: "Whitney", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
 @font-family-monospace:  "WhitneyBold", Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
 @font-family-base:       @font-family-sans-serif;
+
+@body-background: @white-darker;
+@component-background: @white-darker;
+@font-size-base: 14px;
```
adcpm: Font-size should be 16px by default, we dont use 14px font-size on the new-design.

src/styles/custom.less
```diff
@@ -12,6 +13,25 @@
 @pink:   #ff5b77;
 @purple: #4757b2;
```
adcpm: You can add @green too with `#54d2a0` (used in topic button)

Sekhmet: This color is already there (GitHub ommited it in diff).

PR #510, opened 2017-06-01
Add StartNow and InterestingPeople components

Sekhmet:
**StartNow** component:
image
**InterestingPeople** component:
image


src/components/Avatar.js
```diff
@@ -1,12 +1,16 @@
...
 const Avatar = ({ username, size = 34 }) =>
-  <span className={`Avatar ${size}`}>
-    <img
-      key={username}
-      className="rounded"
-      src={`https://img.steemconnect.com/@${username}?s=${size}`}
-    />
-  </span>;
+  <img
+    className={`Avatar Avatar--${size}`}
```
adcpm: For the size we can do like this:
```
style={{ width: `${size}px`, height: `${size}px` }}
```

src/components/Avatar.less
```diff
@@ -0,0 +1,9 @@
+.Avatar {
+  display: block;
+  border-radius: 8px;
```
adcpm: And here we can change `border-radius: 20%`, so it will adapt on any avatar size.

PR #513, opened 2017-06-02
Topics

Sekhmet:
**Favorite topics:**
image
**Trending topics:**
image


adcpm:
Can we have a padding left and right of `10px` for Topic button? Overwise it look's very good!

PR #514, opened 2017-06-02
Use iconfont for icons instead of fontello

Sekhmet:
Customizing antd with (according to [this](https://github.com/vision8tech/dvajs-user-dashboard/pull/2) quotes should be other way around, but that way it was inserting double quotes into path so it was resulting in 404):
```less
@icon-url: "'../fonts/iconfont'";
```
doesn't include custom glyphs inside typeface (setting custom icon's code i.e. `&#xe735;` resulted in empty icon). Decided to use custom classes for it (`<i className="fonticon .icon-*" />`). One might want to play around with using custom iconfont for antd later.

**Sidenav** component has been updated to use iconfont. Comment if you think that icon should be replaced with something else.
image


Sekhmet:
Added search icon to **Topnav**:
image

adcpm:
Header should have only bottom border.

Sekhmet:
This is how it looks now:
image
image
(Colors look messed up here, click on them to see full resolution ones).

src/components/Navigation/Topnav.less
```diff
@@ -4,71 +4,95 @@
...
-}
+  justify-content: flex-end;
 
-.Topnav__menu {
-  flex-shrink: 0;
-  height: 55px !important;
-  line-height: 55px !important;
-  margin-left: @base-spacing;
-}
+  & > ul {
```
p0o: As a best practice it's always better to not style basic html elements in the stylesheet so later on scale we shouldn't override it. If it's an element we use we better give it a classname.

src/styles/iconfont.less
```diff
@@ -0,0 +1,353 @@
+@font-face {font-family: "iconfont";
```
p0o: We use `scss` not `less` please consider using that format for consistency.

Sekhmet: On new-design branch we migrated to antd and we use less there.

src/components/Navigation/Topnav.less
Sekhmet: This selector matches only children of type ul of `.Topnav__menu-container` so I don't think it will be a problem.

adcpm:
Size of this should be 190px
image

src/components/Navigation/Topnav.less
p0o: As I said it's a best practice and the problem will happen if three months later someone wants to add another `ul` to `.Topnav__menu-container`. It's more clean to keep it with searchable names unless it's an external component and hard for you to do the naming?

PR #524, opened 2017-06-06
Add UserHeader

Sekhmet:
Updates #520 

**UserHeader:**
image

PR #525, opened 2017-06-06
Add UserMenu component

Sekhmet:
Updates #521 
**UserMenu:**
image


adcpm:
There is too much space between the element of the menu. Also the purple border bottom should start and end at the same level of the text like this:
image

adcpm:
Also height of the menu should be 46px and text in bold

Sekhmet:
image
This is how it looks now (here it is with bold text, #524 adds fixes for font-weights, so after merging it items will be bold right away).

PR #526, opened 2017-06-06
Add Stories

Sekhmet:
**Story** component:
image


adcpm:
On the top right arrow you can do a popover with this in menu:
- Follow/unfollow guest123
- Report post
- Save post
Similar to FB
image

adcpm:
Great work so far!
Here some inputs:
- Can we have tooltip on these buttons:
image With "Like", "Dislike", "Comment" and "Reblog"
- Avatars should link to user profile
- Inline story should not have full content by default. It should follow a logic that has been implemented by @nil1511 here https://github.com/busyorg/busy/blob/828dc4b57b5ea4343b91c83d92838923bdebc208/src/post/Feed/PostFeedCard.js#L88-L123 . Its probably better to separate this logic into a PreviewStory component.

nil1511:
It can just render it children's and children will contain text and image in the needed order.

adcpm:
Looks good! Just for full story, can you move the topics between `StoryFull__content` and `StoryFooter` like this:
image

PR #529, opened 2017-06-06
Add Notifications

Sekhmet:
**Notifications** component:
image


adcpm:
I've added more data for notification but I've not been able to test the scrollbar. Can you make take we can see max 5 stories and if there is more, scrollbar appear on mouse over notification popover.

Sekhmet:
peek 2017-06-07 14-04

adcpm:
Can we have also the tooltip "Notification" on mouse over.
image
And for the top right button of logged header we can have a popover with this in menu:
- Activity
- Settings
- Logout

Sekhmet:
peek 2017-06-07 14-40

PR #541, opened 2017-06-13
Update button styles and behavior

Sekhmet:
**Topic** button:
peek 2017-06-13 11-10
**Follow** button:
peek 2017-06-13 11-11


adcpm:
Can we have unfollow button in red with `type="danger"` https://ant.design/components/button/

PR #543, opened 2017-06-13
Add editor

Sekhmet:
**Editor** component:
image


Sekhmet:
**Editor** component:
image

Topic selector:
image
image

Preview:
image

adcpm:
Some inputs:
- Move editor toolbar between [Editor / Preview] tabs and editor
- Can we have the preview rendered with `Body` component? To make sure we have same render on interface and editor preview.
- There should be a validator for the editor content input to not be empty.
- Topic should have same style than topic on interface.

Sekhmet:
image

adcpm:
- Topic name should be automatically lowercased. On mobile it give uppercase by default can be annoying. Also there should be only alpha numeric or single dash "-" character for topic name, it should be part of the validation.
- Title should have max length of 255 characters.

Sekhmet:
I changed everything except for forcing lower case input on mobile. I was looking for a way to access native input to disable auto capitalization, but looks like this [is not possible](https://github.com/ant-design/ant-design/issues/1911). `RcSelect` that `antd` uses in `Select` makes it possible to get `input` element by calling `getInputDOMNode()`, but there is no way of accessing `RcSelect` because there is no ref to it (as oppoosed to refs to `input` in `Input`). I'm not sure whether there is a elegant solution to this problem. I'm afraid it's not possible to do with current state of `antd`. Any ideas about overcoming this problem?

https://github.com/react-component/select/blob/master/src/Select.jsx#L503
https://github.com/ant-design/ant-design/blob/master/components/select/index.tsx#L148

PR #547, opened 2017-06-16
Enable actions

nil1511:


src/components/Story/StoryLoading.less
```diff
@@ -7,6 +7,7 @@
   max-width: @width;
   border-radius: 4px;
   background-color: @white;
+  margin-top: 1em;
```
Sekhmet: StoryLoading has `margin-top` if it's not first-child already.  https://github.com/busyorg/busy/blob/677576fb7d3889f40423e1f456c957ed7b259eb7/src/components/Story/StoryLoading.less#L14
If it fits better to new Feed with just normal `margin-top` declaration we should get rid of `:not(:first-child)`.

nil1511: yes. will remove it.

PR #549, opened 2017-06-20
Apply new style to Busy

Sekhmet:


adcpm:
Great work Wiktor! Here is my few inputs: 
- The popover on feed and on single post page with "Follow username, Report post, Save post" should not have bold text. Also the popover should be aligned in bottom left of the dropdown icon, example: http://prntscr.com/fmhejr
- The right sidebar is not properly aligned on the right on single post page. http://prntscr.com/fmhf7n
- On the component "InterestingPeople" link to username is broken, there should be a `@` prefix.
- Topics button not appear like it was on storybook, there is a nested class that give grey border while it should be green.
- "Start Now" button should not be hidden when mouse over.
- There should be a margin bottom on StoryLoading, for not having the component touch the end of the page http://prntscr.com/fmhhm3
- Logo or link should not have underline text decoration http://prntscr.com/fmhiku
- On the profile page, the first name we see should be the `account.json_metadata.profile.name` when available. http://prntscr.com/fmhkwo >> http://prntscr.com/fmhkex

adcpm:
- Right side bar should always have width of `260px`, on small screen its become very small http://prntscr.com/fmyj52
- Can we have sticky sidebars?

Sekhmet:
I implemented above features but I'm adding WIP label as I have to create component for it to allow hybrid scrolling (if sidebar is higher than screen height it should not be fixed until scrolled to the bottom and then become fixed, so user can access bottom of the sidebars).

bonustrack:
I'm merging these changes, we can continue on sticky menu in another PR.

PR #550, opened 2017-06-21
Latest comments

itsyogesh:


adcpm:
Here my inputs:
- Author link should be in bold
- Time in grey color instead of purple
- The comment text body should use "component/story/BodyShort" component. It strip the markdown / html tags and shorten the text.
- You can add on BodyShort comp a parameter `maxLength` with default 140. and for this component set a `maxLength` of 80 characters.

PR #562, opened 2017-07-05
Add Affix sidebars

Sekhmet:


bonustrack:
The behavior is good but there is some issue around it.
- If i left click on scrollbar and move it, it's very fluid, but if i use the scrollwheel of my mouse to scroll, the sidebars are shaking see here: https://user-images.githubusercontent.com/16245250/27869776-28d7215a-61cb-11e7-9568-bc1e1a8cde07.gif
- When content is loading, the frame will be cut in middle, see here: http://prntscr.com/fryq3o
- It doesnt work on Firefox.
- It's shaking a lot and super laggy on Safari.

Sekhmet:
Two problems should be fixed now, but the rest is connected to how browsers handle rendering. [Asynchronous rendering causes it to be out of sync](https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Scroll-linked_effects) (what has been calculated is different from what user sees). Looks like it can be fixed by making sidebars fixed when reaching bottom when scrolling down and reaching top when scrolling up. This requires significant rewrite of affix logic though.

Sekhmet:
Still WIP but it **kinda-works™**.

Bugs I am aware of:
- [ ] After changing size of sidebar it doesn't have scroll  behavior right away. Should recalculate position after height change.
- [ ] After collapsing to smaller sidebar, sidebar disappears until next scroll. Should set it position after that event happens.

If someone has an idea how to achieve this behavior better I am open for comments or contributions.

PR #565, opened 2017-07-07
Trello: Polish the design 2

Sekhmet:
[Trello](https://trello.com/c/2YhWtzBr/36-polish-the-design-2)


bonustrack:
Great, here my review:

> Post title link should be different color when already :visited
- Can we have it grey instead of purple? You can use the same color than sidebar left text color.

> Play button should be white and not be animated. And there should be a 20% opacity background with primary color (purple) over video preview
- Actually 20% is too low, you can change to 60% opacity.

> Use white background on single post
- The complete page should have white background on single post, we dont want a frame around post content.

> On my own profile there shoulnt have "follow" button, but "edit profile" button
- Can this button use same color than "message" or "transfer" buttons?

PR #567, opened 2017-07-11
Updated components LatestRecommendations and LatestComments

itsyogesh:
Used key props to remove test warnings.

### DIFF
--- a/src/components/Sidebar/LatestComments.js
+++ b/src/components/Sidebar/LatestComments.js
@@ -28,13 +28,7 @@ const LatestComments = ({ comments }) => (
     <div className="LatestComments__container">
       <h4 className="LatestComments__title"><i className="iconfont icon-time LatestComments__icon" /> Latest Comments</h4>
       <div className="LatestComments__divider" />
-      {comments && comments.map(comment => (<RenderComment
-        key={comment.id}
-        text={comment.text}
-        author={comment.author}
-        created={comment.created}
-      />
-      ))}
+      {comments && comments.map(comment => (<RenderComment key={comment.id} {...comment} />))}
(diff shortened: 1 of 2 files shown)


src/components/Sidebar/LatestComments.js
```diff
@@ -28,7 +28,13 @@ const LatestComments = ({ comments }) => (
...
       <h4 className="LatestComments__title"><i className="iconfont icon-time LatestComments__icon" /> Latest Comments</h4>
       <div className="LatestComments__divider" />
-      {comments && comments.map(comment => renderComment(comment))}
+      {comments && comments.map(comment => (<RenderComment
+        key={comment.id}
+        text={comment.text}
+        author={comment.author}
+        created={comment.created}
+      />
+      ))}
```
bonustrack: Intend look not good. These 2 should not be in same level:
```
/>
))}
```

You can do like this: 
```
{comments && comments.map(comment => (
  <RenderComment key={comment.id} {...comment} />
))}
```

src/components/Sidebar/LatestRecommendations.js
```diff
@@ -31,7 +31,14 @@ const LatestRecommendations = ({ posts }) => (
     <div className="LatestRecommendations__container">
       <h4 className="LatestRecommendations__title"><i className="iconfont icon-flashlight_fill LatestRecommendations__icon" /> Latest Recommendations</h4>
       <div className="LatestRecommendations__divider" />
-      {posts && posts.map(post => renderRecommendedPost(post))}
+      {posts && posts.map(post => (<RenderRecommendedPost
```
bonustrack: Here too

PR #569, opened 2017-07-12
Add webpack-dev-server for development

Sekhmet:
This PR adds `webpack-dev-server` as a potential replacement for `dev` script (which is broken and not suitable for development).

This PR adds 3 scripts:
1. `dev-server:auth-server` Express.js app that is used for handling `/callback` request
2. `dev-server:client` starts `webpack-dev-server`
3. `dev-server` runs scripts above in parallel.

This removes server bottleneck (100 seconds versus 180) and doesn't require tinkering when starting server (running `yarn run dev` and `yarn run start` to do full rebuild and start properly working server which takes around 290 seconds in total).

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Busy</title>
+    <meta charset="UTF-8">
+    <meta name="description" content="Busy is a decentralized social network based on Steem blockchain">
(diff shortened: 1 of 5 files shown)


bonustrack:
Look's good to me, should we remove the `dev` script then and related files?

Sekhmet:
Not sure about removing `dev` script. We still need `webpack.server.config.js` for production builds, so we could only remove two scripts from package.json: `dev` and `build:server:dev`.
They are related to server side rendering and may be useful even though they are broken.
I'd leave them both for now, maybe renaming `start` and `dev` scripts to something else to make it clear that it's broken and meant for working with server side rendering only.

PR #573, opened 2017-07-13
Integrate editor

Sekhmet:


bonustrack:
I have some trouble runing it on local and Heroku deployment failing (https://busy5-pr-573.herokuapp.com/). Can you try to reset your node_module and see if you got same issue than me? Here is what i get:
image

Sekhmet:
You probably forgot to run `yarn install`.
I fixed the issue on heroku. Had to switch to memory history on non-browser environments.

bonustrack:
I still need to test it fully but for now i have some small inputs:
- Seem like we lost bold text on username and on feed post title, probably because of `antd` update.
  Before: https://user-images.githubusercontent.com/16245250/28360615-34e162f0-6ca0-11e7-9c90-a8e95a228d1b.png
  After: https://user-images.githubusercontent.com/16245250/28360631-3f2bf0f4-6ca0-11e7-815e-dbfed0625994.png 
  Can we have bold back?
- For topics, we dont need autocomplete, it feel more confusing than a real help actually.
- We can remove the "*" (required) on Title and Topics and "Write your story" label.

bonustrack:
On localhost when i try to post i will get an error from the blockchain but nothing will appear on the interface. The error look like this: 
image
This come from the field `json_metadata`. I can see in error detail that it's a json object while it should be a stringified object. If i stringify the `jsonMetadata` var it's working fine.

bonustrack:
Forget my last review, i had not updated `node_module` folder properly. It's now working fine in local.

bonustrack:
Some suggestions:
- Can we have a sidebar right? We would add later some guideline to help post author.
- Can we have the button submit with the label "Post", with fixed width and aligned in right?
- In the same level of the button can we have "Mardown support" indication? Similar to GitHub 
image

mynameisek:
- markdown preview quote ( >) and code (``) does not work

- i suggest we also had (like steemit) the Rewards option (3): 
1. 100% Power up
2. 50%/50
3. Declined

- seems like it has some line break issue when type Enter once, it will not line break in many cases

bonustrack:
Reward option require changes on SC2 SDK we can work on this later.

Sekhmet:
Looks like antd disables default styles of lists so I added styles for inline and multiline code, quotes and lists.
In markdown to add new paragraph you have to add one empty line between text, like this:
image

bonustrack:
Really great work! Latest changes in the style then i think we good:
- Add loading icon inside button when submit is processing
- Change `Create new post` button label to `Post`

PR #583, opened 2017-07-20
Check if containers are not undefined

Sekhmet:
When sidebars have `display: none` property containers are `undefined` causing Affix to throw errors ([reference](https://busy5-pr-506.herokuapp.com/))


bonustrack:
Which error exactly you trying to solve? The one i see for Affix.js is this:
image
But this PR not seem to fix that, i still can see this error here.

Sekhmet:
Right, I missed one check. Should work now.

PR #584, opened 2017-07-21
Integrate follow and post actions

Sekhmet:


src/components/Story/Story.js
```diff
@@ -55,12 +77,12 @@ class Story extends React.Component {
...
-              <PopoverMenuItem key="follow">
-                <i className="iconfont icon-people" />
-                {' '}{!postState.userFollowed ? 'Follow' : 'Unfollow'} {post.author}
+              <PopoverMenuItem key="follow" disabled={pendingFollow}>
+                {(pendingFollow) ? <Icon type="loading" /> : <i className="iconfont icon-people" />}
+                {`${followText} ${post.author}`}
               </PopoverMenuItem>
               <PopoverMenuItem key="save">
-                <i className="iconfont icon-collection" /> Save post
+                <i className="iconfont icon-collection" /> {(postState.isSaved) ? 'Remove from bookmarks' : 'Save post'}
```
bonustrack: We can change the label `Remove from bookmarks` to `Unsave post`

Sekhmet:
This PR adds lazy update to like, follow, reblog and save actions.
There is problem with SteemConnect right now and requests fail most of the time. I think there is no point in implementing error handling for this, because we have to migrate to new version of `sc2-sdk` and we will have to rewrite it anyway.

bonustrack:
Can you link "Follow" button from profile too?

Sekhmet:
I wanted to create two PR for that Trello task as this one I already called `post-actions`.  Not sure which one makes more sense: to include profile action in this PR or create two separate PR and merge them one by one.

bonustrack:
Both are fine for me. I've created a Trello card for error handling: https://trello.com/c/tKBTHDEL/69-handle-error-and-implement-antd-notification

Sekhmet:
I implemented follow action on profile in this PR. I will rename it for clarity.

src/components/Button/Follow.less
```diff
@@ -8,10 +8,13 @@
...
+    margin-right: 12px;
+  }
 }
 
 .Follow--danger {
-  .btn;
-  .btn-default;
-  .btn-danger;
+  color: #fff !important;
+  .colorize-button(#f04134, #f04134);
```
bonustrack: Can we have the same red than the one used on topic close? We can setup this color as default for `@red` aswell so it's easy to reuse.

PR #592, opened 2017-07-28
Refactor feed

Sekhmet:


bonustrack:
- Can you remove sort by "Comments", "Cashout" and "Children"
- Fix "Promoted" sorting
- Remove filters & sort by on "Feed" page.

bonustrack:
- When i'm logged in and go on the homepage, then click on "News", posts will not load and just show error: `Oops! This feed empty, here's a link to the home page.`. This should not be the case.
- Promoted sorting still not working for me
- When unlogged and i stay on the homepage, the filter should be visible.

PR #593, opened 2017-07-28
Windows compatibility

gregory-latinier:


src/app/Sidebar.js
```diff
@@ -115,7 +115,7 @@ export default class Sidebar extends Component {
     const { auth, app, hideSidebar } = this.props;
     const { user } = auth;
     const { rate, sidebarLoading, props = {}, categories } = app;
-
+    console.log(formatter);
```
bonustrack: Can you remove this?

bonustrack:
The yarn.lock file should be udpated

bonustrack:
Working fine for me in localhost but Heroku deployment fail with this error `sh: 1: gulp: not found`.

.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   },
   "extends": "airbnb",
   "rules": {
+    "linebreak-style": 0,
```
Sekhmet: Can we delete this line? We don't want to have multiple different line endings in our code. Just force users to use LF. They can configure their environment to use those.

gregory-latinier: Yes you're right.
I'll change it on all project and configure my IDE accordingly

bonustrack:
The deployment is now failing on Heroku because missing the dependency "del". `Error: Cannot find module 'del'`

bonustrack:
Working 👍

PR #598, opened 2017-08-01
Add max-height of 700px to StoryPreview image

Sekhmet:


Sekhmet:
I missed two properties earlier. This will fix the issue with overflowing image.

PR #600, opened 2017-08-03
Fix image upload

bonustrack:


bonustrack:
Due to the changes of the solution for avatar, i've redirected the links from img.steemconnect.com to a Cloudinary solution instead of AWS Lambda. This broken the image upload which still using AWS, this PR change the upload link to fix upload.

PR #602, opened 2017-08-07
Add location and webiste to User sidebar

Sekhmet:


bonustrack:
Look good, just for the order: `Joined` should appear in last.

PR #605, opened 2017-08-08
Add reputation score

Sekhmet:


bonustrack:
Can we have the reputation score on comments and single post header too?

bonustrack:
- Can we have the tooltip on comment reputation too?
- And this space is not required on UserHeader: 
image , tag has already a margin

Sekhmet:
This blank space is part of user's name:
image

PR #606, opened 2017-08-08
Add image upload

Sekhmet:


bonustrack:
- Pasting from clipboard should also work for comment.
- Would be nice to have a picture icon in the footer of editor and comment box:
image
- We can add the text "Select image, or pasting from the clipboard." next to picture icon.
- And when picture is uploading, change the picture icon to [Loading icon] + "Uploading your image"

PR #613, opened 2017-08-15
Add voting power slider to single post view

blueyedgeek:
### What does this PR do?

It adds a voting power slider to the comment view when the `thumbs up` icon is hovered on

### How can I manually test this?

Open a post modal and hover on the thumbs up icon, a voting power slider will appear if the user has enough `steem` power.

### Image(s)

screen shot 2017-08-15 at 7 19 59 pm

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,0 +1,5 @@
+const LIKE_BAR_MIN_POWER = 125;
+
+export default {
+  LIKE_BAR_MIN_POWER
+};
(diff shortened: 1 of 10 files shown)


src/constants.js
```diff
@@ -0,0 +1,5 @@
+const LIKE_BAR_MIN_POWER = 125;
```
bonustrack: Instead of creating a new file we can add this constant into the root file `config.json`.

Sekhmet: We are using `config` package on `new-design` so it might be good idea to continue doing this on `dev` as well.

PR #617, opened 2017-08-16
Use same sort algorithm as Steemit for sorting comments close #428

### LINKED ISSUE #428: Trending comment listing does not take downvotes into account
#### Expected behavior

Comments with less payout, pending or not, should be listed below comments with more payout.

#### Actual behavior

It seems that the trending sort view of comments are ranked by upvotes only, disregarding downvotes.

#### How to reproduce

See kingscrown's comment here:
https://busy.org/steem/@pfunk/arguments-for-keeping-the-steem-reward-pool-whole

It was the top pending payout comment until my downvote, and displays in the proper position on steemit.com. However on Busy, it is displayed at the top.

* Browser: Firefox 52.0.1

Sekhmet:


bonustrack:
I've updated the Trello card, we need a different sorting for "Best" https://trello.com/c/ksOQA7Xs/92-comment-sorting-does-not-count-downvotes

Sekhmet:
This follows Steemit rules for sorting, but won't be 1:1 all the time (it uses `net_rshares` and `totalPayout` for sorting, but if they are equal it might be random - based on the order of  unsorted comments).

PR #620, opened 2017-08-17
Add reactions modal

Sekhmet:


bonustrack:
- Dislike tab or like tab should appear only if it's not empty
- Remove buttons "Cancel" and "Ok"
- Don't show scrollbar by default, show scrollbar only when reaction list is bigger than max height modal
- Vote should be sorted by `rshares` bigger to smaller, and downvote by `rshares` smaller to bigger, this way we see the vote with biggest impact first.

bonustrack:
- Also there should be a tooltip with the list of 20 most powerful vote like this:
image

PR #621, opened 2017-08-18
Update UserLoading screen

Sekhmet:


bonustrack:
Look great. Just it would be nice if the header loading fit exactly the header loaded. So the frame not moving after being loaded. See difference here: 
image

PR #624, opened 2017-08-18
Add cover

bonustrack:
Add cover on user header
image


Sekhmet:
I think that we should add contrast shadow to text to prevent blending text to image (white for black text and vice versa).

PR #637, opened 2017-08-22
Add FollowButton

Sekhmet:
- Create `FollowButton`
- Use `FollowButton` in `UserCard`
- Use `FollowButton` in `UserHeader`

`FollowButton` is container component that wraps `Follow` component with logic. It's used in `UserCard` and `UserHeader` and should be used for every place where we need Follow button. It doesn't need anything else than `username` so it should work well without any tinkering.

**Related Trello task:** https://trello.com/c/zD170Bjp/115-enable-follow-buttons-user-card

### DIFF
--- a/src/widgets/FollowButton.js
+++ b/src/widgets/FollowButton.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import {
+  getAuthenticatedUserName,
   getFollowingList,
   getPendingFollows,
 } from '../reducers';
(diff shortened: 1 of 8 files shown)


bonustrack:
Excellent, it will be easy to integrate where-ever we want the button. Can we just:
- Not showing this button i'm logged with the same username
- Add the button in "Interesting People"

src/widgets/FollowButton.js
```diff
@@ -0,0 +1,59 @@
...
+    authenticatedUserName: PropTypes.string,
+    followingList: PropTypes.arrayOf(PropTypes.string).isRequired,
+    pendingFollows: PropTypes.arrayOf(PropTypes.string).isRequired,
+    followUser: PropTypes.func,
+    unfollowUser: PropTypes.func,
+  };
+
+  static defaultProps = {
+    authenticatedUserName: undefined,
+    getAccountWithFollowingCount: () => {},
```
bonustrack: Is this been used?

Sekhmet: Looks like I've copied too much text.

PR #639, opened 2017-08-23
Nd/update icons

bonustrack:
- Update iconfont icons
- Remove Material icons

image
image
image


Sekhmet:
It looks great now! However, is this possible to rename `icon-danye-` to `icon-bold` and `icon-daifacaigouxuqiu` to `icon-write`? It would make it easier to use in other places without having to look for its name?

PR #642, opened 2017-08-23
Move Post styles to Body

Sekhmet:
`Body` component now has `full` prop that is used to specify if it should render post in full style (like in `Post`) or not (line in `Comment`).


bonustrack:
There is still a difference how new line are handled on preview and on real post, the preview dont do linebreak while single post render do:
editor

Sekhmet:
Now `Body` exports `remarkable` parser and it is used inside `Editor` to make output the same.

PR #645, opened 2017-08-24
Integrate CrowdIn translations

Sekhmet:
https://trello.com/c/i0f6X4wO/123-remove-plain-text-and-add-translation-files


Sekhmet:
I'm not sure about how CrowdIn works, but when exporting it creates separate folder for each language and that folder contains all files added to CrowdIn.
I wanted to split translation into different objects, but I don't think CrowdIn can handle it. CrowdIn works best with CSV files - with JSON we cannot add new strings via interface - it's CSV only feature. Exporting translations happens in the same format as source files (so if we use CSV for source file it will export all CSV files).
Not sure if it's issue with GitHub, CrowdIn or some kind of misconfiguration but GitHub sync doesn't work (Sync Issue).

Sekhmet:
Closing this because CrowdIn doesn't support `/` in branch name.
Use #646

PR #650, opened 2017-08-25
Show voting power bar in comments if enabled for user

blueyedgeek:
### What does this PR do?

Adds voting power slider to user comments on a post

### Relevant trello link

https://trello.com/c/N5oZYBHo/6-voting-power-slider-on-comment

### DIFF
--- a/src/comments/CommentItem.js
+++ b/src/comments/CommentItem.js
@@ -3,6 +3,7 @@ import { Link, withRouter } from 'react-router';
 import { FormattedRelative } from 'react-intl';
 import numeral from 'numeral';
 import _ from 'lodash';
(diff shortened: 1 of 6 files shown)


blueyedgeek:
@bonustrack on what page(s) does this occur?

bonustrack:
It's really hard to not having the voting slider disappear when trying to move it see:
slider
This behavior is only in comment like sliders, not in root post like sliders. The implementation should be same.

blueyedgeek:
@bonustrack, the issue has been addressed, can you check again?

bonustrack:
@blueyedgeek looking better now.

PR #663, opened 2017-08-28
Add auto language setting

Sekhmet:


bonustrack:
I'm using Chrome FR and when i click in "Use browser language" it doesnt work, i'm still in english. But anyhow we dont need a link for it in settings we should remove it along with translations. By default if user not yet specified a language, the interface should be in browser language. When open https://busy-dev-pr-663.herokuapp.com/ it should be in French for me, Polish for you.

Sekhmet:
It defaults to `navigator.languages` which is list of language preferences, which don't have to be equal to browser language (`navigator.language`). I'm not sure which one we want to use.

bonustrack:
We should default to `navigator.userLanguage` then `navigator.language` then `navigator.languages` then `en`. Now it's in the reverse order.

Sekhmet:
We don't store language determined by script inside redux nor local storage. It's just the old default value that remained this way. We should limit number of settings stored inside local storage because it's nightmare to migrate from version to version with them.

bonustrack:
In future, localStorage will be used only to store sc2 `access_token`

PR #665, opened 2017-08-28
Add notifications translations

Sekhmet:


src/locales/en.json
```diff
@@ -16,6 +16,10 @@
   "bookmarks":"Bookmarks",
   "bookmarks_empty":"You don't have any story saved.",
   "notifications":"Notifications",
+  "notification_following_username": "{username} is now following you.",
+  "notification_mention_username_post": "{username} mentioned you on this post {post}.",
+  "notification_reply_username_post": "{username} replied you on your {post}.",
+  "notification_transfer_username_amount": "{username} sent you {amount} STEEM.",
```
bonustrack: `STEEM` should be changed to `{asset}`

bonustrack: Actually you can change 
`{username} sent you {amount} STEEM.`
To
`{username} sent you {amount}.`
The field "amount" on the transfer operation include asset.

PR #668, opened 2017-08-29
Update Transfers related components

Sekhmet:


Sekhmet:
image
I didn't manage to handle formatting amount in `0.000` format. `format` and `parser` props are only available for `InputNumber`, but it doesn't support `addonBefore` and `addonAfter` so we won't be able to use currency switcher embedded inside input. I also tried achieving this effect with `InputNumber` but after reformatting antd moves cursor to the end which makes typing nightmare (when trying to type `100` after every digits it will add `.000` and move cursor to the end).
I don't think it's common practice to format numerical inputs with decimal places and it may hurt UX. I checked Coinbase and looks like they are not dealing with this.
Theoretically we could create our own version of `Input` and implement it by hand, but I don't think we should as it might have negative effect on UX (for example how we should insert digits when user is typing in middle of number - should it add numbers before cursor, add them after or replace digits after cursor). I searched for some implementations of this behavior and it felt awkward and was extremely annoying.
Instead I added validators that just check if it's in correct format and if it's not gives users hint about it. Feels a lot more natural.

bonustrack:
It's fine, just few things to improve the modal:
- When i open the modal from user profile the username should be populated for the input "to"
- There should be username validator (see how is done in sc2 https://github.com/steemit/sc2/blob/dev/src/utils/validator.js#L10-L18)
- Can you change the STEEM SBD selector to a dropdown like this https://user-images.githubusercontent.com/16245250/29882124-50bd9000-8dd6-11e7-987a-ea6e67d4688d.png
- We might want to allow only number or point for the amount like this: https://ant.design/components/form/#components-form-demo-customized-form-controls
- Memo should have auto resize filed with `autosize={{ minRows: 2, maxRows: 6 }}`

Sekhmet:
I created default `to` value. Doesn't it work for you?

bonustrack:
Alright it's working on LeftSidebar but not in UserHero popover

PR #678, opened 2017-08-31
Add primary Action button for post

Sekhmet:
This adds `primary` prop to `Action` component and uses it inside Editor.
image


bonustrack:
The border of the button is not same as the others purple buttons, border is lighter than bg, but it should be darker instead.

PR #679, opened 2017-08-31
Fix comment loading

Sekhmet:
https://trello.com/c/P1PamCsn/177-user-comments-bug-pagination-infinite-loading


bonustrack:
I'm not able to load the second page of comment when user have multiple page of comment example:`/@fabien/comments`

PR #687, opened 2017-09-01
Refactor trending topics

Sekhmet:
- Switches from `trending/busy` to `getTrendingTags`
- Moves trending topics logic to redux store

https://trello.com/c/71uhCMAH/195-sidebarwithtopics-should-be-replaced-with-redux-state


bonustrack:
Loading should be aligned in left

PR #693, opened 2017-09-04
Fix image upload and markdown editing

blueyedgeek:
### What does this PR do?

- It adds in a fix for image upload in the editor that was previously not working.
- It adds in a fix for markdown shortcuts that was previously not working.
- It adds in a fix for copy paste that was previously not working.
- It adds in a fix for return key that was previously not working.

### Relevant trello link

https://trello.com/c/hItHsFKm/8-fix-editor


bonustrack:
@blueyedgeek everytime you change a dependency you must update yarn.lock otherwise deployment will fail.

PR #698, opened 2017-09-04
Fix scrolling to comments

Sekhmet:


bonustrack:
I think the scroll should not be triggered multiple time. We have an issue about this: https://trello.com/c/yV2fsD3S/159-bug-scroll-with-comment-link .

Sekhmet:
This is pretty tricky to achieve as when content is loaded and height change (i.e. images are loaded) comments are pushed off screen. Not sure how to tackle this. I was thinking about detecting user scroll and if user did scroll stop trying to scroll comments into view but `scrollIntoView` causes `onscroll` event to fire making it useless for this problem. Maybe we can use specific user related actions like `onwheel` but I don't think it's viable idea either as we will have to specify all of them (mouse wheel, scrollbars, arrow keys, touchpad, touch etc.)

Any ideas how to solve this?

cc @gregory-latinier

Sekhmet:
Problem with `componentDidMount` not triggering scroll was caused by not rendering `#comments` handle if `content` was `undefined`.
`Post` became very bloated so I created `PostContent` component, that is container of `StoryFull`. `PostContent` is only rendered if `content` is loaded so `scrollIntoView` is triggered there.
It's triggered only once, so in some cases (long post with major part of images) comments might be scrolled down (and same thing happens in Condenser).

This should fix those issues:
https://trello.com/c/yV2fsD3S/159-bug-scroll-with-comment-link
https://trello.com/c/NqRVZUMd/196-click-on-comment-icon-from-feed-redirect-to-top-of-post

bonustrack:
Seem to solve this issue too: https://trello.com/c/gjchs1fB/184-bug-scroll-when-voting-on-page-with-comment-hashtag

PR #711, opened 2017-09-07
Update table style

Sekhmet:
This adds background for header row and even body rows making it easier to read and recognize.

This is related to: https://trello.com/c/ZjJ1Og2w/225-body-render-table-head-should-have-same-alignement-than-table-rows
Body is indeed that same size as header, but it wasn't visible because of missing visual clues of table boundaries.

image


bonustrack:
This should be left aligned, similar to the table rows.
image

Sekhmet:
**Updated style**:
image

PR #715, opened 2017-09-08
Focus comment form after clicking comment icon

### LINKED ISSUE #443: Single Post: Add scroll and auto focus when click on reply icon

Sekhmet:
This focuses and scrolls to comment form after clicking comment icon on single post page.

close #443 

https://trello.com/c/IMho0urX/110-single-post-when-click-on-comment-icon-focus-to-comment-form


bonustrack:
Can we also have the focus when we click "Reply" on others comments?

PR #720, opened 2017-09-08
Limit comment indentation on mobile

Sekhmet:
- Adds `Comment__replies--no-indent` class.
- If `depth` (drawn, not logical from API) is higher or equal to 2 replies will stop indenting.
- Works for `@small` devices only.

https://trello.com/c/p21C3QGx/230-responsive-comment-intend-should-have-limit

### DIFF
--- a/src/components/Comments/Comment.less
+++ b/src/components/Comments/Comment.less
@@ -148,6 +148,11 @@
     }
   }
 
+  &__replies--no-indent {
+    position: relative;
+    left: -44px;
+  }
+
   &__icon_dislike:before {
     display: inline-block;
     transform: rotate(180deg);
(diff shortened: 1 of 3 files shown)


src/components/Comments/Comment.less
```diff
@@ -148,6 +149,14 @@
     }
   }
 
+  &__replies--no-indent {
```
bonustrack: Wouldnt be simpler to do handle the intend just with css like this `&__replies > &__replies > &__replies` ?

Sekhmet: It would match only 3rd level replies. However, there can be 4th, 5th ... nth level. Not really viable to handle this with selectors.

PR #730, opened 2017-09-12
Use SteemConnect metadata for storing user settings and data

Sekhmet:
Syncing state to local storage with `redux-localstorage` is problematic and causes problems when updating state architecture. Furthermore, local storage is device specific so draft posts and other settings won't be accessible from different devices.

Migrated to metadata
- [x] Language
- [x] Draft posts
- [x] Bookmarks

https://trello.com/c/dbDNU0YZ/254-migrate-from-localstorage-to-sc2-metadata


bonustrack:
- Unsaving a post or removing a draft is not working for me. I can see that the request to updateMetadata include all drafts when deleting one draft.
- There should not be the option "Auto" on locale select. By default the current language should be selected.
- Can you remove this alert:
image
If user dont have localStorage or have expired `access_token` we dont need the alert, we just show him unloged.
- Would be nice to have an indicator in the editor when draft is been saved similar to google drive:
image This could be positioned here:
image

Sekhmet:
@bonustrack how we should handle language select then? By default web browser determines user language. Should we save web browser language to metadata when user visits Busy? Should it happen only once, or every time user language changes?

bonustrack:
We should not save web browser language to metadata when user visit Busy, it should only be saved when user go to settings and change it with the language select.

Sekhmet:
We are using `lodash-webpack-plugin` on deployment and it was causing problems on deployment with `omit`. After enabling `flattening` `omit` works properly.

https://trello.com/c/0JgGTkTn/223-bug-delete-drafts - this is the same issue caused by `omit`. This will be fixed after merging this PR.

PR #749, opened 2017-09-20
Add See All and image uploading notifications translations

Sekhmet:
This adds missing `See All` and image uploading notifications text.

We have notifications in action creators as well but it is impossible to use `react-intl` from there.
It's possible to use it statically like [this](https://github.com/yahoo/react-intl/issues/41#issuecomment-227268578), but it doesn't fit our code base (we would have to pass `intl` into `redux-thunk` when creating store, but we won't be able to create `IntlProvider` because we store locale in store. This won't allow changing language as well). We might want to migrate to platform agnostic API like [counterpart](https://github.com/martinandert/counterpart)  - same as Condenser. However, I don't think that's priority right now.

https://trello.com/c/FUq0MJgH/241-translate-see-all-and-uploading-image-and

### DIFF
--- a/src/components/Navigation/Notifications/Notifications.js
+++ b/src/components/Navigation/Notifications/Notifications.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
 import NotificationFollowing from './NotificationFollowing';
 import NotificationReply from './NotificationReply';
 import NotificationTransfer from './NotificationTransfer';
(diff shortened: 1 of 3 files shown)


Sekhmet:
```js
    // src/comments/commentsActions.js
    if (!auth.isAuthenticated) {
      return dispatch(notify('You have to be logged in to comment', 'error'));
    }

    if (!body || !body.length) {
      return dispatch(notify("Message can't be empty", 'error'));
    }
```

I guess that's not something that users will see a lot as user shouldn't be able to send comment if they are unlogged or body is empty.

PR #752, opened 2017-09-21
Add SidenavUser

Sekhmet:
This adds alternative `Sidenav` version for user specific routes:
image

https://trello.com/c/4WAqLfXt/156-settings-pages


bonustrack:
- `Activity` should link to steemd.com same than topnav setting popover (with `target="_blank"`)
- `Edit profile` should link to https://steemit.com/@:username/settings (with `target="_blank"`)
- Drafts should not have right sidebar
- Settings and Drafts loading should be aligned to left
- When i switch from one page to another for 0.1 second there i can see the title in left like this: 
image
can we remove this effect?

bonustrack:
I forgot one point, can we have `StoryLoading` instead of classic loading on bookmarks page? Now there is both the circle loading and storyloading.

PR #762, opened 2017-10-07
Use raw YouTube embed in single post

Sekhmet:
This removes thumbnail and disables autoplay in single post.
https://trello.com/c/NksPm24B/286-single-post-youtube-preview-image-should-take-full-width


bonustrack:
Can you set full-width for the embed preview picture on feed?
image

Sekhmet:
Not sure how do we want to handle this. This image is smaller than feed container. We can change `max-width` to `width`, but this will cause image to scale thus destroying its quality.

PR #794, opened 2017-10-14
Add Edit Link to Post Footer

jm90m:
* Add Edit link (if own post) to bottom of post based on attached image in Trello Ticket
   - https://trello.com/c/OQ9wcT8J/314-edit-link-in-bottom-of-post

# Before
screen shot 2017-10-13 at 11 01 20 pm

# After

screen shot 2017-10-13 at 10 57 07 pm
screen shot 2017-10-13 at 10 57 12 pm

### DIFF
--- a/src/components/Story/Story.js
+++ b/src/components/Story/Story.js
@@ -75,6 +75,7 @@ class Story extends React.Component {
       ownPost,
       onLikeClick,
       onShareClick,
+      onEditClick,
     } = this.props;
 
     let followText = '';
@@ -206,6 +207,8 @@ class Story extends React.Component {
               pendingLike={pendingLike}
               onLikeClick={onLikeClick}
               onShareClick={onShareClick}
+              ownPost={ownPost}
+              onEditClick={onEditClick}
             />
           </div>
         </div>
--- a/src/components/Story/StoryFooter.js
+++ b/src/components/Story/StoryFooter.js
@@ -22,18 +22,23 @@ class StoryFooter extends React.Component {
     pendingLike: PropTypes.bool,
     onLikeClick: PropTypes.func,
     onShareClick: PropTypes.func,
+    ownPost: PropTypes.bool,
+    onEditClick: PropTypes.func,
   };
 
   static defaultProps = {
     pendingLike: false,
+    ownPost: false,
     onLikeClick: () => {},
     onShareClick: () => {},
+    onEditClick: () => {},
   };
 
   state = {
     shareModalVisible: false,
     shareModalLoading: false,
     reactionsModalVisible: false,
+    loadingEdit: false,
   };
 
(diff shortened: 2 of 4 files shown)


Sekhmet:
- It looks not aligned with the rest of text. I think we should have it aligned in the middle.
- It takes a while to save draft (what is done before editing post), can we add loading state to this link? Like in this popover menu:
image
- I'm not sure about this color. Isn't it to intrusive? Not sure if it was just for showcasing feature, or desired color @bonustrack

bonustrack:
Yes the color of "Edit" should be grey similar than "Reply" color:
image

jm90m:
@Sekhmet  just updated the commit
* edit should now be aligned, removed line-height, but kept `vertical-align: top` to align it.
screen shot 2017-10-14 at 12 26 20 pm
* Added Loading state, text should change into spinner when you click on text
* Changed color to be exact same gray hex used in reply link

Sekhmet:
- Loading icon is not aligned with other links. I think it would look better to just add loading icon instead of replacing text with it.
image
- Posts can be edited only in 7 days from posting, there is logic behind it: https://github.com/busyorg/busy/blob/new-design/src/components/Story/StoryFull.js#L160

@bonustrack should we remove edit post menu item from popover menu?

jm90m:
@Sekhmet shold we include an icon next to 'Edit' text similar to the one in poperover menu? That way loading can replace that icon instead of text?

Sekhmet:
I think that might look good. Now it kinda looks that `Edit` has share icon. Not sure how this will look like with two icons one after another though.

bonustrack:
We can keep both edit for now (in bottom and popoer menu).

bonustrack:
The reblog icon will not be visible in near future on the feed for our own post. It should be less confusing with Edit.

jm90m:
@Sekhmet Added edit icon & used same `__link` less styles as all other links in footer
screen shot 2017-10-14 at 1 36 35 pm

loading should be aligned too

screen shot 2017-10-14 at 1 40 16 pm

bonustrack:
Did you see the link "Edit" does not have the same color than the icon or other links:
image

jm90m:
@bonustrack what page? on hover the edit link has the purple color, but not on hover it has same color

Sekhmet:
What Fabien means is that this link has wrong color `#737fc4` but should have `#3e4a9a`.
image
vs
image

jm90m:
@Sekhmet did you mean `#4757b2`? I couldn't find `3e4a9a` but `#4757b2` is defined as @purple in `custom.less`

Sekhmet:
Yeah. Sorry, must have copied wrong one.

jm90m:
@Sekhmet alright updated it to use the @purple on hover from the less constants file

PR #798, opened 2017-10-14
Add Voting Power Slider

Sekhmet:
- [x] Voting Power settings.
- [x] Save settings with **Save** button.
- [x] Add Voting Power slider depending on settings.
- [x] Handle auto Voting Power.
- [x] Load previous vote value on that post (if any) and use it as initial value.
- [x] Add estimated vote value.
- [x] Default slider value.


bonustrack:
Did you see the link "Edit" does not have the same color than the icon or other links:
image

bonustrack:
I don't think we need the option `Enable and disable slider depending on my Voting Power`. Who will chose this? Someone either want slider or not want it.

Sekhmet:
It's still WIP.

Sekhmet:
Settings UI has been updated to add Voting Power + added Save button.
Any reviews, ideas, problems so far?
cc: @bonustrack @mynameisek @jm90m

bonustrack:
Wow! there is even tests, great work @Sekhmet! 🥇  Looking good for me.

bonustrack:
I have just 1 feedback, for the english text, should we talk about "upvote" or "like" ? "Voting power" or "Liking power"? ok maybe the last one is too much

Sekhmet:
I think we should use like/dislike across our app. I would use `Voting Power` though (capitalized this way).

Sekhmet:
Slider UI:
like2

Sekhmet:
There is one feature missing (saving previous vote percentage to SteemConnect metadata for feature votes). I don't think it is something very important (or even required) so I think we can review changes made in this PR and maybe merge them and work on that feature in other PR.

cc @bonustrack @mynameisek

bonustrack:
I think it's required @Sekhmet. Some users want to vote 50 posts with low power, without saving last voting power it's not friendly. If you think SC2 metadata is too much you can use localStorage.

bonustrack:
@Sekhmet can you remove `US` from the vote value and display it in a same way than another like `$0.15`.

Also 0% give me value of 0.01$, it should not be the case as 0% mean "unvote"
image

And can we have this style for the tooltip similar than vote list value:
image

bonustrack:
Another possibility @Sekhmet for the default voting power is to create a setting to define the fault voting power but not update on every vote. This setting could also be used with disabled voting slider (for vote with 1 click with specific voting power).

bonustrack:
I might found some issues, for some reason, it take a lot of time to get vote confirmed. I can see the loading for more than 10 seconds. Is this happening to you too @Sekhmet ? Also sometime my vote is not confirmed on the UI (icon stay grey).

src/components/StoryFooter/Payout.less
```diff
@@ -0,0 +1,18 @@
+.Payout {
+  position: absolute;
+  right: 0;
+  color: #99aab5;
```
jm90m: colors can probably go in `styles/custom.less`.

I usually name colors by their `hue-colorName`, I get the color name from here: (http://www.color-blindness.com/color-name-hue/)

So for `#99aab5` it would be, `@blue-botticelli: #99aab5`

bonustrack:
@Sekhmet 
- I'm not able to unvote with disabled slider.
- When slider disabled and there is a a default voting power value (unless it's 100%) there should the % in the tooltip
image
- With the slider, if i choose 0%, it will show me 0.01$ value of upvote which is not correct, should be 0.

src/app/appActions.js
```diff
@@ -1,53 +1,50 @@
...
-export const SET_LOCALE_START = '@app/SET_LOCALE_START';
-export const SET_LOCALE_SUCCESS = '@app/SET_LOCALE_SUCCESS';
-export const SET_LOCALE_ERROR = '@app/SET_LOCALE_ERROR';
 
 export const GET_TRENDING_TOPICS = '@app/GET_TRENDING_TOPICS';
 export const GET_TRENDING_TOPICS_START = '@app/GET_TRENDING_TOPICS_START';
 export const GET_TRENDING_TOPICS_SUCCESS = '@app/GET_TRENDING_TOPICS_SUCCESS';
 export const GET_TRENDING_TOPICS_ERROR = '@app/GET_TRENDING_TOPICS_ERROR';
 
+export const GET_REWARD_FUND = '@app/GET_REWARD_FUND';
```
jm90m: I see that we are having to recreate these action types a lot, it might be good to have a helper like what I did here: https://github.com/busyorg/busy/blob/27c72a602096ff802996f0aef30a912fc1e750b8/src/helpers/stateHelpers.js#L106

We can probably go back and refactor the other action types in a seperate PR, keep the lines of code in each file smaller too

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,100 @@
...
+  handleChange = (value) => {
+    this.setState({ value }, () => {
+      this.props.onChange(value);
+    });
+  };
+
+  formatTip = value => (
+    <div>
+      <FormattedNumber
+        style="percent" // eslint-disable-line react/style-prop-object
```
jm90m: kinda weird that they chose to use the `style` prop for this, but I guess thats what the library says to do (https://github.com/yahoo/react-intl/wiki/Components#formattednumber)

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,100 @@
...
+        value={value / 100}
+      />
+      <span style={{ opacity: '0.5' }}>
```
jm90m: is inline styles better here than using a className and applying styles from there? I have no preference, just wondering, when to use inline styles vs classNames.

src/components/Slider/Slider.less
```diff
@@ -0,0 +1,15 @@
...
+  margin: 0 auto;
+  padding: 0 16px;
+  max-width: 560px;
+
+  &__info {
+    margin-bottom: 12px;
+    text-align: center;
+
+    &__amount {
+      font-family: "Source Sans Pro", monospace;
```
jm90m: might be good to use one of the `@font-family` variables in custom.less

src/components/Story/Story.js
```diff
@@ -46,18 +57,20 @@ class Story extends React.Component {
   };
 
   handleClick = (key) => {
+    const { post } = this.props;
+
     switch (key) {
       case 'follow':
-        this.props.onFollowClick(this.props.post);
+        this.props.onFollowClick(post);
         return;
```
jm90m: Is there a reason to use `return` here vs `break` as in all other ones?

src/components/Story/Story.js
```diff
@@ -81,36 +98,57 @@ class Story extends React.Component {
     let followText = '';
 
     if (postState.userFollowed && !pendingFollow) {
```
jm90m: There seems to be a lot of logic here in the render function, it might be a good idea to create a helper function within this component called `this.getFollowText()` or something and have all the logic in there rather then in the render function.

src/components/Story/Story.js
```diff
@@ -81,36 +98,57 @@ class Story extends React.Component {
...
     }
 
     let popoverMenu = [];
 
     if (ownPost && post.cashout_time !== '1969-12-31T23:59:59') {
-      popoverMenu = [...popoverMenu, <PopoverMenuItem key="edit">
-        {saving ? <Icon type="loading" /> : <i className="iconfont icon-write" />}
-        <FormattedMessage id="edit_post" defaultMessage="Edit post" />
-      </PopoverMenuItem>];
+      popoverMenu = [
```
jm90m: Same thing here, looks like a lot of logic in this render function, it might be good to break it down into its own function maybe `this.renderPopoverMenu()`, also same with the rebloggedUI logic.

Sekhmet: Although I agree with this I don't think that it's related to this PR and working on multiple unrelated pieces in the same PR shouldn't be encouraged. I don't think we should rewrite half of our code base while adding one feature. I think we should work on this in new PR.

src/components/Story/Story.js
Sekhmet: Although I agree with this I don't think that it's related to this PR and working on multiple unrelated pieces in the same PR shouldn't be encouraged. I don't think we should rewrite half of our code base while adding one feature. I think we should work on this in new PR.

src/components/Slider/Slider.js
Sekhmet: We use this style quite a lot, so I think we should create global `translucentText` class for this property and use it everywhere it's needed.

src/components/Story/Story.js
Sekhmet: No good reason, it was this way before this PR. It's not big change so I will change this here.

src/components/Story/Story.js
jm90m: yeah good idea, this PR is already pretty big

PR #799, opened 2017-10-14
Set left and right margins to 0 for .Body pre; set width to 96%

### LINKED ISSUE #771: Bug display Code in comments
commentbug

eastmaels:
fixes #771

### DIFF
--- a/src/components/Comments/Comment.less
+++ b/src/components/Comments/Comment.less
@@ -88,6 +88,14 @@
       max-width: 100%;
     }
 
+   .Body {
+        pre {
+            padding-right: 0px;
+            padding-left: 0px;
+            width: 96%;
+        }
+    }
+
     &__collapsed {
       font-weight: bold;
       text-align: center;


Sekhmet:
Thanks for contributing.
Looks like this still overflows on smaller screens:
image
We should find a way to make this work well on different screen sizes.

You don't have to create new pull request. If you push new commits to this branch they will be added to this pull request as well.

eastmaels:
ok. thanks. will look into it.

Sekhmet:
I'm not even close to CSS expert so I'm not sure how useful this will be, but I noticed that if there is `white-space: normal` on `pre` or `code` element it renders properly (meaning it renders width as should but it removes scrollbars which is not what we want). Not sure if it's any useful, but it might help.

jm90m:
Hi, thanks for contributing!

So a deeper issue is that the width of the entire comment container itself is overlapping the right sidebar. Might be good to look towards fixing that first and the rest should be fixed.
screen shot 2017-10-14 at 1 11 06 pm

Sekhmet:
It is only the case if there is `pre>code` inside. If it's not it has proper width.

eastmaels:
Yes, based on my observation, this is caused by the `pre > code` blocks adding margins to the whole element so I tried removing the margins and shrinking the width.

image

Compared to that which does not contain a `pre > code`

image

jm90m:
Hello,

So I've been thinking about this one a lot, I think a better solution would be to wrap the text like so: https://github.com/busyorg/busy/pull/834

Rather than set a width. This fix should work well with responsive design.

bonustrack:
I'm closing this PR, the issue has been fixed here #834

eastmaels:
Nice workaround @jm90m . Thanks.

PR #801, opened 2017-10-14
Hide Reblog Icon for own post

jm90m:
* Trello: https://trello.com/c/H9CR4I1q/320-reblog-action-on-my-own-post
* Reblog link is hidden when it is the current user's own post
* When user is not logged in, it redirects to auth page when they click on reblog link

reblog-redirect-auth-busy

### DIFF
--- a/src/components/Story/Story.js
+++ b/src/components/Story/Story.js
@@ -22,7 +22,6 @@ class Story extends React.Component {
     pendingBookmark: PropTypes.bool,
     saving: PropTypes.bool,
     ownPost: PropTypes.bool,
-    isAuthenticated: PropTypes.bool,
     onFollowClick: PropTypes.func,
     onSaveClick: PropTypes.func,
     onReportClick: PropTypes.func,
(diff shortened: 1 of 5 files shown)


Sekhmet:
Looks good but I don't think that we should add redirect just yet.
I hope to implement this globally with one unified system and add modal with pitch to acquire new users or something like this. There is Trello card for this: https://trello.com/c/i2dwEj9r/91-show-login-signup-modal-for-unlogged-user-who-hit-action-button.

Sorry for confusion. I mentioned login modal to reason why it needed to show reblog icon for unauthenticated user.

jm90m:
@Sekhmet alright should I remove the redirect for now? or did you want me to look into adding the login signup modal?

bonustrack:
Yes @jm90m you can remove the redirection and make https://trello.com/c/i2dwEj9r/91-show-login-signup-modal-for-unlogged-user-who-hit-action-button in another PR

jm90m:
@bonustrack alright I removed passing `isAuthenticated` down and the redirect

PR #802, opened 2017-10-15
Show login/signup modal for unlogged user when they click on reblog link

jm90m:
* Trello: https://trello.com/c/i2dwEj9r/91-show-login-signup-modal-for-unlogged-user-who-hit-action-button
* Display login modal when clicking on reblog link, text on modal was based on comments in Trello.

screen shot 2017-10-14 at 10 39 56 pm


Sekhmet:
I'm not sure that we should just start working on that straight away.
There is a lot we should take care of to handle this properly (likes, dislikes, comments, saving posts, reporting posts, following users). This needs to be implemented in different places with different specifics (we should handle redirect links that will allow user to finish that actions right after they log in, so we need data about that action (post to be liked/disliked, comment, post to be saved, user to be followed).

I think it's not as trivial as it seems (it might be, but it won't result in high quality code and leave us with tons of duplicated and unmaintainable code).

See #803

jm90m:
@Sekhmet good point, I'll close the PR for now? and move the trello card back over to in progress?

Sekhmet:
You can use Blocked label and Trello board.

Issue #803, opened 2017-10-15
Implementation details on handling unauthenticated user actions


Sekhmet:
I feel like we could use redux middleware for this, that will pass data about required permissions per action in metadata (so whether user needs to be logged in or not) and if user doesn't have required permissions we could trigger global modal with payload that we could use to handle this action later after signing in.

We could add `meta.authAction = true` to action to notify our middleware that this action should be handled.

Middleware could behave like this.

For **authenticated** user:
1. Action is dispatched.
2. Middleware checks if action requires authentication and if user is authenticated.
3. Action continues normally.

For **unauthenticated** user:
1. Action is dispatched.
2. Middleware checks if action requires authentication and if user is authenticated.
3. Middleware opens modal with payload based on action dispatched.
4. Modal can cancel action or redirect user to SteemConnect (with redirect URL with parameters based on action) depending on user's choice.
5. User logs in on SteemConnect and get's redirected back to Busy.
6. We retrieve action from URL.
7. Continue action execution.

We could use this to handle different action permissions requirements (instead of `authenticated/unauthenticated` we could have actions that require `none`, `posting`, `active`, `master` or `memo` permissions). If user logged in with key that has insufficient permissions to perform such action we could request upgrading access.

What do you think?

bonustrack:
Related issues:
https://github.com/busyorg/busy/issues/315
https://github.com/busyorg/busy/issues/469

jm90m:
I like this idea of adding it in the middleware, but I went through the entire codebase and checked where a user would need to perform an action that requires authentication. There are only a few places that would require a user authentication, I think adding a middleware overcomplicates things in our stores, when we already passing in `steemAPI` and `steemConnectAPI` into the middlewares.

![screen shot 2017-12-21 at 2 30 04 pm](https://user-images.githubusercontent.com/5067716/34271711-839a93be-e65b-11e7-818f-06570f3e6d90.png)

I guess its all personal preference but I added a login modal here to all components that require it -https://github.com/busyorg/busy/pull/1227

And since we use React V16 and up, we can return multiple elements in a component without wrapping them (https://pawelgrzybek.com/return-multiple-elements-from-a-component-with-react-16/)

Like see how I did it here for the FollowButton in the PR - https://github.com/busyorg/busy/blob/05944c75af66de3deec46ec46054adbc53d2ae55/src/client/widgets/FollowButton.js#L92

Issue #469, opened 2017-05-01
Show alert when clicking on button action: "need to login" to Like/Upvote/Bookmark


mynameisek:
Do same like we did with the Comment button, but with all other actions which require login (Like/Dislike/Bookmark-hide)
![image](https://cloud.githubusercontent.com/assets/8390894/25573099/52e3917a-2e42-11e7-9aa7-b9f5ab5edc12.png)

sirrius-a-b:
This issue seems to be for v1 and it is already referenced for v2 here: #803 . Maybe this issue can be closed.

Issue #315, opened 2017-02-03
Direct the logged out users to the login page when they hit Comment


tibonova:
Users can start writing a comment without being logged in, but they cannot submit it, which may cause frustration.

p0o:
@tibonova Is this only happens in case of #339?

tibonova:
It happens in post view. 
I'm logged out. When I hit the comment button, I can start a comment. I cannot send it with Enter, because I'm not logged in, and there are no hint that I should do it.

sirrius-a-b:
This is no longer relevant because the comment textarea and button are not showing up anymore for logged out users. I think this can be closed.

Sekhmet:
Right, also related to #803.

PR #804, opened 2017-10-15
Update User Profile Responsive styles

jm90m:
* For medium/small screen sizes: keep username on separate lines, and centered.
* Trello: https://trello.com/c/PWPN1ZVS/255-responsive-user-hero-center-avatar-and-bring-name-in-another-line
* 
screen shot 2017-10-15 at 2 56 24 pm
responsive-user-page

### DIFF
--- a/src/styles/custom.less
+++ b/src/styles/custom.less
@@ -48,6 +48,3 @@
 @layout-zero-trigger-width   : 36px;
 @layout-zero-trigger-height  : 42px;
 
-// Screen Sizes
-@medium-screen: 990px;
-
--- a/src/styles/custom.less
+++ b/src/styles/custom.less
@@ -47,3 +47,7 @@
 @layout-trigger-height       : 48px;
 @layout-zero-trigger-width   : 36px;
 @layout-zero-trigger-height  : 42px;
+
+// Screen Sizes
+@medium-screen: 990px;
+
--- a/src/components/UserHeader.js
+++ b/src/components/UserHeader.js
@@ -78,13 +78,15 @@ const UserHeader = ({
             <i className="iconfont icon-more UserHeader__more" />
           </Popover>
         </div>
-        <div className="UserHeader__row UserHeader__handle">@{handle}</div>
-        <div className="UserHeader__rank">
-          <i className="iconfont icon-ranking" />
-          <FormattedMessage
-            id={getUserRankKey(vestingShares)}
-            defaultMessage={getUserRank(vestingShares)}
-          />
+        <div className="UserHeader__handle-rank-container">
+          <div className="UserHeader__row UserHeader__handle">@{handle}</div>
+          <div className="UserHeader__rank">
+            <i className="iconfont icon-ranking" />
+            <FormattedMessage
+              id={getUserRankKey(vestingShares)}
+              defaultMessage={getUserRank(vestingShares)}
+            />
+          </div>
         </div>
       </div>
     </div>
--- a/src/components/UserHeader.js
+++ b/src/components/UserHeader.js
@@ -13,7 +13,6 @@ import './UserHeader.less';
 
 const UserHeader = ({
   intl,
-  authenticated,
   username,
   handle,
   userReputation,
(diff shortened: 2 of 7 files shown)


src/styles/custom.less
```diff
@@ -47,3 +47,7 @@
 @layout-trigger-height       : 48px;
 @layout-zero-trigger-width   : 36px;
 @layout-zero-trigger-height  : 42px;
+
+// Screen Sizes
+@medium-screen: 990px;
```
bonustrack: We already have existing value for screensize see here: https://github.com/busyorg/busy/blob/702bc7cb698fb8f200a768024cfbc0285258ad13/src/styles/modules/variables.less#L7-L9
I think we the responsive layout should be triggered when size is lower than `768px`. Similar than right sidebar or search in topnav.

Sekhmet:
I feel like it would be more logical if `@handle` and user rank was rendered directly below username and above buttons. I'm afraid that it might require some refactoring of UserProfile to make this change though.
What do you think?

jm90m:
@bonustrack should we keep username and user rank on same line?
@Sekhmet yeah I agree, should I refactor to move @handle and user rank above the action buttons?

Sekhmet:
I think that it should look like this on mobile:
- Avatar
- Username
- Handle
- Rank
- Buttons

On desktop it should still look like it looks now.

bonustrack:
I agree too, maybe also these 2 could be on same line:
image
What you think?

Sekhmet:
Yeah, I think it would make sense to put those buttons on the same line.

jm90m:
I like that order @Sekhmet 

@bonustrack so those are 2 action items that do 2 different things, I think its good to keep them in separate lines, since they do different actions. One takes you to a different page while another brings a menu up

bonustrack:
@jm90m the button will mostly be "Follow" when not seeing your own profile and the menu [...] appear in down on small screen. IMO it look more sexy if the hero frame has not too much height.

jm90m:
@bonustrack ahh I see makes sense then to keep them on same line

bonustrack:
Menu [...] could also appear in top right like this:
image

src/styles/custom.less
jm90m: we can't use these properties, these are set to be `min-width:` rather than `max-width`

jm90m:
Just updated to move the menu to top right and move buttons down when using small screen
Also updated order based on what @Sekhmet said.
screen shot 2017-10-15 at 7 15 59 pm

bonustrack:
- Red and blue line should be same size. Now border in top and in down is not same.
image
- Also there is too much space between name and @handle (see green line)
- I still think we need the @handle and Grade in same line

jm90m:
Updated styling to be more aligned @bonustrack 
* also moved handle and rank on same line
screen shot 2017-10-16 at 4 16 30 pm
screen shot 2017-10-16 at 4 16 03 pm

screen shot 2017-10-16 at 4 17 28 pm

bonustrack:
Almost good. Can you change the breakpoint to `768px` `small`? 
And also can you update the style of the loader too? Now when the account is loading the frame is smaller than what expected in result:
loading

jm90m:
@bonustrack alright changed break point and updated loader

screen shot 2017-10-16 at 4 52 58 pm
screen shot 2017-10-16 at 4 53 04 pm

bonustrack:
Idealy it the loader look like this:

image

jm90m:
@bonustrack ahh i see, should be centered now

screen shot 2017-10-16 at 5 25 45 pm

bonustrack:
Can you make the space between element (in red) little bit bigger?
image
Also make the height of the frame bigger so load or not loaded it has the same size, now there is a difference see:
image
For width `768px` in CSS it would be better if we dont hardcode it but use this one https://github.com/busyorg/busy/blob/702bc7cb698fb8f200a768024cfbc0285258ad13/src/styles/modules/variables.less#L8

jm90m:
@bonustrack alright made updates based on your suggestions, but we can't use the current variable `@small` on there, since it is `min-width 768px`; for this case we need it to be `max-width: 768px;`

bonustrack:
The layout is broken when unlogged, everything appear in same line see:
image

Also it would be nice to keep the button "Follow" even unlogged. Later we would add a popup to login when click on "Follow" while unlogged

jm90m:
@bonustrack I updated it to show follow button for unlogged user, that seems to fix layout issue, but also follow button currently does nothing.
* I will create ticket on trello & github to address unlogged follow button issue

bonustrack:
@jm90m dont create ticket in Trello anymore, we will use only GitHub. But there is already issues for unlogged actions here:

https://github.com/busyorg/busy/issues/803
https://github.com/busyorg/busy/issues/469
https://github.com/busyorg/busy/issues/315

Sekhmet:
I think that we should show Follow button for unlogged user as well as we will ideally show login modal after clicking on it.

Sekhmet:
image
This is how it looks like at 900px width.

I think that we should add margin-left to avatar here.

jm90m:
Added padding for medium / large screen widths, also updated User Header Loader for those screen sizes.

screen shot 2017-10-17 at 1 05 08 pm

bonustrack:
@jm90m avatar should be aligned with the logo when window size is bigger than `998px`
image

jm90m:
@bonustrack removed padding for window bigger than 998px

screen shot 2017-10-17 at 2 41 08 pm

bonustrack:
Can we have same padding for window bigger than 998px for loading placeholder too? Then we will be perfect!

jm90m:
@bonustrack correct me if I'm wrong but with my latest changes window does have same padding for loading placeholders. My latest changes affected `UserHeader__container` which the loading container is a part of:

screen shot 2017-10-17 at 3 03 26 pm

bonustrack:
I mean in left here:
image

Also when loaded it's not aligned to logo
image

jm90m:
@bonustrack ahh I see what you mean, looks like the width previously set was not the right length of the alignment, should be updated now!

screen shot 2017-10-17 at 3 45 45 pm
screen shot 2017-10-17 at 3 52 18 pm

jm90m:
@bonustrack awesome thanks 🙏

PR #813, opened 2017-10-16
Add Post Recommendation component

jm90m:
WIP - Add Post Recommendation component

screen shot 2017-10-16 at 10 15 50 am


jm90m:
Updated based on feedback from trello card: https://trello.com/c/9ZM6ctlZ/118-single-post-sidebar-right-component-post-recommendation

mynameisek:
- [ ] When user clicks on a link from the component, he should land on the **Top of the page**. In the current situation it keeps the scroll position.

bonustrack:
- Post recommendation should also appear for unlogged users
- I can see that 2 request are being sent when opening a post see:
image
There should be only 1 request and the limit should be `6` (For showing 5 lastest post without current post)
- This border should not appear in last post recommended:
image
- `More from username` should not be centered
- There should be an icon near title, similar than "Interesting People", you can use the icon `headlines`

bonustrack:
Idealy we should use same class for all box title and avoid copy style.

jm90m:
@bonustrack made adjustments based on your feedback, also added the scroll to top on click of post navigation as @mynameisek requested.

So for the limit in the request, I had it set to 20 since posts that are 'resteemed/reblogged' get sent back in the request, is there any way just to get the post authored by user not the resteemed one?

jm90m:
Also added icon, and use same classname / removed duplicate styling on header title for both sidebar blocks, `Interesting People` and `Post Recommendation` titles
screen shot 2017-10-17 at 4 05 58 pm

bonustrack:
@jm90m we need only 6 result because we want also the resteemed ones. There is no way to get only post from author with the blockchain API, for some account like my account, you would need to paginate too much to get 5 post from me. You can change the title to "Recommended Posts".

bonustrack:
Or "Interesting Posts" ?

jm90m:
Oh I see what you mean @bonustrack, so instead of the `More from <username>`, it will be either "Recommended Posts" or "Interesting Posts", then we should also display the author maybe in between title and number of comments. I think "Recommended Posts" makes more sense here, since the PostRecommendation components get posts that are authored/resteemed by the current author of the post that you are viewing.

bonustrack:
Can you align Title and comment in red line?
image

bonustrack:
Can you add this style to Post title:
`font-size: 15px;` and `font-weight: bold;`
And `font-size: 15px;` for the "100 comments"

jm90m:
@bonustrack updated styles as you suggested, also added the author in each post recommended, with Link to post author

screen shot 2017-10-17 at 4 35 38 pm

jm90m:
alignment should be correct now too:
screen shot 2017-10-17 at 4 40 13 pm

bonustrack:
- Sorry can you set size of title back to `16px` (keep other text to `15px`).
- And show only 3 stories (request with limit 4)
- Also link comment to post url `/tag/@author/permlink#comments`

jm90m:
@bonustrack Post title should now be 16px while all others are at 15px, limited the stories to 4, and added link to comments

src/locales/en.json
```diff
@@ -82,6 +82,7 @@
   "amount_error_zero":"Amount has to be higher than 0.",
   "amount_error_funds":"Insufficient funds.",
   "balance_amount":"Your balance: {amount}.",
+  "by": "By",
```
bonustrack: This expression should include username like this `By {username}`

jm90m: @bonustrack  ahh I see, updated to use `FormattedMessage` component with updated expression

bonustrack:
We are still sending 2 time a request to the blockchain API for the same result see:
image
You can reproduce this when you login and refresh single post. The loading will be triggered 1 time when unlogged then, triggered a second time when logged

jm90m:
@bonustrack are you still seeing the 2 times request with latest commits? looks like a deeper issue was the `RightSidebar` component was being re-rendered over 5+ times

bonustrack:
Yes @jm90m i still seeing 2 requests. You need to login to see them.

bonustrack:
Also when switch from a post to another post from another author, the post recommended will not be loaded again (it will still show previous author recommendations).

jm90m:
@bonustrack the issue with multiple requests, should be fixed now, I don't see it on my end even when I'm logged in now. Also made call to change posts when clicking a new author post.

bonustrack:
I still can see 2 requests:
image

bonustrack:
When i'm logged and refresh a single post page, sometime i can see the post recommendation before behind logged but the block disapear and change to loading when login is success.

bonustrack:
Here we can see properly that the content is been loaded 2 time:
not

bonustrack:
Not sure if you waiting a review here but i still see 2 requests

Sekhmet:
I think that it would look better with even space above and below heading (now it has 8 pixels at top and 12 pixels at bottom.
image

I think that this padding is too big now:
image
I think that we should use `padding-bottom: 0px` for `.PostRecommendation__link:last-of-type`.

src/components/Sidebar/PostRecommendation.js
```diff
@@ -0,0 +1,152 @@
...
+    if (filteredRecommendedPosts.length === 0) {
+      return <div />;
+    }
+
+    return (
+      <div className="PostRecommendation">
+        <h4 className="PostRecommendation__title SidebarBlock__content-title">
+          <i className="iconfont icon-headlines PostRecommendation__icon" />
+          {' '}
+          {intl.formatMessage({ id: 'recommended_posts', defaultMessage: 'Recommended Posts' })}
```
Sekhmet: Can we use `FormattedMessage` whenever possible?

jm90m:
@Sekhmet alright made those changes that you suggested

Sekhmet:
I think that you missed first point from my review about uneven space around title. In my opinion it looks better with `8/8` pixels margins instead of `8/12`. Could we change that? What do you think @bonustrack? 

image

bonustrack:
You mean this @Sekhmet ? 
image

jm90m:
@Sekhmet @bonustrack ahh i see what you mean, ill make another PR for it

bonustrack:
@jm90m just found something. Now the component "Interesting People" appear also when unlogged. It's supposed to appear only to logged users, can you revert that too?

bonustrack:
@jm90m

jm90m:
@bonustrack so only show "Interesting People" component for authenticated users?

jm90m:
@bonustrack I hide Interesting People section for unathenticated users in my PR here: https://github.com/busyorg/busy/pull/857

PR #816, opened 2017-10-16
Enable STEEMJS_URL config var

bonustrack:
Remove hardcoded node url so we can switch to different node easy.


bonustrack:
I wonder if there is a standard for naming this file

jm90m:
When I worked with thoughbot, they named it `.sample.env`. 
* `https://github.com/thoughtbot/administrate/blob/master/.sample.env`

bonustrack:
I've used `.env.example` it seem the most popular

PR #834, opened 2017-10-18
Fix for Code in comments overflowing

jm90m:
* https://github.com/busyorg/busy/issues/771


jm90m:
Example here: `/philosophy/@michellecarter/what-is-the-highest-power` in the comments

Sekhmet:
I've tried this solution but it causes scrollbars not showing up which is not ideal for code blocks. Code is used to be displayed without braking lines with scrollbars if needed.

jm90m:
@Sekhmet @bonustrack Updated styles on this, should look like this now w/ scrollbar:

example here: `/test/@hellosteem/5ihr1p-test#@hellosteem/re-hellosteem-re-hellosteem-re-hellosteem-5ihr1p-test-20171018t142248652z`

screen shot 2017-10-18 at 5 24 23 pm
screen shot 2017-10-18 at 5 24 28 pm

bonustrack:
The element is bigger than expected, i'm not able to see the right part of the scrollbar
image

src/components/Story/Body.less
```diff
@@ -99,7 +99,8 @@
   }
 
   pre {
-    overflow-y: auto;
+    overflow: auto;
+    max-width: 90%;
```
bonustrack: There should be a better way than this :/

jm90m:
@bonustrack check latest commit, that should fix it, removed max-width

jm90m:
thank you 🙏  @bonustrack

PR #852, opened 2017-10-19
 Don't show transfer/block links on own profile

mktcode:
added checks for isSameUser to "Transfer" and "Block this user" links in UserInfo.js and UserHeader.js


mktcode:
Thankyou @bonustrack! Fixed it in https://github.com/busyorg/busy/pull/852/commits/5e89619471a59d47935452806e60f33eeaf367db

PR #870, opened 2017-10-20
Improve responsiveness

mktcode:
First I have to apologies for this rather messy pull request... containing different changes. I promise to use branches for future changes... :D

- On small screens the the topnav now has a search menu item triggering a popover with the search input. (I don't know if it's intended behavior that popovers appear at the bottom of the screen. If not... I have a fix for that ready. :>)
- The content has now a padding on small screens.

busy-changes-gif

And I fixed a small grammar issue in the german translation when the remaining time is "tomorrow". It should be correct grammar now in all cases (in x days, tomorrow, ...)

grammar-fix


mktcode:
I don't really understand this test procedure (test stories? never heard of that...:D) but I found this generated snapshot file... and updated it by deleting it and running the tests. Feels right... :D

bonustrack:
Each PR should have solve a single issue, the fix for German language should be in a different PR.

For the search feature on mobile we would like to have the input in top with an icon to go back like this:
image

mktcode:
@bonustrack To go back to where?

Sekhmet:
After clicking on search icon `Topnav` on mobile devices, `Topnav` should change into search layout as on screenshot above (back icon + search bar). After clicking back icon it should go back to the same layout as before.

PR #872, opened 2017-10-21
Update zh-CN again

ivysrono:
I found sth. better by https://github.com/utopian-io/utopian.io/pull/3


src/locales/zh.json
```diff
@@ -160,12 +160,12 @@
   "drafts_empty":"您没有保存任何草稿",
   "draft_untitled":"没有标题的草稿",
   "draft_delete":"删除这个草稿",
-  "payout_limit_reached":"这篇文章达到支出条件",
+  "payout_limit_reached":"这篇文章达到支付条件",
   "payout_promoted_amount":"推广： {amount}",
-  "payout_total_past_payout_amount":"收益总额：{amount}",
+  "payout_total_past_payout_amount":"收益总额： {amount}",
```
bonustrack: It feel like there is 2 spaces, what the point of this change?

Sekhmet: It looks like it because it's different character than `:` - `：`. There was space missing after that character previously.

ivysrono: Chinese symbols look wider.
This change is to keep in line with:
>  "payout_promoted_amount":"推广： {amount}",
  "payout_total_past_payout_amount":"收益总额： {amount}",
  "payout_author_payout_amount":"作者收益： {amount}",
  "payout_curators_payout_amount":"点赞收益： {amount}",
  "payout_potential_payout_amount":"预计收益： {amount}",

PR #876, opened 2017-10-21
added mobile padding

mktcode:
added mobile padding from #870 (https://github.com/busyorg/busy/pull/870/commits/b3278f6b72d236a2824af96c347e3426f1c925d5)


bonustrack:
@mktcode What issue you are trying to solve with this PR? This padding is making the content in mobile much smaller, i don't think it's good for user experience, we just lost some space.
Before: 
image
After: 
image

mktcode:
Well... it'looked rather unintended because of the border-radius.

screenshot-2017-10-22 mkt - busy

And on the settings page the text had no space at all.

screenshot-2017-10-22 busy
I don't know if there are more places where this happens.

And last but not least I just think it looks better but that might be a matter of taste. At least the border-radius should be removed on mobile or maybe the padding I added could just be smaller.

mktcode:
I reduced the padding now to match the space between posts, so it's smaller now. I really think it looks better than without the padding. But you decide... :)

bonustrack:
I think there is not need of padding or margin, we should just remove the left and right borders and border radius of the story in feed when screen is smaller and story appear in full width.
image

mktcode:
Ok I'll do that...

PR #879, opened 2017-10-21
Use babel-preset-env instead es2015

Ne3l:
Fixes npm install warning suggesting use babel-preset-env instead of deprecated es2015,es2016,es2017.


Sekhmet:
Hey,
we are using different code style. Please use Prettier with eslint integration, or update it manually, but we suggest using Prettier so it won't cause messed up commit diffs later on.

Ne3l:
I use Prettier but if i didn't disable it on the workspace, on saving webpack/webpack.server.config.js it would format the whole file.

Sekhmet:
Thanks a lot!

Issue #807, opened 2017-10-16
Comment form should not be cleared on upvote


espoem:
![comment-vote](https://user-images.githubusercontent.com/16621517/31608356-66b27c18-b270-11e7-97e6-81c372c08949.gif)

#### Expected behavior

The content in the comment form should stay untouched after upvoting the post.

#### Actual behavior

Comment form is cleared after upvoting the post.

#### How to reproduce

1. Start writing a comment for a post. Don't submit it yet.
2. Upvote the post.
3. The comment form is cleared.

* Browser: Chrome 61.0.3163.100
* Operating system: Windows 10

jkarthionline:
I started looking at this issue. Thanks

PR #881, opened 2017-10-22
Comment form should not be cleared on upvote #807

SimpliciTea:
in components/Comments/CommentForm.js:

-- move this.setState({ inputValue: '' }) from
   componentWillReceiveProps to onSubmit method

I believe that this creates the intended behavior, although I'm not positive
that I see everything that the prior componentWillRecieveProps was doing


bonustrack:
Thank you for your PR @SimpliciTea. I spotted an issue when i click "Comment" to send a comment, the textarea will be cleared. It should not be the case:
comment

SimpliciTea:
@bonustrack So it should not clear on upvote or comment submission. It should clear on comment submission success?

Sekhmet:
I can't figure out what @bonustrack means exactly from this GIF, but we should clear comment form after submission success, but only the comment form which was submitted. If there were other comment forms opened and some text was typed into them they should remain in the same state even after submitting one of them, so only one of them is cleared after submitting (the one that succeeded).

bonustrack:
What i mean in the GIF is that my comment is hidden in the input when i hit the button "Comment". Instead of my comment, I see the placeholder "Write a comment". On current busy version it's not the case. It's confusing and if comment fail to be sent i think it's lost.

bonustrack:
@SimpliciTea Yes it should be cleared only on comment success only.

ryanbaer:
I don't think the package-lock.json file should be committed unless the scope of your PR actually required it - i.e. adding / updating / removing a dependency or something. It doesn't seem to me that the file should've been committed here, right?

bonustrack:
Agree with @ryanbaer about the file `package-lock.json`, as there is no change in `package.json` i think it should not be commited on this PR @SimpliciTea.

bonustrack:
I'm closing with PR due to inactivity

PR #891, opened 2017-10-23
Create Wallet page

jm90m:
* WIP - create wallet page (https://github.com/busyorg/busy/projects/6#card-5213698)
screen shot 2017-10-23 at 3 09 31 pm


bonustrack:
- Can you put white background on these frames?
image
- Looks like the icons Steem Power and Steem Dollar are reversed
- Can we have the values in Saving on the same line like this:
image

bonustrack:
1: Can you change the icon and background with the same color than Paresh design?
2: Can you put any amount aligned on the right? (include claim reward amounts)
3: For claim reward, can you hide the 0 amounts? And change VESTS to SP
image

bonustrack:
- Numbers should be a formatted number like `1,300.000 STEEM` similar than others number in the interface
image
- Also can you use 22px size for icons? Include balance icons, and change the icon for claim operation to `success_fill` icon?

jm90m:
@Sekhmet can you review when you have time? Thanks!

src/wallet/SavingsTransaction.js
```diff
@@ -0,0 +1,84 @@
...
+          id="transfer_to_savings"
+          defaultMessage="Transfer to savings {amount} to {username}"
+          values={{
+            amount: numberWithCommas(transactionDetails.amount),
+            username: <Link to={`/@${transactionDetails.to}`}>{transactionDetails.to}</Link>,
+          }}
+        />
+      );
+    case 'transfer_from_savings':
+    default:
```
Sekhmet: Is this intentional? If it is, are we sure that any other text will mean that this transfer was `transfer_from_savings`?

src/wallet/UserWalletTransactions.js
```diff
@@ -0,0 +1,88 @@
...
+import TransferTransaction from './TransferTransaction';
+import SavingsTransaction from './SavingsTransaction';
+import PowerUpTransaction from './PowerUpTransaction';
+import ClaimReward from './ClaimReward';
+
+const UserWalletTransactions = ({ transactions, currentUsername }) => (
+  <div className="UserWalletTransactions">
+    {transactions
+      .map((transaction, index) => {
+        const key = `${transaction.trx_id}${index}`;
```
Sekhmet: Is it necessary to append map's `index` to key? Can there be duplicate transactions?

src/wallet/walletActions.js
```diff
@@ -1,7 +1,53 @@
...
+  type: GET_GLOBAL_PROPERTIES.SUCCESS,
+  payload,
+});
+
+export const getGlobalProperties = () => dispatch =>
+  getDynamicGlobalProperties().then((result) => {
+    dispatch(getGlobalPropertiesSuccess(result));
+  });
+
+export const getUserTransactionsSuccess = (username, transactions) => ({
```
Sekhmet: Can we use `createAction` from `redux-actions` wherever possible?

src/wallet/walletActions.js
```diff
@@ -1,7 +1,53 @@
...
+  type: GET_USER_TRANSACTIONS.SUCCESS,
+  payload: {
+    username,
+    transactions,
+  },
+});
+
+export const getUserTransactions = username => (dispatch) => {
```
Sekhmet: Wouldn't it be better to use `redux-promise-middleware` for this (and in other actions that are working with promises, i.e. `getGlobalProperties`, `getUserEstAccountValue`)?

src/wallet/walletReducer.js
```diff
@@ -1,8 +1,19 @@
-import { OPEN_TRANSFER, CLOSE_TRANSFER } from './walletActions';
+import {
+  OPEN_TRANSFER,
+  CLOSE_TRANSFER,
+  GET_GLOBAL_PROPERTIES,
+  GET_USER_TRANSACTIONS,
+  GET_USER_EST_ACCOUNT_VALUE,
+} from './walletActions';
```
Sekhmet: We mostly use `import * as xTypes from './xActions';` and use `xTypes.ACTION_TYPE` inside reducer. I think we should stick to this naming convention.

src/wallet/walletReducer.js
```diff
@@ -18,10 +29,49 @@ export default function walletReducer(state = initialState, action) {
         ...state,
         transferVisible: false,
       };
+    case GET_GLOBAL_PROPERTIES.SUCCESS: {
+      const { total_vesting_shares, total_vesting_fund_steem } = action.payload;
```
Sekhmet: I don't think that we should use destructuring inside reducers when variables are only used once. This adds unnecessary block inside case.

src/helpers/stateHelpers.js
```diff
@@ -102,3 +102,9 @@ export const sortCommentsFromSteem = (list, commentsState, sortBy = 'trending')
     )
     .reverse();
 };
+
+export const createAsyncActionType = type => ({
+  PENDING: `${type}_PENDING`,
```
Sekhmet: We use `START`, `SUCCESS`, `ERROR` for different states of action progress: https://github.com/busyorg/busy/blob/7b122f361e3923cd3e18016659b25b990bb8c68e/src/store.js#L26

src/wallet/UserWalletTransactions.js
jm90m: @Sekhmet, yeah I have had cases where there were duplicate transaction ids :(

src/styles/custom.less
```diff
@@ -10,11 +10,17 @@
...
+@blue-botticelli: #99aab5;
+@blue-heather: #A1AFBA;
+@blue-solitude: #F5F6F9;
 @teal:   #5bc0de;
 @pink:   #ff5b77;
 @purple: #4757b2;
 @normal-grey: #4f545c;
 @grey-mortar: #595959;
+@white-gainsboro: #e9e7e7;
+@grey-night-rider: #303030;
```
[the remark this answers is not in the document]
jm90m: I have no preference but I like your format! I'll switch to that `basecolor-camelCasedHue` so `grey-nightRider`

src/wallet/walletActions.js
jm90m: @Sekhmet yup, refactoring it to use `redux-promise-middleware`

src/wallet/walletActions.js
jm90m: @Sekhmet was kinda confused at using `redux-promise-middleware` at first, but looking into it,  I was able to refactor these actions using it 👍

src/helpers/stateHelpers.js
jm90m: @Sekhmet ahh I see now... since it goes along with the `redux-promise-middleware`, updated it to include the action

src/wallet/SavingsTransaction.js
jm90m: yeah it was intentional, I figured since I filter by `savings` transactions anyways then one of them can be defaulted.. but I updated it now to default to return null

src/wallet/ClaimReward.js
```diff
@@ -0,0 +1,68 @@
...
+      >{`${numberWithCommas(parsedRewardSteem.toFixed(3))} STEEM`}</span>,
+    );
+  }
+
+  if (parsedRewardSbd > 0) {
+    payouts.push(
+      <span
+        key="SBD"
+        className="UserWalletTransactions__payout-rewards"
+      >{`${numberWithCommas(parsedRewardSbd.toFixed(3))} SBD`}</span>,
```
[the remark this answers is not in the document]
jm90m: I think with FormattedNumber the default is 3 digits, so don't need the extra props

src/wallet/ClaimReward.js
```diff
@@ -0,0 +1,68 @@
...
+const getFormattedPayout = (rewardSteem, rewardSbd, rewardVests) => {
+  const payouts = [];
+  const parsedRewardSteem = parseFloat(rewardSteem);
+  const parsedRewardSbd = parseFloat(rewardSbd);
+  const parsedRewardVests = parseFloat(rewardVests);
+
+  if (parsedRewardSteem > 0) {
+    payouts.push(
+      <span
+        key={'STEEM'}
```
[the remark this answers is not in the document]
jm90m: woops missed this, good catch

jm90m:
@Sekhmet Made updates based on your suggestions, whenever you got time could I get another review? Thanks!

src/wallet/walletActions.js
```diff
@@ -1,7 +1,50 @@
...
 
 export const openTransfer = createAction(OPEN_TRANSFER);
 export const closeTransfer = createAction(CLOSE_TRANSFER);
+
+export const getGlobalProperties = () => (dispatch) => {
+  dispatch({
+    type: GET_GLOBAL_PROPERTIES.ACTION,
+    payload: {
+      promise: getDynamicGlobalProperties().then(result => ({
+        ...result,
```
Sekhmet: What does it do? Won't it result in exactly the same object?

src/helpers/regexHelpers.js
```diff
@@ -1,3 +1,6 @@
 export const imageRegex = /https?:\/\/(?:[-a-zA-Z0-9._]*[-a-zA-Z0-9])(?::\d{2,5})?(?:[/?#](?:[^\s"'<>\][()]*[^\s"'<>\][().,])?(?:(?:\.(?:tiff?|jpe?g|gif|png|svg|ico)|ipfs\/[a-z\d]{40,})))/gi;
 
+export const numberWithCommas = number =>
```
Sekhmet: Can we replace this with `FormattedNumber` altogether or we require to use this format no matter what locale does user use?

src/wallet/walletActions.js
jm90m: yeah you're right lol, I'm just to use copying objects, will change though

src/helpers/regexHelpers.js
jm90m: @Sekhmet So I initially added this because I thought it would be better than having to parse out the different currencies for each amount, in the different transactions. Since some transactions are in STEEM and some are in SBD. But I just pushed up changes based on your suggestions, lmk if you think this way is better with FormattedNumber

src/wallet/walletActions.js
```diff
@@ -18,9 +18,7 @@ export const getGlobalProperties = () => (dispatch) => {
   dispatch({
     type: GET_GLOBAL_PROPERTIES.ACTION,
     payload: {
-      promise: getDynamicGlobalProperties().then(result => ({
-        ...result,
-      })),
+      promise: getDynamicGlobalProperties().then(result => result),
```
Sekhmet: Isn't `getDynamicGlobalProperties().then(result => result)` same as `getDynamicGlobalProperties()`?

jm90m: yeah you're right, haha i think i need some coffee this morning, thanks for catching these things 👍

jm90m: just updated it now.

src/components/Navigation/Sidenav.js
```diff
@@ -4,14 +4,15 @@ import { FormattedMessage } from 'react-intl';
 import { NavLink } from 'react-router-dom';
 import './Sidenav.less';
 
-const isNews = (match, location) => location.pathname !== '/';
+const isNews = (match, location) => location.pathname.match(/trending/);
+const isUserWallet = (match, location) => location.pathname.match(/.@.*(wallet)/);
```
Sekhmet: This will match `busy.org/@sekhmetwallet` which shouldn't be the case. `.@.+\/(wallet)` seems to work, but I'm not regex magician.

src/components/Sidebar/SteemTrendingCharts.js
```diff
@@ -0,0 +1,97 @@
+import React, { Component } from 'react';
+import Trend from 'react-trend';
+import fetch from 'isomorphic-fetch';
+import _ from 'lodash';
+import Promise from 'bluebird';
+import { FormattedMessage } from 'react-intl';
+import './SteemTrendingCharts.less';
```
Sekhmet: Style is usually last import. Can we change that to make it concise with the rest of our app?

src/wallet/UserWalletTransactions.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedNumber } from 'react-intl';
+import './UserWalletSummary.less';
```
Sekhmet: Can we move styles down as well?

src/wallet/UserWalletSummary.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, FormattedNumber } from 'react-intl';
+import steem from 'steem';
+import Loading from '../components/Icon/Loading';
+import './UserWalletSummary.less';
```
Sekhmet: Should be moved down as well.

src/components/Sidebar/SteemTrendingCharts.js
jm90m: sure

src/components/Navigation/Sidenav.js
jm90m: yeah that works.

src/wallet/UserWalletTransactions.js
jm90m: yup done

bonustrack:
@jm90m can you change the url for the wallet to `/wallet` instead of `/@fabien/wallet`?

bonustrack:
Everything else looking good for me to merge, well done!

jm90m:
@bonustrack just updated it to use`/wallet`

Issue #810, opened 2017-10-16
Already followed user appears among Interesting People after (re)loading the page


espoem:
![int-people-known](https://user-images.githubusercontent.com/16621517/31609301-d6c523f4-b273-11e7-9d8c-64526a88a561.gif)

#### Expected behavior

Only users that I am not following should appear in *Interesting People* bar.

#### Actual behavior

A user that I am already following may appear in the bar.

#### How to reproduce

1. Go to `/trending` page. You need to load/reaload the page. Clicking the _News_ won't work.
2. Repeat step 1 until you see a user you are following.

* Browser: Chrome 61.0.3163.100
* Operating system: Windows 10

ryanbaer:
Working on a fix for this

Sekhmet:
Solved in #893.

PR #893, opened 2017-10-24
Fix for #810: followed user appears among recs after reloading the page

ryanbaer:
This is a fix for the issue that caused previously followed users to appear in new recommendations for Interesting People.

The issue had to do with React lifecycle. `getRandomPeople` was being called from within the constructor of the component, quite a while before `@user/GET_FOLLOWING_SUCCESS` returns with the list of the user's current followers. As a result, on page load, `getRandomPeople` always received an empty array of followers and hence did not filter any users out.

My approach to the solution was to pull the local state out of the component and into the Redux state tree. This way, `RightSidebar` no longer has to deal with the follower list directly, and  the logic for generating the recommendations is handled in the reducer. Any components making use of the interesting people list don't have to be concerned with how it's generated.

This approach also allows us to update that list in response to other events, should we want to. For example, I had a small implementation going where, if the user followed someone (`@user/FOLLOW_USER_SUCCESS`) that was in the list, it would:
- Remove that person from the Interesting People list
- Keep the remaining 4 people in the list
- Populate the empty slot with another random person

It was a quick implementation that I decided to leave out since it diverges from the original behavior of being able to click unfollow after you've clicked follow, but I think it's a good example of what else could be done.

Thanks.


ryanbaer:
Hey @Sekhmet I saw you moved this to Done in the workflow - was that you signing off or do you have review comments?

bonustrack:
@ryanbaer it's me i've merged to "Done" because i was thinking you were waiting for review. I've just gave you "Write" authority on this repo so you can add tag "WIP" when you are still working on a PR and remove it when you need a review

ryanbaer:
Thanks for the add & explanation @bonustrack, I am waiting for review so it's good where it is.

src/app/Sidebar/RightSidebar.js
```diff
@@ -18,41 +18,24 @@ import SignUp from '../../components/Sidebar/SignUp';
...
+  }), {
+    updateRecommendations,
+  },
 )
 export default class RightSidebar extends React.Component {
   static propTypes = {
     authenticated: PropTypes.bool.isRequired,
     authenticatedUser: PropTypes.shape().isRequired,
-    followingList: PropTypes.arrayOf(PropTypes.string).isRequired,
+    recommendations: PropTypes.arrayOf(PropTypes.object).isRequired,
```
Sekhmet: Can we use `PropTypes.shape()` instead of object with correct properties (so it contains `name` and other properties if any)? I think we should start being more strict about type checking.
You can add the same prop type here: https://github.com/ryanbaer/busy/blob/9d6899bf4256f224f40a4d610b4914794d14fbaf/src/components/Sidebar/InterestingPeople.js#L32

src/user/userReducer.js
```diff
@@ -12,13 +12,22 @@ const initialState = {
...
     isFetching: false,
   },
 };
 
+// filterRecommendations generates a random list of `count` recommendations
+// include users followed by the current user.
+const filterRecommendations = (following, count = 5) => {
+  const usernames = Object.values(following);
+  const filtered = usernames.length > 0 ? people.filter(p => !usernames.includes(p)) : people;
+  return filtered.sort(() => 0.5 - Math.random()).slice(0, count).map(r => ({ name: r }));
```
Sekhmet: I don't think that checking `usernames.length` is necessary. I think we can chain this to one expression like this:
`const filterRecommendations = (following, count = 5) => Object.values(following).filter(...).sort(...).slice(...).map(...)`
Variable `r` in map has not clear meaning so I think we can change it to `name` (so we can use object property literal shorthand like this: `map(name => ({ name }))`

src/app/Sidebar/RightSidebar.js
ryanbaer: Sure, done.

src/user/userReducer.js
ryanbaer: Good suggestions, especially the object property literal shorthand.

The reason for the `usernames.length` check is to not iterate over the `people` list (which is fairly large) an extra time if it's not necessary (if the followers list hasn't loaded yet or the user has no followers). Since it's only a list of 101 people at this time, I suppose it's not a huge deal to iterate it more than necessary.

Your suggestion for rewriting the function in the arrow syntax shorthand of `param => returnval` does look nice, but what you wrote is actually the other way around from how it is.

It's not `Object.values(following)` that we're filtering on - it's `people`. We want to say, "Take the list of interesting people and filter out anyone who the user already follows".

So if I rewrite the function as suggested, we get:
```
const filterRecommendations = (following, count = 5) =>
  people.filter(p => !Object.values(following).includes(p))
        .sort(() => 0.5 - Math.random())
        .slice(0, count)
        .map(name => ({name}));
```

This I have a problem with, because in order to gain the syntactic sugar of `param => returnval`, we have to recompute Object.values(following) for every item of the `people` list (101 times), and I just don't think that's a good idea. In the case that a user follows 1,000 or more users, that's could get slow, and I definitely don't want reducer code to be slow.

So here's what I'm proposing:
- No more usernames.length check
- Sacrifice the syntactic sugar function definition in favor of only iterating the user's following list *once* instead of `people.length` times.
- Definitely clean up the `map` to use the literal shorthand

It will look like this:
```
const filterRecommendations = (following, count = 5) => {
  const usernames = Object.values(following);
  return people.filter(p => !usernames.includes(p))
        .sort(() => 0.5 - Math.random())
        .slice(0, count)
        .map(name => ({name}));
}
```

I've pushed the updated code. Let me know what you think.

src/user/userActions.js
```diff
@@ -133,3 +133,6 @@ export const getFollowing = (userName = '') => (dispatch, getState) => {
     },
   });
 };
+
+export const UPDATE_RECOMMENDATIONS = '@user/UPDATE_RECOMMENDATIONS';
+export const updateRecommendations = () => dispatch => dispatch({ type: UPDATE_RECOMMENDATIONS });
```
Sekhmet: Can you use `createAction` from `redux-actions` for `updateRecommendations` instead of creating it by hand?

src/app/Sidebar/RightSidebar.js
```diff
@@ -18,41 +18,24 @@ import SignUp from '../../components/Sidebar/SignUp';
...
+    updateRecommendations,
+  },
 )
 export default class RightSidebar extends React.Component {
   static propTypes = {
     authenticated: PropTypes.bool.isRequired,
     authenticatedUser: PropTypes.shape().isRequired,
-    followingList: PropTypes.arrayOf(PropTypes.string).isRequired,
+    recommendations: PropTypes.arrayOf(PropTypes.shape({ name: PropTypes.string })).isRequired,
+    updateRecommendations: PropTypes.func.isRequired,
```
Sekhmet: We tend not to use required functions as props. Can you make it not required and set NOOP (`() => {}`) as `defaultProp` for it?

ryanbaer:
Cool, updated. Thanks for all of your feedback, it's been really helpful to get an idea of the general style / approach used in the codebase.

Sekhmet:
Thanks for update.
I'm pretty sure you don't have to create another action creator to wrap previous action creator with `dispatch`.
If you do:
```js
export const updateRecommendations = createAction(UPDATE_RECOMMENDATIONS);
```

And use that `updateRecommendations` in `connect` inside `mapDispatchToProps` as an **object of functions** (like `{ updateRecommendations }`) react-redux should wrap that function call in dispatch itself.

Could you verify it and update it this way if it works?

ryanbaer:
If you're referring to something like this:
```
@connect(
  state => ({
    authenticated: getIsAuthenticated(state),
    authenticatedUser: getAuthenticatedUser(state),
    isAuthFetching: getIsAuthFetching(state),
    recommendations: getRecommendations(state),
  }), {
    updateRecommendations,
  },
)
```

That's what my approach originally was, and it yielded some unpleasant results, mainly a *very* slow refresh button (on the Interesting People pane) with a bunch of logspam from React about SyntheticEvent.

image


Creating the action creator as a function which receives dispatch and invokes it resolved the problem.

Searching through the codebase a bit, all of the actions I found which import themselves directly into mapDispatchToProps as an object of functions have action creators that follow the style I used.

More importantly, [this thread](https://github.com/reduxactions/redux-actions/issues/121) on redux-actions documents the problem well. The best solution, proposed in the thread, that keeps with your approach, is to change it to the following:

```
export const UPDATE_RECOMMENDATIONS = '@user/UPDATE_RECOMMENDATIONS';
export const updateRecommendations = createAction(UPDATE_RECOMMENDATIONS, () => {});
```

Definitely good to know on my part.

Anyway, I pushed the changes with the above style since it's more in line with your suggestion.

Sekhmet:
Hey, thanks for taking look at this problem.
I think we shouldn't change `payloadCreator`. Our problem is that we dispatch an action with payload, not that we transform that payload wrong way. We have to make sure we dispatch actions properly with proper arguments.
We should have `handleRefreshClick` method inside `RightSidebar` that dispatches action properly, with correct arguments (none).
That should fix issue with synthetic event and won't require us to hack around `redux-actions`.

PR #894, opened 2017-10-24
fixed .StoryLoading border/border-radius on small screens

### LINKED ISSUE #888: Feed story loading: remove borders left, right and radius on small screen
When loading a feed there will be a container loader for the story. This container on small screen have border right and left and border-radius:
image
We should remove these border and the border radius in small screen similar than what @mktcode did in #876

mktcode:
fixes #888

### DIFF
--- a/src/components/Story/StoryLoading.less
+++ b/src/components/Story/StoryLoading.less
@@ -2,11 +2,16 @@
 
 .StoryLoading {
   position: relative;
-  border-radius: 4px;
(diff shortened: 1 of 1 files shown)


bonustrack:
@mktcode The build fail on Heroku with this error: `Variable @small is undefined`

mktcode:
Oops! Fixed it... I just wonder why my local install didn't complain. o.0

PR #898, opened 2017-10-25
Update PostRecommendation title padding to 8px

jm90m:
@Sekhmet Is this what you mean? Update title to have same padding as top.


jm90m:
@Sekhmet awesome, thanks, yeah I removed the `LatestRecommendations` component as you suggested 👍

PR #900, opened 2017-10-25
Add cancel button on post creation.

### LINKED ISSUE #871: Busy.org new version should add a Cancel button in the Write a Post page
#### Expected behavior
When I click on Write Post, and then in the midst of writing I change my mind to Cancel, there is no way I can do it now as there is no Cancel button. A Cancel option should be there.

The Cancel button should allow to cancel and go back to profile page, if the user changes his mind to write and instead Cancel.
#### Actual behavior
Currently there is no provision of Cancel.

#### How to reproduce
Log in to https://nd.busy.org/
Click on Write Post , and give some title. There is no way to cancel from here.

* Browser: Firefox
* Operating system: Windows 8.1

write post

Ne3l:
Fixes #871.

image

There are some points that i have doubts:

1. Should we create a new translate item with "Discard Draft"?
2. After discard it should redirect to Profile page, Index or last page before entering write post?
3. Is fine to redirect in the action or should be handled outsided by the caller?


Sekhmet:
I wonder whether it is good idea to delete entire draft after user click cancel on editor.
I would expect to just go back, or go back and revert draft to state from before editing.

Ne3l:
Updated the behaviour of cancel.

If the draft existed before entering, it would discard the changes made.
If the draft was created, it would discard the draft and go to the start.

src/post/Write/Write.js
```diff
@@ -102,17 +113,42 @@ class Write extends React.Component {
...
+
+      this.props
+        .saveDraft({ postData: this.getNewPostData(initalForm), id: this.props.draftId })
+        .then(() => {
+          this.props.goBack();
+        });
+    } else {
+      // If didn't exist just delete it.
+      this.props.deleteDraft(this.props.draftId).then(() => {
+        this.props.replace('/');
```
Ne3l: Not sure how to get the previous state and replacing the current route with last one.

Also, in case the draft was saved (got id) it should not only replace the current one but the last two("/write","/write?draft=id").

Sekhmet:
@bonustrack do you have opinion on this behaviour?

bonustrack:
When i click "Cancel" nothing happen for few second then i'm redirected to drafts where i can still see my canceled post as draft without the latest changes. If i open the draft again and click "Cancel" again, i'm redirected to the drafts but the post remain same. I feel it's too confusing.

I would rather see a button "Delete" with a modal to confirm that you want to delete the draft (similar than reblog modal), then delete draft for real and redirect user to /write with an empty page. See how medium prompt user for delete:
image


Also the button in red is too much eye catching i would prefer to see a button with this colors:
image

Ne3l:
So there will be 1 button for delete the draft and 1 for discard changes?

Sekhmet:
I don't think that we need discarding changes option just yet. If we would want to have something like this we should migrate to versioning system for drafts based on patches and allow user to go back to specific version. This is rather big feature and we are in middle of reworking our workflow so I think we should postpone this feature (but we can work on deleting draft from editor) for a while.
You can create issue to discuss details about this versioning system and assign it to yourself if you wish to work on it. When we have everything figured out (both about our workflow and this particular feature) we can start working on this.
For now I would use this PR for deleting drafts from editor.

Ne3l:
When you are creating a draft there isn't anything to delete, what is the expected behaviour?

Ne3l:
Updated the behaviour. Also implemented the are you sure modal when deleting in drafts page.

The Spinner inside modal i don't like much, any idea?

src/components/Editor/Editor.less
```diff
@@ -131,6 +131,18 @@
         text-align: right !important;
       }
     }
+
+    &__cancel {
+      margin-bottom: 0 !important;
+      padding-right: 15px;
```
bonustrack: It feel there is too much space between the 2 buttons. Can you use `10px` instead of `15px`?

src/post/Write/DeleteDraftModal.js
```diff
@@ -0,0 +1,75 @@
...
+        okText={intl.formatMessage({ id: 'confirm', defaultMessage: 'Confirm' })}
+        cancelText={intl.formatMessage({ id: 'cancel', defaultMessage: 'Cancel' })}
+        onOk={() => {
+          this.deleteDraft();
+        }}
+        onCancel={() => {
+          this.props.onCancel();
+        }}
+      >
+        {this.state.loading ? <Loading /> : 'Deleted Drafts are gone forever. Are you Sure?'}
```
bonustrack: We should translate any texts using `react-intl` and `FormattedMessage`.

src/post/Write/DeleteDraftModal.js
```diff
@@ -0,0 +1,75 @@
...
+    super(props);
+    this.state = {
+      loading: false,
+    };
+  }
+
+  deleteDraft() {
+    this.setState({ loading: true });
+    this.props.deleteDraft(this.props.draftId).then(() => {
+      this.props.notify('Draft have been deleted', 'success');
```
bonustrack: This text `Draft have been deleted` should be translated

src/post/Write/DeleteDraftModal.js
bonustrack: Can you change the text to `Are you sure you want to delete this draft permanently?`

Ne3l:
@bonustrack Changes applied.

src/components/Editor/Editor.js
```diff
@@ -605,6 +609,20 @@ class Editor extends React.Component {
                 <FormattedMessage id="saving" defaultMessage="Saving..." />
               </span>
             )}
+            <Form.Item className="Editor__bottom__cancel">
+              {draftId &&
+              <Button
+                type="danger"
+                disabled={loading}
+                onClick={(e) => {
```
Sekhmet: Can we use `handleCancel` method or something in like this instead? This creates new function on every re-render.

src/post/Write/DeleteDraftModal.js
```diff
@@ -0,0 +1,84 @@
...
+      <Modal
+        title={intl.formatMessage({
+          id: 'draft_delete',
+          defaultMessage: 'Delete this draft?',
+        })}
+        visible
+        confirmLoading={this.state.loading}
+        okText={intl.formatMessage({ id: 'confirm', defaultMessage: 'Confirm' })}
+        cancelText={intl.formatMessage({ id: 'cancel', defaultMessage: 'Cancel' })}
+        onOk={() => {
```
Sekhmet: Can we have `handleOk` and `handleCancel` here as well for the same reason?

src/post/Write/DeleteDraftModal.js
```diff
@@ -0,0 +1,84 @@
...
+    deleteDraft: () => {},
+    notify: () => {},
+    onDelete: () => {},
+    onCancel: () => {},
+    draftId: null,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
```
Sekhmet: Can you set state outside of constructor? Constructor isn't necessary in this case.

src/post/Write/DeleteDraftModal.js
```diff
@@ -0,0 +1,84 @@
...
+import { deleteDraft } from './editorActions';
+import { notify } from '../../app/Notification/notificationActions';
+
+@injectIntl
+@connect(null, {
+  notify,
+  deleteDraft,
+})
+class DeleteDraftModal extends React.Component {
+  static propTypes = {
```
Sekhmet: Can you sort props that functions are at the bottom? In propTypes and defaultProps. We tend to use this code styles everywhere.

src/post/Write/DraftRow.js
```diff
@@ -2,28 +2,30 @@ import React from 'react';
...
-    pending: PropTypes.bool,
-    deleteDraft: PropTypes.func,
   };
 
-  static defaultProps = {
-    pending: false,
-    deleteDraft: () => {},
+  constructor(props) {
+    super(props);
+    this.state = {
```
Sekhmet: Can we remove this constructor as well and use `state` outside of this as well?

src/post/Write/DraftRow.js
```diff
@@ -2,28 +2,30 @@ import React from 'react';
...
+      showModalDelete: false,
+    };
+  }
+  showModal = () => {
```
Sekhmet: Those two functions don't have to have body. It would look cleaner this way:
`X = () => this.setState({ showModalDelete: X });`

src/post/Write/Write.js
```diff
@@ -265,13 +286,24 @@ class Write extends React.Component {
...
               onSubmit={this.onSubmit}
+              onDelete={this.onDelete}
               onImageInserted={this.handleImageInserted}
             />
           </div>
+          {this.state.showModalDelete && (
+            <DeleteDraftModal
+              draftId={draftId}
+              onDelete={this.onDeleteDraft}
+              onCancel={() => {
```
Sekhmet: This should it's own handler as well.

src/post/Write/Write.js
```diff
@@ -60,6 +65,8 @@ class Write extends React.Component {
     createPost: () => {},
     saveDraft: () => {},
     notify: () => {},
+    replace: () => {},
+    draftId: null,
   };
 
   constructor(props) {
```
Ne3l: Should this state be also outside of constructor?

Sekhmet: I wouldn't change unrelated code for now.

Ne3l:
Resolved conflicts. Ready for review.

src/locales/en.json
```diff
@@ -168,11 +176,15 @@
...
   "image": "Add image",
   "get_boost": "Get the Busy boost!",
-  "get_boost_content": "Want to boost your post? Add the topic \"busy\" and get a like from us.",
+  "get_boost_content":
+    "Want to boost your post? Add the topic \"busy\" and get a like from us.",
   "drafts": "Drafts",
   "drafts_empty": "You don't have any draft saved",
   "draft_untitled": "Untitled draft",
   "draft_delete": "Delete this draft",
+  "draft_delete_success": "Draft have been deleted",
```
Sekhmet: It should be `Draft has been deleted`. We should change this here and in `defaultMessage` as well.

src/post/Write/DraftRow.js
```diff
@@ -34,14 +31,23 @@ class DraftRow extends React.Component {
...
-            <FormattedMessage id="draft_delete" defaultMessage="Delete this draft" />
-          </Button>
-        </div>
+        <Button type="danger" onClick={this.showModal}>
+          <FormattedMessage id="draft_delete" defaultMessage="Delete this draft" />
+        </Button>
+        {this.state.showModalDelete && (
+          <DeleteDraftModal
+            draftId={this.props.id}
+            onDelete={this.hideModal}
```
Sekhmet: We don't need to have `onDelete` handler there.
`hideModal` uses `this.setState` which is no-op on unmounted components. `DeleteDraftModal` deletes draft so `DraftRow` is unmounted before `hideModal` fires, causing this error:
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the DraftRow component.
```

Ne3l:
@Sekhmet  Changes applied.

PR #905, opened 2017-10-27
Add Avatar image error handler

jm90m:
* When img.busy randomly goes down, we need to replace avatar with a default avatar
* Add onError handler to img tag

@Sekhmet @bonustrack - this is a possible solution to when `img.busy` randomly goes down. What default avatar url should I use?

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -22,7 +22,7 @@ class Avatar extends Component {
 
   onError = () => {
     this.setState({
-      imgUrl: 'https://res.cloudinary.com/hpiynhbhq/image/fetch/s--hWKFE2sA--/c_fill,h_200,w_200/http://res.cloudinary.com/hpiynhbhq/image/upload/v1501526513/avatar_juzb7o.png',
+      imgUrl: 'https://res.cloudinary.com/hpiynhbhq/image/upload/v1506948447/p72avlprkfariyti7q2l.png',
     });
   };
 
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -1,23 +1,45 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import './Avatar.less';
 
-const Avatar = ({ username, size }) =>
-  (<img
-    className="Avatar"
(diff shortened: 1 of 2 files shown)


ryanbaer:
Was just about to take on this issue as well, thanks for putting this up. I'm sure all the users who mentioned it in the testing channel will appreciate it once it gets in.

One comment - if we've trying to mediate the issue of img.busy going down, I think the safest bet is to bundle the default avatar asset into the app, so the app is just loading a static file rather than making a network request to somewhere that may be on the same affected host that's down.

Sekhmet:
`http://img.busy.org/@steemconnect?s=200` is the default one.

jm90m:
@ryanbaer good point...

@Sekhmet if img.busy is down, wouldn't that default be down as well?

Sekhmet:
It would, I just referenced default image link that is used. You can just download it using this URL with proper size (not sure how big we need this image) and add it to assets.

bonustrack:
@jm90m The default avatar is this here https://github.com/busyorg/busy-img/blob/dev/helpers/index.js#L14 https://res.cloudinary.com/hpiynhbhq/image/upload/v1506948447/p72avlprkfariyti7q2l.png

src/components/Avatar.js
```diff
@@ -1,23 +1,45 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      imgUrl: `${process.env.IMG_HOST}/@${props.username || 'steemconnect'}?s=${props.size}`,
+    };
+  }
+
+  onError = () => {
+    this.setState({
+      imgUrl: 'https://res.cloudinary.com/hpiynhbhq/image/fetch/s--hWKFE2sA--/c_fill,h_200,w_200/http://res.cloudinary.com/hpiynhbhq/image/upload/v1501526513/avatar_juzb7o.png',
```
bonustrack: Can you use https://res.cloudinary.com/hpiynhbhq/image/upload/v1506948447/p72avlprkfariyti7q2l.png ?

jm90m: yeah just updated @bonustrack

PR #908, opened 2017-10-27
Add code of conduct and PR template

bonustrack:


Sekhmet:
I don't think that pull request should be used for proposing changes. I don't want people to start working on something before agreeing if it's needed and how it should work beforehand through issues. It's not any good to figure out someone spent couple hours or even days working something that is not needed.

I'd not change PR/issue templates before we have everything figured out and clear.

bonustrack:
The sentence `Changes proposed in this pull request:` is probably confusing. But the point of this PR template is to suggest dev to reference an issue with `Fixes # .`. The PR should fix an issue approved by maintainers, this way we are sure someone is not working on something we don't want.

Sekhmet:
PR doesn't propose changes, PR implements them.
I think that using just `Changes:` instead of `Changes proposed in this pull request:` are more clear.

bonustrack:
Or `Changes implemented:` ?

Sekhmet:
I don't think that adding `implemented` will be useful. Ideally contributors won't add PR when everything is done, but just after they start working on it, so not all changes would be implemented from start. It would be easier to create one category `Changes` like this:

Changes:
- [x] Some feature
- [x] Yet another feature
- [ ] Not finished feature yet

than to use `Changes implemented` and `Changes not implemented`.

bonustrack:
Ok i was expecting you would not want `implemented` and I changed to just `Changes:`

bonustrack:
We can do better docs but meanwhile i suggest that we merge this.

Issue #833, opened 2017-10-17
Topics overflow the bounding box if they are too long


espoem:
<img src="https://user-images.githubusercontent.com/16621517/31687460-97865e1c-b389-11e7-9e1c-37b2c59a33ef.png" height="280px"/>

<img src="https://user-images.githubusercontent.com/16621517/31687467-9d9a47d2-b389-11e7-9ac1-f54f8b4f76b5.png" height="280px"/>

#### Expected behavior

The topics should be always within the bounding box.

#### Actual behavior

When topics words are too long, they overflow the box. The box is fixed and doesn't change its size.

#### How to reproduce

1. Open editor
2. Input long topics so that any of them appears on the following line

* Browser: Chrome / Brave
* Operating system: Windows 10 / Android 7.1.2

eastmaels:
> The topics should be always within the bounding box.

Or should it be that the bounding box should automatically resize to encompass the tags?
Making the text (tags) be 'always within' the bounding box may mean resizing the font size of the tags.

bonustrack:
Fixed in #909

PR #909, opened 2017-10-27
set Editor__topics > div height to auto

eastmaels:
Screenshot of fix:

## Before

image

## After

image


eastmaels:
Fixes #833

bonustrack:
@eastmaels When there is only 1 line, the input is smaller than what we have currently on https://nd.busy.org/write. It should not be the case

Before:
image

After:
image

eastmaels:
Hello @Sekhmet , I added a 8px padding to bottom of topic div.

image

Issue #917, opened 2017-10-28
Recommended Posts should behave like regular links


ryanbaer:
#### Expected behavior
1. When using Cmd or Ctrl + Click on a post in the Recommended Posts pane, the link should open in a new tab
1. When right clicking on a post in the Recommended Posts pane, the native browser context menu should appear

tl;dr links in Recommended Posts should behave like regular links

#### Actual behavior
1. Cmd or Ctrl + Click is ignored, and the link is browsed to on the existing page
1. No context menu appears

#### How to reproduce
1. Navigate to a post page
1. Perform either of the above actions on one of the entries in the Recommended Posts page

* Browser: macOS 10.12.6
* Operating system: Chrome

PR #918, opened 2017-10-28
Use Link component instead of <a> tags in PostRecommendation component

jm90m:
Fixes #: https://github.com/busyorg/busy/issues/917

Changes:
- Use Link component

### DIFF
--- a/src/components/Sidebar/PostRecommendation.js
+++ b/src/components/Sidebar/PostRecommendation.js
@@ -12,7 +12,6 @@ class PostRecommendation extends Component {
   static propTypes = {
     location: PropTypes.shape().isRequired,
     intl: PropTypes.shape().isRequired,
-    history: PropTypes.shape().isRequired,
     isAuthFetching: PropTypes.bool.isRequired,
   };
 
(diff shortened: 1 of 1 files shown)


src/components/Sidebar/PostRecommendation.js
```diff
@@ -77,8 +75,8 @@ class PostRecommendation extends Component {
     }
   };
 
-  navigateToPostComments = (category, author, permlink) => {
-    this.props.history.push(`/${category}/@${author}/${permlink}#comments`);
+  navigateToPostComments = () => {
+    document.getElementById('comments').scrollIntoView();
     this.setState(this.state);
```
Sekhmet: What's the purpose of `this.setState(this.state)`? If it's to force re-render we should use `forceUpdate()` instead. Is there no other way around this than to force re-render? I have problems understanding what is going on in this component to be honest, but as far as I understood it I think we should re-render only when props or state changes, so we should implement this logic in `shouldComponentUpdate`.
This code got to our code base earlier and is not really related to this PR, so I won't use this PR for solving this problem, but I think we should improve logic there in another PR.

jm90m: It's to force an update. I read online that it was bad practice to use forceUpdate.

Sekhmet: Yes it is, but `this.setState(this.state)` is equally as bad or even worse as it would trigger more lifecycle methods. Bad practise is better than bad practise + hacking it around.

jm90m: For some reason I thought that forceUpdate would be slower then this.state, since it forces a re-render of all child components, but looking in deeper and testing it out, I think forceUpdate might faster, since it doesn't go through shouldComponentUpdate, will change.. thanks for the feedback

PR #922, opened 2017-10-28
Add Loading to Est Account & Steem Trending Charts

### LINKED ISSUE #929: SP amount in "Claim Rewards" transactions is not SP
#### Expected behavior

The number in Claim Rewards transactions should be actually SP.

image

#### Actual behavior

The number is much higher. It seems that it is VESTS (not confirmed).

image

#### How to reproduce

1. Go to wallet
2. Find "Claim Rewards" transaction
3. Compare with Steemit

* Browser: Chrome

jm90m:
Fixes #929.
* Est. account value doesn't show up

Changes:
- Add loading to Est Account Value
- Add loading to steem trending charts


bonustrack:
Sometime there is a `NaN` showed while loading. Also for the Estimated account value i don't see loading icon but `0.00$` i think the loading icon is less confusing.
image

bonustrack:
On small screen (when sidebar are not visible) we should remove border on left & right and border-radius of the boxes in center (similar than on the feed):
image

bonustrack:
Just spoted this also, something wrong with SP conversion for claim reward, i did not claim 54K SP :)
image

jm90m:
@bonustrack just made those fixes you requested

bonustrack:
@jm90m border radius should be removed too on small screen:
image

jm90m:
@bonustrack done, removed border radius for small screens

bonustrack:
You need to remove all of them.
image

bonustrack:
Also remove double border here:
image

jm90m:
@bonustrack good catch just removed

PR #923, opened 2017-10-28
Use busy.org instead of steemit.com for top search

jm90m:
@bonustrack I'm not sure if this was intentional but I noticed that our top search refers to steemit.com.... shouldn't it refer to busy.org?

### DIFF
--- a/src/components/Navigation/Topnav.js
+++ b/src/components/Navigation/Topnav.js
@@ -139,7 +139,7 @@ const Topnav = ({
         <div className="center">
           <div className="Topnav__input-container">
             <Input
-              onPressEnter={event => window.open(`https://www.google.com/search?q=${encodeURIComponent(`site:steemit.com ${event.target.value}`)}`)}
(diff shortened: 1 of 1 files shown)


bonustrack:
It was intentional because busy.org has less results, steemit.com results are more relevant.

mktcode:
Actually this is a serious issue because it pulls people away from the site. Are there any thoughts on this already? AskSteem?

jm90m:
@bonustrack ok I'll close this then

bonustrack:
@mktcode i've compared Google and AskSteem but Google with steemit.com is what giving better results for me. There could be multiple search choice but anyway it's temporary.

ryanbaer:
I actually had a similar question to @mktcode.

I looked into Google Site Search, and how one could maybe custom-style the results in an iframe. Google of course disabled this behavior, and even if we found a way to do it, it's obviously a bad idea to go against them.

And then I saw that Google Site Search is EOL in 6 months (April 2018). I don't know what Steemit plans to do about this. And their replacement (GCSE) is way too cost-prohibitive for something open-source.

If this is temporary, what's the longer term solution for Busy?

Btw if there's another issue or place where this is being discussed, let me know, since this PR is already closed.

PR #927, opened 2017-10-28
Add current voting power to UserInfo

mktcode:
In the left sidebar, when viewing a user's profile, the current voting power is now shown.

screenshot-2017-10-29 mkt - busy

In a future PR this should be implemented:

```
function calculateVotingPower(votingPower, lastVoteTime, precision) {
    let secondsPassedSinceLastVote = (new Date - new Date(lastVoteTime + "Z")) / 1000;
    votingPower += (10000 * secondsPassedSinceLastVote / 432000);

    return Math.min(votingPower / 100, 100).toFixed(precision);
}
```

I took this code from another project. It takes into account the replenishment of the voting power since the last vote.


src/app/Sidebar/UserInfo.js
```diff
@@ -57,6 +59,13 @@ const UserInfo = ({ intl, authenticated, authenticatedUser, user, ...props }) =>
               }}
             />
           </div>
+          <div>
+            <i className="iconfont icon-flashlight text-icon" />
+            {intl.formatMessage({
```
Sekhmet: Can you use `FormattedMessage` from `react-intl` instead of this?

src/app/Sidebar/UserInfo.js
```diff
@@ -57,6 +59,13 @@ const UserInfo = ({ intl, authenticated, authenticatedUser, user, ...props }) =>
...
+              id: 'voting_power',
+              defaultMessage: 'Voting Power',
+            })}: {currentVotingPower} %
```
Sekhmet: Can you use `FormattedNumber` from `react-intl` instead of this as well?
Like here: https://github.com/busyorg/busy/blob/493639ed919abc5a70f3b9a8384ff7dd7fde9ce5/src/components/Reactions/ReactionsList.js#L48-L51

mktcode:
I don't know if I use it correctly and I also don't know why I have to divide by 10000 but... it works.

src/app/Sidebar/UserInfo.js
```diff
@@ -57,6 +57,14 @@ const UserInfo = ({ intl, authenticated, authenticatedUser, user, ...props }) =>
               }}
             />
           </div>
+          <div>
+            <i className="iconfont icon-flashlight text-icon" />
+            <FormattedMessage id="voting_power" defaultMessage="Voting Power" />: <FormattedNumber
+              style="percent" // eslint-disable-line react/style-prop-object
+              value={user.voting_power / 10000}
```
jm90m: So to get the percent of something you divide it by the total... so is 10,000 the total voting power that a user can have? @Sekhmet @bonustrack might know the answer

Sekhmet: @jm90m 
Steem uses 0-10000 range (10000 being 100%). `react-intl` uses 0-1 range, thus we need to divide it by 10000.

mktcode: Ah ok.... so `style="percent"` converts 0-1 to 0-100? Can't find this conversion in the code.

Sekhmet: It just tells `react-intl` to render this as per cent.
https://github.com/yahoo/react-intl/wiki/Components#formattednumber

PR #931, opened 2017-10-29
Update draft saving logic

### LINKED ISSUE #819: Draft saves multiple times
drafts

#### Expected behavior

Post draft should be saved only once.

#### Actual behavior

Sometimes, the draft is saved multiple times. There is a delay so the drafts are not totally identical. This may also have a side effect. When you want to delete one draft, more drafts might be deleted at once.

#### How to reproduce

1. Click "Write post"
2. Start typing __slowly__
3. Multiple drafts are saved. You can check URL being changed several times.

* Browser: Chrome 61.0.3163.100

### LINKED ISSUE #844: Clear fields when click "Write post" while a draft is open
#### Expected behavior

All fields should be cleared when clicked __Write post button__ while a draft is open.

#### Actual behavior

Writing new post while a draft is open changes the URL so a new post is really prepared to be created. All the content remains from the draft that was previously open.

#### How to reproduce

1. Open a draft
2. Click __Write post button__
3. New post is created filled with content from the previous draft

* Browser: Chrome

Sekhmet:
Fixes #819, fixes #844

Changes:
- [x] Generate `draftId` when editor opens.
- [x] Redirect to `write?draft=${post.id}` before saving draft to SteemConnect metadata.
- [x] Reset form when user clicks on Write icon.

@espoem could you check if everything works properly?


Sekhmet:
@espoem confirmed that this PR fixes both those issues.

src/post/Write/Write.js
```diff
@@ -105,14 +105,33 @@ class Write extends React.Component {
         isUpdating: isUpdating || false,
       });
     }
+
+    if (draftId) {
+      this.draftId = draftId;
+    } else {
+      this.draftId = Date.now().toString(16);
```
jm90m: If 2 people are saving there drafts at same time, won't they have the same draftId?

Sekhmet: They would, but it would only matter if two people on the same account would open editor exactly at the same time. We could use `uuid` for this, just in case :D

PR #932, opened 2017-10-29
Add avatar lightbox

### LINKED ISSUE #926: Click on profile avatar should open a modal with a bigger image
In some cases it might be useful to be able to open bigger preview of user avatar. We might add modal with preview of avatar (with higher resolution) after clicking on users' avatar on their profile page.

Sekhmet:
Closes #926 

Changes:
- Add `AvatarLightbox` component that renders avatar and lightbox for that avatar with given image size.

avatar-lightbox


bonustrack:
@Sekhmet Isn't better to use https://ant.design/components/carousel/ for it and remove `react-image-lightbox` in favor of antd for single post too?

Sekhmet:
I don't think this would be good candidate for lightbox, as it is designed as carousel.
Lightbox renders on top of everything by default and supports zoom and pan, which is helpful for big images. If we are going to switch to carousel we would have to implement those features ourselves.

src/components/AvatarLightbox.js
```diff
@@ -0,0 +1,45 @@
...
+
+export default class AvatarLightbox extends React.Component {
+  static propTypes = {
+    username: PropTypes.string,
+    size: PropTypes.number,
+    previewSize: PropTypes.number,
+  };
+
+  static defaultProps = {
+    username: undefined,
```
jm90m: Would it be better to use the default avatar username here? I guess it doesn't matter if it fails on that request then it goes to avatar default image anyways. So your call :)

Sekhmet: `@steemconnect` uses different avatar than we use as error fallback. I don't know whether there is username that has avatar set to that default avatar.

bonustrack: Wrong username got default avatar example https://img.busy.org/@xxxx123456 . We should not rely on @steemconnect avatar, at some point we will want to have the logo of SC as avatar.

Sekhmet:
I've updated `Avatar` so it uses `defaultImage` instead of `steemconnect` as username if `username` is `undefined`.

bonustrack:
@Sekhmet The image size should not be bigger than the original and also should keep aspect ratio. You can do this by changing the crop style like this:
http://img.busy.org/@blocktrades?crop=limit&s=800
http://img.busy.org/@fabien?crop=limit&s=800

Issue #924, opened 2017-10-28
Pasting Data into Tags Puts Draft into Unrecoverable State


ryanbaer:
#### Expected behavior
The Tag Editor should strongly validate what can be inputted, and should not `PUT` to `/me` if the client-side validation fails

#### Actual behavior
1. Pasting a large enough string of data puts the draft into an unrecoverable state
1. Validation Errors on the client-side display an error but still `PUT` data to `/me

![oct-28-2017 12-59-21](https://user-images.githubusercontent.com/11914327/32138011-070048f8-bbe0-11e7-9742-8e7f3ecb055c.gif)

#### How to reproduce
1. Create a new draft
1. Enter a title
1. Enter some article content so the draft is saved (i.e. `PUT` to `/me` occurs)
1. Get a large string of text. A good example is the entire Markdown of another article
1. Click in the Tags input and press Ctrl or Cmd +V to paste in your clipboard data
1. Note that you get a validation errror, the tags box remains blank, and you can no longer interact with it
1. Note that you cannot publish your draft
1. If watching Network Requests, note that `jsonMetadata.tags` for your draft is now a very large array - in my case, it looks as follows:
![image](https://user-images.githubusercontent.com/11914327/32137913-e0619870-bbdd-11e7-9967-057b395eb538.png)

I've already started looking into this one, so feel free to assign to me.

* Browser: Chrome 61
* Operating system: macOS 10.12.6

ryanbaer:
This issue can now be closed

Sekhmet:
Fixed in #945.

PR #945, opened 2017-10-30
Sync Drafts more Carefully & Upgrade AntDesign

ryanbaer:
## Pasting Topics can Break a Draft

This PR addresses #924, in which a draft is still synced to SteemConnect in spite of validation errors with the form - particularly the `topics` field. Once synced, the bad data perpetuates the problem since it is repopulated on page load.

## Root Cause
**Note: if you don't care about the Root Cause, scroll down to the Fixes section**

### Issue 1: No Validation Checks before Syncing Drafts

The syncing in spite of errors results in a bug in which:
- If provided with the wrong input, the topics field breaks and becomes completely blank, resulting in a dead-end for fixing the problem on the page as is - i.e. no backspacing or deleting tags because the Select component breaks and renders as empty.
- Because the invalid input is still synced, upon reload, the post draft itself becomes unrecoverable. This is due to the bad input being retrieved from SteemConnect and populated into the topics field again, resulting in the same error on each load.
- The error with the topics field prevents any actual submission of the post, so the user must start over with a fresh post, copying their data from the broken draft.

Here's what I found.

All updates to drafts go through the `onChange` method in `Editor.js`, which triggers a dispatch that syncs any changes from the form to SteemConnect.

First, there was no validation taking place in this method. If any value was changed in the form, it was immediately synced; even, for example, with 6 or more topics.

As a result, the form's validation methods effectively become suggestions to the user when saving a draft, with the hard-stop being that you can't actually publish the post if there are the validations that fail.

Since it's a pretty simple form, the only other issue I found with not honoring the validation errors was with the title, which has a max length of 255.

So Issue 1 is resolved by ensuring that we check for validation errors on the two relevant fields before syncing, and refuse to do so if there are any.

However, resolving Issue 1 doesn't address the issue of being able to paste bad data into the tags field and break the Select component.

For the record, the scenario that breaks the Select is:
- Pasting in any non-separator character, followed by two or more separator characters, followed by another non-separator character.

### Issue 2: Bug in AntDesign's Select Component (via [react-component/select](https://github.com/react-component/select))

The bug mentioned above does not come from our code, but from AntDesign's. And ultimately, it stems from its dependency, [react-component/select](https://github.com/react-component/select). It is reproducible with a CodePen for AntDesign, and with the Demo Page for rc-select:
- [CodePen: Ant Design Select Bug](https://codepen.io/ryanbaer/pen/OOJBWO)
- [Demo : react-component/Select Bug](http://react-component.github.io/select/examples/tags.html)
   - Paste in the following input: `tag1,,,,,,,,,tag2`
   - Notice that a space appears between the two elements, and if inspecting the HTML, notice that a blank tag has rendered there

Since I had already found my way to the root cause of the issue, I decided to fork `rc-select` and make the changes to fix it, creating a pull request. I don't know how long it will take them to review, or if they'll deem my changes acceptable, but I'll keep abreast of the situation.

In the meantime, I don't think there's much we can do about preventing users from pasting bad data into the topics field. I also think it's incredibly unlikely that it'll happen. I had only previously seen one user on Discord complain of having the topics field be broken.

The fix for Issue 1 should mitigate the issue enough that the user can simply refresh the page and get back to writing their post.

### (Related) Issue 3: Our Usage of `onChange`
This is more of a related issue that I need some clarification on. It's about our usage of `onChange` on antd's `Form` component. If my point ends up being moot, we can just close the issue. You can view the issue here: https://github.com/busyorg/busy/issues/946

## Fixes
In viewing these changes, please divide up into two categories:
- The validation fix
- Cleanup for bad data

### The Validation Fix
The validation fix involves changes to `onUpdate` and `checkTopics`, as well as the addition of `checkTopicsValidator`. `checkTopics` was broken up so it could be invoked by the form's validator (via `checkTopicsValidator`) as well as manually in our `onChange` function.

You may notice that validation for topics in `onChange` seems a little redundant, since it happens again shortly after. However, this seems to be caused by what I mentioned above in Issue 3, with how I'm not so sure we're supposed to be using `onChange` on these fields.

As a result, the fix does require that check in `onChange`, but only because `onChange` actually completes (and hence syncs the draft) before validation runs on the form with the most current data. This at least ensures that, if there were any validation errors, the form won't be synced to drafts.

### Cleanup for Bad Data
From stumbling onto this bug, and from testing it, I had several drafts that were in the unrecoverable state. I wrote some additional code to handle this scenario that will actually recover the drafts, putting them back into a usable state.

**Whether or not to include this behavior is for discussion.**
These changes consist of:
- Adding `filterBadData` to fix any invalid `topics` data and truncate `title` to 255 chars or less
- Invoking `filterBadData` from `componentWillReceiveProps`
- A fix for what I believe is an additional issue in `componentWillReceiveProps` in which two arrays were undergoing a strict equality check and hence would not return true even if they contained the same entries.


ryanbaer:
### Update
`rc-select` [accepted my PR](https://github.com/react-component/select/pull/239)

I've just pushed changes in `package.json` and `package-lock.json` containing the update to `antd` and subsequently `rc-select`. It is now no longer possible to break their component by pasting in bad input.

ryanbaer:
After taking a break from this for a bit, I think that it's actually undesirable behavior to have the editor stop syncing for validation errors. It would be pretty frustrating to keep working on a draft, close the window, and realize that all your changes are gone because you had one too many tags.

So the new change is to strip away a lot of my modifications. Now `onUpdate` simply slices off anything greater than 255 chars for the title, and anything greater than 5 topics. Everything syncs on update, and if you happen to go over the limit, it won't sync the extra invalid data.

ryanbaer:
Btw, now that this is being reviewed, I know that JSON.stringify is a bit of a primitive method for determining equality of two arrays, so if you have a preferred library function or other method, feel free to suggest.

src/components/Editor/Editor.js
```diff
@@ -91,9 +91,10 @@ class Editor extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     const { title, topics, body, reward, upvote } = this.props;
+
     if (
       title !== nextProps.title ||
-      topics !== nextProps.topics ||
+      JSON.stringify(topics) !== JSON.stringify(nextProps.topics) ||
```
Sekhmet: As you said `JSON.stringify` is not the best option there. Can you use lodash's `isEqual` method for comparsion?

ryanbaer: Added

PR #953, opened 2017-10-30
Add Claim Rewards Button

jm90m:
* https://github.com/busyorg/busy/projects/6#card-5364055
screen shot 2017-10-30 at 7 53 20 pm

Fixes # .
https://github.com/busyorg/busy/projects/6#card-5364055
Changes:
- Add Claim Rewards Button

@bonustrack whats a good way to test this?


jm90m:
Updated to display Rewards in Current User's Wallet Summary

screen shot 2017-10-31 at 12 10 48 pm

jm90m:
Updated with new design

screen shot 2017-11-01 at 12 29 34 pm
screen shot 2017-11-01 at 12 29 17 pm

bonustrack:
@jm90m When reward are claimed the reward balance should be updated and disappear as they will be empty. Also i feel strange the button "transfer" in middle, i would rather bring in to the top, and use purple color and use grey color for claim reward button.

jm90m:
screen shot 2017-11-01 at 5 19 07 pm
@bonustrack Updated with new button colors and claimed reward balance gets hidden after claiming the reward

claim-rewards-gif

bonustrack:
@jm90m I think the box should be completly removed when reward claimed. Now i still see the box but empty.
Also the last line of balance (see in green) should not be visible. And the padding in top and bottom should be same height (see red lines)
image

jm90m:
updated the styling here, changed padding @ top from 8 to 10px since padding at bottom is already 16px, but top also includes title padding with it.

src/app/Sidebar/UserInfo.js
```diff
@@ -75,6 +75,7 @@ const UserInfo = ({ intl, authenticated, authenticatedUser, user, ...props }) =>
         defaultMessage: 'Transfer',
       })}
       onClick={() => props.openTransfer(user.name)}
```
Sekhmet: Can you add handler for `onClick`? This creates new function ever time `UserInfo` re-renders.
We usually put boolean type props first (`primary`).

src/wallet/Wallet.js
```diff
@@ -19,15 +20,17 @@ const Wallet = props => (
...
         <div className="right">
-          <SteemTrendingCharts />
           <Action
-            style={{ margin: '10px 0' }}
+            style={{ marginBottom: '10px' }}
             text={props.intl.formatMessage({
               id: 'transfer',
               defaultMessage: 'Transfer',
             })}
             onClick={() => props.openTransfer('')}
```
Sekhmet: I think that we can use handler and move `primary` up here as well, as it's not a big change and somewhat related to this PR.

jm90m:
@Sekhmet just updated UserInfo and Wallet components, is that what you mean by add handler?

Issue #957, opened 2017-10-31
Voting power value in profile is wrong


gregory-latinier:
#### Expected behavior

The voting power display in the profile should be calculated using this formula:

```
const secondsago = (new Date().getTime() - new Date(account.last_vote_time + "Z").getTime()) / 1000;
const votingPower = account.voting_power + (10000 * secondsago / 432000);
```

#### Actual behavior

The voting power value is not correctly calculated in the profile page

![image](https://user-images.githubusercontent.com/13222767/32213119-43dca014-be1a-11e7-9a4a-2a697f6b9a3f.png)

#### How to reproduce

Vote for a content and wait a few minutes for the voting power to replenish
If you were at 100% at the moment of the vote, you should be at 98.01% for a 100% power vote after a while

* Browser: any
* Operating system:  any

mktcode:
This is totally true. I mentioned the issue in the original PR https://github.com/busyorg/busy/pull/927
Even sites like steemstats.com or steemdb.com show wrong values (if they didn't fixed it already).
I'm pretty sure there will be a PR fixing this soon. :)

jm90m:
Updated Voting Power Calculation based on chat on our public discord @gregory-latinier :

![screen shot 2017-10-31 at 11 16 24 am](https://user-images.githubusercontent.com/5067716/32231826-0c5dd172-be2d-11e7-8ae1-d32a49eeda59.png)

jm90m:
Fixed in #961

PR #961, opened 2017-10-31
Update Voting Power Calculation

jm90m:
* https://github.com/busyorg/busy/issues/957

Fixes # .
https://github.com/busyorg/busy/issues/957
Changes:
- Updating Voting Power based on as discussed on here: https://github.com/busyorg/busy/issues/957


jm90m:
@cron9 no it was not intentional, good catch, I didn't push up latest changes to this branch

bonustrack:
@jm90m I have lot of voting power, do you see same?
image

jm90m:
@bonustrack was dividing by 100, instead of 10,000. sry fixed now

Issue #928, opened 2017-10-29
User can't see the whole history in wallet


espoem:
#### Expected behavior

User should be able to see whole history of his transactions.

![image](https://user-images.githubusercontent.com/16621517/32141725-8deaa352-bc87-11e7-9f62-ff88b09fb967.png)

#### Actual behavior

Wallet shows a lot of records but they are not all.

![image](https://user-images.githubusercontent.com/16621517/32141728-92a669c6-bc87-11e7-9d93-73121c1388ff.png)

#### How to reproduce

1. Go to wallet.
2. Scroll to the end. It won't fetch other records.
3. Compare with Steemit.

* Browser: Chrome
* Operating system: Windows 10

jm90m:
Fixed in https://github.com/busyorg/busy/pull/976. Closing this but If you re-test this issue, and there is a bug, you can re-open or create a new one.

PR #965, opened 2017-10-31
Use getStateAsync instead of getAccountHistoryAsync for User Wallet Transactions

jm90m:
So I've been thinking about this, right now we getAccountHistoryAsync, but when we use that, we have to filter through every single user action just to get the wallet transactions. So even though we get 2500, sometimes it won't be all of the user's wallet transactions... so how would we know when the user has more wallet transactions?

Also for this, I tried to use React Infinite: (https://github.com/seatgeek/react-infinite) 
* but it has issues when rendering element sizes of dynamic heights (https://github.com/seatgeek/react-infinite/issues/231)

After trying to get React Infinite to work, I settled on using what we currently have, `ReduxInfiniteScroll`.

Let me know your thoughts @Sekhmet @bonustrack  

Fixes # 
https://github.com/busyorg/busy/issues/928

Changes:
- Use getStateAsync instead of getAccountHistoryAsync to get user transaction history

### DIFF
--- a/src/helpers/apiHelpers.js
+++ b/src/helpers/apiHelpers.js
@@ -118,9 +118,12 @@ const isWalletTransaction = actionType =>
   actionType === 'claim_reward_balance';
 
 export const getTransactionHistory = account =>
-  SteemAPI.getAccountHistoryAsync(account, -1, 2500).then(results =>
-    _.compact(
-      results.map((action) => {
+  SteemAPI.getStateAsync(`@${account}/transfers`).then((results) => {
+    const transferHistory = results.accounts[account]
+      ? results.accounts[account].transfer_history
+      : [];
+    return _.compact(
+      transferHistory.map((action) => {
         const actionDetails = action[1] || { op: [] };
         const actionType = actionDetails.op[0];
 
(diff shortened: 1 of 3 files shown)


bonustrack:
@jm90m It was intentional to use getAccountHistory instead of getState. getState dont scale, there is no pagination and can give a too much big file to download for users. That's not conveninant at all for mobile. Also with getAccountHistory we have all data, which make it possible to create page "Activity" in same time and get data for notifications later.

Sekhmet:
We already got `react-infinite-scroller`. Would it work?

jm90m:
@bonustrack The problem I faced with `getAccountHistory` is that there is no way to tell when a user has anymore wallet transactions. So for example on the first request to `getAccountHistory`, I get the first 2500 records, but say only 4 of them are wallet transactions. That doesn't fill the page all the way down to the infinite scroll portion. So how would we know when to fetch more records?

jm90m:
@Sekhmet yeah I think `react-infinite-scroller` will work but is that better than the `redux-infinite-scroll` ? I see the ReduxInfiniteScroll is used in 2 other places (Feed / UserList)

bonustrack:
@jm90m We should use `getAccountHistory` it doesnt matter if some requests dont give results, if there is more than `x` empty result requests you can stop the loop.

jm90m:
@bonustrack alright thats fine with me, I'll keep looking into making it work with getAccountHistory, closing this and moving to new branch.

Issue #883, opened 2017-10-22
Mobile responsive: Avatar doesn't show in comment conversation


espoem:
#### Expected behavior

Avatar shows next to each comment.

<img src="https://user-images.githubusercontent.com/16621517/31861535-3b935390-b72f-11e7-9512-c121a2ccfb46.png" height="280px"/>

#### Actual behavior

Avatar shows only next to the first comment in comment conversation for each user on mobile (screen/window width smaller than 768px).

<img src="https://user-images.githubusercontent.com/16621517/31861544-624b55fa-b72f-11e7-8ab5-e16b65a949e1.png" height="280px"/>

#### How to reproduce

1. Go to `/travel/@sam.hsuu/sunscreen-is-for-the-weak-and-that-incident-at-a-strip-club` or `/utopian-io/@elear/utopian-now-pays-moderators-in-steem-power-profit-by-assuring-quality` on mobile
2. See comments
3. Check on desktop that avatars show next each comment

* Browser: Brave
* Operating system: Android 7.1.2
* Model: Moto G3 2015

Sekhmet:
This has been fixed in #969.

PR #969, opened 2017-11-01
Fix for hidden avatars near comments on small screens 883

sirrius-a-b:
Fixes # .
- for issue #883 

Changes:
- un-hide avatars in the comments on smaller screens by
- setting overflow:unset; for .Comment__text on small screens


jm90m:
Hey thanks for the contribution :D!

So this might not work... as it is related to this issue right here: https://github.com/busyorg/busy/issues/771

The `overflow: hidden;` was added to fix comments with code blocks from not scrolling. 

Test out your fix here: 

`test/@hellosteem/5ihr1p-test#@hellosteem/re-hellosteem-5ihr1p-test-20171018t132820462z`

and look into the comments with codeblocks, the scroll won't appear all the way with this.

sirrius-a-b:
I see, well, the `overflow:hidden;` does prevent those comments from scrolling but they get cut and are behind the right menu. I think #771 should maybe be reopened.

This can be better seen in the screen shot below. Notice how long the comment `code` tag actually is when I highlight it:
comments-behind-menu-small

sirrius-a-b:
I am looking for a fix to satisfy both cases.

sirrius-a-b:
Now it is fixed, the comments do not overflow or scroll and the avatars are showing up. Give it a good test though and let me know if anything.

jm90m:
@bonustrack can we get a deploy on this so we can test?

bonustrack:
Yes, on the way

jm90m:
@sirrius-a-b So your current fix removes the scrollbar in comment code blocks, we want to retain the scrollbar. 

screen shot 2017-11-01 at 3 55 23 pm

src/components/Comments/Comment.less
```diff
@@ -11,6 +11,14 @@
   }
 }
 
+.Comments > .Comment {
+    overflow:hidden;
```
jm90m: should be `overflow: hidden` with a space

src/components/Comments/Comment.less
```diff
@@ -11,6 +11,14 @@
...
+}
+
+.Comments > .Comment > .Comment__text {
+    width:calc(~'100% - 52px');
```
jm90m: this seems to work, but I'm not too sure if hardcoded values would be right.

src/components/Comments/Comment.less
```diff
@@ -50,9 +58,14 @@
   &__text {
     position: relative;
     top: -4px;
-    width: 100%;
+    width: calc(~'100% - 44px');
```
jm90m: it might be good to put this in `common.less` and accompany it with the avatar width:

```
@default-avatar-size: 40px;
@comment-text-offset: @default-avatar-size + 4px;
```

But see in `https://github.com/busyorg/busy/blob/new-design/src/components/Comments/Comment.js#L252` ... avatar size can be 40px or 32px, and I think default avatar in the `Avatar.js` component is 34px;

src/components/Comments/Comment.less
sirrius-a-b: It is `52px` because the zero level comment has `.Avatar` set at `40px` and `.Comment__text` has `margin-left:12px;` , together making 52. I will look on how to not be hardcoded.

jm90m: thanks! great fix though :)

src/components/Comments/Comment.less
sirrius-a-b: Should I not better put in in `custom.less` instead of `common.less` ? I see that in `custom.less` are most of the variables. Please advise.

jm90m: yeah thats what i meant sorry, `custom.less`

sirrius-a-b: Thanks, I will update it shortly.

sirrius-a-b:
I implemented all @jm90m recommendations to make it not be hard coded by adding vars for comments in `custom.less` and used them in `Comment.less` . I also added a space after the `:` on a few occasions as it has been suggested. I will remember this code convention now.

src/styles/custom.less
```diff
@@ -55,3 +55,10 @@
 @layout-trigger-height : 48px;
 @layout-zero-trigger-width : 36px;
 @layout-zero-trigger-height : 42px;
+
+// Comments
+@comment-text-left-margin: 12px;
```
jm90m: This is great, thanks for providing more context on what the values actually mean. @bonustrack what do you think of this fix? It seems to be working for me.

PR #973, opened 2017-11-02
Use FormattedNumber instead of USDDisplay component for est account

jm90m:
Fixes # .
* Est account value in User Wallet Summary was using the USDDisplay component which forced the "en" locale on the Formatted Number
Changes:
- Changed the display component of est account value to use FormattedNumber instead of USDDisplay


bonustrack:
@jm90m I think we use `FormattedNumber` in `USDDisplay` directly so the problem is solved across the app. For example the reward are currenctly been showed with `en` syntax.

Sekhmet:
Originally USD amounts were displayed using `FormattedNumber`, but it was requested to use hardcoded `en` locale for this (#688). I think that we should use localised `FormattedNumber` inside `USDDisplay` but just add `$` prefix like in this PR. This would make number style consistent and still have $ at the beginning.

PR #976, opened 2017-11-02
Add Infinite Scroll to User Wallet Page & Refactor getAccountHistory parsing

jm90m:
Fixes # .
- https://github.com/busyorg/busy/issues/928
- https://github.com/busyorg/busy/projects/6#card-5384665
Changes:
- Add Infinite Scroller, (Used Redux Infinite Scroller since React Infinite Scroller did not have a loader option)
- Refactor the parsing of getAccountHistory, to fetch all account histories and create 2 maps, one for all of the users account history and another for just user wallet transactions.


bonustrack:
LGTM, what you think @Sekhmet ?

src/helpers/apiHelpers.js
```diff
@@ -108,29 +107,9 @@ export const mapToId = (content) => {
...
 
-const isWalletTransaction = actionType =>
-  actionType === 'transfer' ||
-  actionType === 'transfer_to_vesting' ||
-  actionType === 'cancel_transfer_from_savings' ||
-  actionType === 'transfer_from_savings' ||
-  actionType === 'transfer_to_savings' ||
-  actionType === 'delegate_vesting_shares' ||
-  actionType === 'claim_reward_balance';
+export const defaultAccountLimit = 2500;
```
Sekhmet: Isn't it too big? When it does fetch 2500 transactions my computer is having problems rendering them and scrolling is lagging. I would say ~100 or even smaller would be better for this.

src/wallet/UserWalletTransactions.js
```diff
@@ -1,114 +1,169 @@
...
-      <FormattedNumber
-        value={transactionAmount}
-        minimumFractionDigits={3}
-        maximumFractionDigits={3}
-      />
-      {` ${transactionCurrency}`}
-    </span>
-  );
-};
+class UserWalletTransactions extends Component {
```
Sekhmet: I think we use `React.Component` everywhere instead of importing `Component` directly.

src/wallet/UserWalletTransactions.js
```diff
@@ -1,114 +1,169 @@
...
+    this.state = {
+      userHasMoreTransactions: firstTransaction.actionCount > defaultAccountLimit,
+      lastActionCount: firstTransaction ? firstTransaction.actionCount : 0,
+      loadingMoreTransactions: false,
+    };
+  }
+
+  getFormattedTransactionAmount = (amount, currency) => {
+    const transaction = amount.split(' ');
+    const transactionAmount = parseFloat(transaction[0]).toFixed(3);
```
Sekhmet: Do we need `toFixed`? Wouldn't `FormattedNumber` do the same?

src/helpers/apiHelpers.js
jm90m: So since we are fetching all account history, it would be good for it to be a large number since sometimes, we don't get all the wallet transactions enough to fill the page  :(.. When I did the wallet page I played around with the numbers 1000, 2500, 5000, and even 10,000. But after discussing with @bonustrack we settled on 2500.

src/wallet/UserWalletTransactions.js
jm90m: yeah we do, will update now

src/wallet/UserWalletTransactions.js
jm90m: yeah you're right, it's unnecessary here. good catch

jm90m:
@Sekhmet updated based on your feedback, I think we should keep it @ 2500 transactions, since we have to filter out the wallet transactions from all account history, it gives some room to for wallet transactions to fill out the page, and it's a reasonable size rather than 10,000. I think 100-1000 is too small of a range.

PR #989, opened 2017-11-03
Update TopicSelector styles

### LINKED ISSUE #988: Long topics break TopicSelector sort selector
#### Expected behavior

Long topics shouldn't cause sort selector to break. This should be taken care of for reasonable screen sizes and topic lengths.

#### Actual behavior

image

Sort selector renders improperly.

#### How to reproduce

Check feed with long topic on smaller screens below ~400px.

Sekhmet:
Fixes #988.

This changes how TopicSelector behaves on smaller screens.
- TopicSelector items are now aligned in center
- "Sort by" and current sort now have white-space set to nowrap to prevent
breaking it.
- "Sort by" and current sort now have display set to inline-block to
cause them to split to two lines if there is no room.

**Before**:
image
**After**:
image

### DIFF
--- a/src/components/TopicSelector.less
+++ b/src/components/TopicSelector.less
@@ -3,6 +3,7 @@
 .TopicSelector {
   display: flex;
   justify-content: space-between;
+  align-items: center;
   margin-bottom: 16px;
   padding: 0 @mobile-padding;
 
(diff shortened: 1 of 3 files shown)


src/components/TopicSelector.less
```diff
@@ -20,11 +21,18 @@
 
   &__sort {
     color: #99aab5;
+    text-align: right;
+
+    span:first-of-type {
```
jm90m: Would it be better to use a className here rather than selecting the first span? Maybe use &__label for `className="TopicSelector__sort__label"` or `className="TopicSelector__sort__title"` or something else similar?

PR #992, opened 2017-11-03
Remove storybooks

### LINKED ISSUE #925: Remove Storybook and replace them with regular snapshots
We don't use storybooks anymore and we are only using snapshots for tests. We have plenty of stories, with almost all components. We can't add stories with components that include containers (it would help us with removing lots of props being passed down 2-5 components). Additionally storyshots require additional logic in our code (special handling of environment variables for example). I think we should remove storybook and use snapshots for some components (only those that could benefit from this).

Components that should have snapshots:
- [x] `Follow`
- [x] `Action`
- [x] `Topic`
- [x] `Topics`
- [x] `InterestingPeople`
- [x] `LatestComments`
- [x] `UserMenu`
- [x] `TopicSelector`

Do you think I should add snapshots for something more? I don't mention components that already use snapshots. This is just a part of components from Storybook.

Sekhmet:
Fixes #925

Changes:
- Removed Storybooks and related code
- Added regular snapshot tests for some components: 
    - [x] `Follow`
    - [x] `Action`
    - [x] `Topic`
    - [x] `Topics`
    - [x] `InterestingPeople`
    - [x] `LatestComments`
    - [x] `UserMenu`
    - [x] `TopicSelector`


jm90m:
This looks fine to me, but I noticed that you named your test files the same as the component names. For my tests that I added for wallet page I named them with a `.test.js` as the file extension, so for `UserWalletSummary`, I had it as `UserWalletSummary.test.js`. Do you think we should name them with file extension like `.test.js ` or `-test.js`? Or keep it same as component name?

Sekhmet:
Jest watches files that are inside `__tests__` or `__specs__` or files that have `.test.js` or `.spec.js`.
Adding this suffix to files already inside `__tests__` folder seems a bit too verbose.

jm90m:
The reason why I added those file extensions, is when I do a quick search for filename it doesn't display directory path, and only file names in WebStorm. :(

screen shot 2017-11-03 at 11 02 49 am

But I guess it's fine without them

PR #993, opened 2017-11-03
Add babel-polyfill

### LINKED ISSUE #845: New busy not working in Internet Explorer
#### Expected behavior

New busy should work in Internet Explorer.

#### Actual behavior

The page is blank. No content appears.

#### How to reproduce

1. Go to new busy in Internet Explorer

* Browser: Internet Explorer 11

Sekhmet:
Fixes #845

This solves problem with Busy not working in IE11. It works on production builds, not on development builds.
Supporting IE < 11 is more problematic, because of the lack of Web Cryptography API. I don't think we should try to make it work on IE 10 and below as it is not used anymore, insecure and using polyfills for something as important as cryptography might introduce vulnerabilities.
Changes:
- Add `babel-polyfill` to the entry of the app.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,4 @@
+import 'babel-polyfill';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -34,9 +35,9 @@ if (process.env.SENTRY_PUBLIC_DSN) {
 }
 
 if (
-  process.env.STEEMCONNECT_CLIENT_ID
-  && process.env.STEEMCONNECT_REDIRECT_URL
-  && process.env.STEEMCONNECT_HOST
+  process.env.STEEMCONNECT_CLIENT_ID &&
```
jm90m: I see prettier autoformatted this file :rofl:

Sekhmet: I'm not sure if it's a bad thing. We have to reformat files at some point.

jm90m:
this looks fine to me, I use a mac so I can't test unless I dual boot / use remote desktop. Everyone should just switch to chrome / firefox anyways when you're dealing with crypto websites that hold sensitive information

PR #1001, opened 2017-11-03
Use production and development index files

### LINKED ISSUE #997: Unify HTML templates
We currently have different types of index.html files for `webpack-dev-server` and development and production builds in two different folders. We should use only one.

Sekhmet:
Fixes #997

Changes:
- `dist/` folder has been deleted
- `templates/index.html` is used for development purposes
- `templates/production_index.html` is used for production


bonustrack:
Isn't more common to have a `dist` folder instead of `templates`?

Sekhmet:
- dist/ folder has been deleted
- templates/index.html is used for development purposes
- templates/production_index.html is used for production

PR #1003, opened 2017-11-03
Update webpack-dev-server.js

### LINKED ISSUE #943: Can't run dev server
When I run `npm run dev-server` I get this error:

busy-install-issue

When I fix [webpack-dev-server.js](https://github.com/busyorg/busy/blob/new-design/webpack/webpack-dev-server.js#L24) by removing the commas, ESLint complains about the missing trailing commas.

busy-install-issue2

What shall I do? Submit PR with removed commas or not?

Node 6.11.5

Kennybll:
Fixes #943.

These two commas stopped "npm run dev-server" from running correctly.

### DIFF
--- a/webpack/webpack-dev-server.js
+++ b/webpack/webpack-dev-server.js
@@ -21,10 +21,10 @@ module.exports = {
         SENTRY_PUBLIC_DSN: null,
         STEEMCONNECT_CLIENT_ID: JSON.stringify(process.env.STEEMCONNECT_CLIENT_ID || 'busy.app'),
         STEEMCONNECT_REDIRECT_URL: JSON.stringify(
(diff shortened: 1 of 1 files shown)


Kennybll:
So I see in version 8.9.0 this is not a problem. But in older versions the commas deleted crash the dev-server. I also tested on version 8.9.0 without commas, and it still works.

Sekhmet:
Please use pull request template next time.

bonustrack:
@Sekhmet do you approve this?

Ne3l:
Unless we modify the eslint rule for trailing commas or ignore this file, this change gonna be a patch. The next time we touch this file is gonna put commas again. I would look for a long term solution.

PR #1007, opened 2017-11-04
Merge previous jsonMetadata when saving post

### LINKED ISSUE #914: Editing a post overwrite post json_metadata
When posting on DSound or DTube there will be information about the sound or video stored in the post `json_metadata` field. It seem that if someone post on Dtube and edit the post on Busy, the metadata will be overwritten. This clear any information about the video in the post and make the video not accessible anywhere. It happen on Busy1, we should try to reproduce it on Busy 2 if it's also the case.

Sekhmet:
Fixes #914 

Changes:
- If draft has metadata already it will be merged into new metadata. Topics, images, users, and links will be updated anyway to match current content.

### DIFF
--- a/src/post/Write/Write.js
+++ b/src/post/Write/Write.js
@@ -187,11 +187,20 @@ class Write extends React.Component {
 
     if (this.state.isUpdating) data.isUpdating = this.state.isUpdating;
 
-    const metaData = {
+    let metaData = {
       community: 'busy',
       app: `busy/${version}`,
       format: 'markdown',
     };
+
+    // Merging jsonMetadata makes sure that users don't lose any metadata when they edit post using
+    // Busy (like video data from DTube)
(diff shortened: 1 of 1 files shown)


src/post/Write/Write.js
```diff
@@ -187,12 +187,21 @@ class Write extends React.Component {
...
-    const metaData = {
+    let metaData = {
       community: 'busy',
       app: `busy/${version}`,
       format: 'markdown',
     };
 
+    // Merging jsonMetadata makes sure that users don't lose any metadata when they edit post using
+    // Busy (like video data from DTube)
+    if (this.props.draftPosts[this.draftId] && this.props.draftPosts[this.draftId].jsonMetadata) {
```
jm90m: I think you could use `_.has` (https://lodash.com/docs/4.17.4#has) for this so:
```
_.has(this.props.draftPosts[this.draftId], 'jsonMetadata')
```

PR #1008, opened 2017-11-04
Add voting slider to comments

### LINKED ISSUE #998: Voting power slider for comments
We currently have voting slider for posts, but we should add voting slider for comments as well.

Sekhmet:
Fixes #998.

Changes:
- Refactor `Buttons` and `CommentFooter` out of `Comment` component
- Add voting slider to comments

comment-like


bonustrack:
Great work! Working well. In the worst case scenario (a subcomment in small screen) the voting slider will be little bit small:
image
Do you have some idea to make better experience on small screen @Sekhmet ? Maybe a modal.

Sekhmet:
I have updated Voting Slider styles so it looks nicer on mobile.
375px width:
image

src/components/CommentFooter/Buttons.js
```diff
@@ -0,0 +1,263 @@
...
+    const upVotes = getUpvotes(comment.active_votes).sort(sortVotes);
+    const downVotes = getDownvotes(comment.active_votes)
+      .sort(sortVotes)
+      .reverse();
+
+    const totalPayout =
+      parseFloat(comment.pending_payout_value) +
+      parseFloat(comment.total_payout_value) +
+      parseFloat(comment.curator_payout_value);
+    const voteRshares = comment.active_votes.reduce((a, b) => a + parseFloat(b.rshares), 0);
```
jm90m: shouldnt this be `voterShares` ?

src/components/CommentFooter/CommentFooter.less
```diff
@@ -0,0 +1,58 @@
...
+      color: @purple;
+      & > .iconfont {
+        color: @purple;
+      }
+    }
+  }
+
+  &__count {
+    margin-right: 8px;
+    color: #99aab5;
```
jm90m: missed color constant here

src/components/CommentFooter/Buttons.js
Sekhmet: They are actually called `rshares` so I think `voteRshares` is okay-ish name

PR #1020, opened 2017-11-05
Handle deleted posts

### LINKED ISSUE #906: DTube video shows in post after the post was "deleted"
#### Expected behavior

I tested DTube. After video upload it created a post on steem blockchain. I edited the post in a way that the title and body contain only **deleted**. The video should not display after this edit.

#### Actual behavior

I edited the post, removed everything that was there before and typed **deleted** only. The video still shows in the post even though the body doesn't contain any link.

image

image

#### How to reproduce

1. Upload a video on DTube
2. Open the post that is created in busy
3. Mark it as deleted (title and body contain only **deleted**)
4. The video still shows

* Browser: Chrome
* Post: `/dtube/@espoem/7ovj6noc`

Sekhmet:
Fixes #906.

Changes:
- Add `isPostDeleted` helper that detects if post is deleted (`body` and `title` equal to `deleted`).
- Hide deleted posts in feed.
- Display `StoryDeleted` component instead of content if post has been disabled in single post view.

image

### DIFF
--- a/src/components/Story/StoryDeleted.less
+++ b/src/components/Story/StoryDeleted.less
@@ -1,8 +1,10 @@
+@import (reference) "../../styles/custom.less";
+
 .StoryDeleted {
   text-align: center;
   margin: 32px 0;
-  
+
   h3 {
-    color: #99aab5;
+    color: @blue-botticelli;
   }
 }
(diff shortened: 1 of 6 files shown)


src/components/Story/StoryDeleted.less
```diff
@@ -0,0 +1,8 @@
+.StoryDeleted {
+  text-align: center;
+  margin: 32px 0;
+  
+  h3 {
+    color: #99aab5;
```
jm90m: should probably use constants for this :P

Issue #826, opened 2017-10-17
Add space before an image


espoem:
<img src="https://user-images.githubusercontent.com/16621517/31648828-238ac6ca-b310-11e7-94dc-2fc0a6057e19.png" height="240px"/>

#### Expected behavior

There should be a bigger vertical space between the text and the image.

#### Actual behavior

Sometimes the image is placed right after the text that comes before the image whithout any space.

#### How to reproduce

1. Go to `/busy/@busy.org/busy-new-design-and-multilanguage-call-for-translators`
2. Scroll down and see images

* Browser: Chrome
* Operating system: Windows 10

bonustrack:
This issue has been solved

PR #1021, opened 2017-11-05
Fixes img and p spacing issues

### LINKED ISSUE #892: There may be double vertical space between an image and a paragraph
#### Expected behavior

Ideally, the space between an image and a preceding text should be consistent.

#### Actual behavior

There is double vertical space if the text and the image are each in their own paragraphs.

#### How to reproduce

1. Go to `/travel/@rosatravels/travel-gem-27-take-you-back-in-time-water-taxi-across-dubai-creek`
2. Compare with `/busy/@busy.org/busy-new-design-and-multilanguage-call-for-translators`

* Browser: Chrome

This is related to #826 and #882  

This post `/children/@breezin/what-s-a-mother-to-do` can be a test case. There are a lot of images in different locations.

sirrius-a-b:
Fixes #892, #826 .

Changes:
- split the margins for `.Body img` and `.Body--full p` to avoid double margins and crowded content

Because this is present in 2 issues, I will give an explanation for these changes:

How I see it, there are 3 cases:
- A: `<p>Some text </p> <p>Other text</p>`
- B: `<p>Some text  extra text</p> <p>Other text</p>`
- C: `<p>Some text</p>  <p>Other text</p>`

The only way I can think of right now for these three cases to have the same space between the texts and the image is for both `p` and `img` to have vertical margins set to `0`. We don't want that because it does not satisfy #826.

The next best thing is to split the vertical margins so that both `p` and `img` has equal top and bottom margins. This will avoid double margins for C(#892) and will partially satisfy A and B(#826) because the texts and the images will not get crowded.

Lastly, I removed `:not(:last-child)` from `p` because since the vertical margins are split (not just bottom margin), it would have had the last `p` closer to the penultimate element.

Suggestions are welcome.


sirrius-a-b:
Also, the case for removing `:not(:last-child)` and not use `:last-child` is that in cases where `p` is not the last element or if there are `p` elements inside lists, then `:last-child` will still be met and will only have top margin.

A good test case is: `/busy/@busy.org/busy-new-design-and-multilanguage-call-for-translators`, because the post ends in a list and each list element has a `p` inside `li` which will trigger `:last-child` for each `p` inside that list.

src/components/Story/Body.less
```diff
@@ -200,8 +201,8 @@
       font-size: 21px;
     }
 
-    p:not(:last-child) {
-      margin: 0 0 1.5rem 0;
+    p {
+      margin: 0.75rem 0 0.75rem 0;
```
jm90m: isn't this the same as `margin: .75rem 0;` ?

sirrius-a-b:
I cleaned-up the `p` tag's margin.

bonustrack:
I don't think we should change the margin of `p`, now the space between 2 paragraph look too small, not sure what you trying to solve with a smaller space.

sirrius-a-b:
You are right about the smaller space. I fixed it now.

With this PR I wanted to fix or at least satisfy #892 and #826 . 

To better see what I am talking about, see `/mushrooms/@sirrius/fun-facts-about-mushrooms-featuring-original-images`

I that post the letters **A**,**B** and **C** correspond to the 3 cases I mentioned above.

On the live site, **A** has equal space above and below the image, **B** has no space below the image and **C** has double-space on top of the image( #892 ).

I tried to solve the double spacing and no-spacing issues. But we might not want to fix these issues and just let the users craft their posts as they want, because it is hard to make the layout always look the same when the HTML structure is different.

Ultimately, if you think this is not worth pursuing, we could ignore this PR.

PR #1042, opened 2017-11-07
Remove Google Analytics and add Segment

bonustrack:
Google Analytic integration was not working since we updated to react-router v4. This PR fixes this issue and also switch from Analytics code to Segment code which give more flexibility.

Fixes https://github.com/busyorg/busy/projects/6#card-5480527

Changes:
- Remove Google Analytics code
- Add Segment code
- Link Google Analytics to Segment

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -24,7 +24,9 @@ import RequireLogin from './auth/RequireLogin';
 import Wallet from './wallet/Wallet';
 
 const logPageView = () => {
-  window.analytics.page({ url: window.location.pathname });
+  if (window.analytics) {
+    window.analytics.page({ url: window.location.pathname });
+  }
   return null;
 };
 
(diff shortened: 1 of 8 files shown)


src/routes.js
```diff
@@ -23,83 +23,93 @@ import Drafts from './post/Write/Drafts';
...
-            <Wallet />
-          </RequireLogin>
-        )}
-      />
-      <Route path="/@:name" component={User} />
-      <Route path="/:category/@:author/:permlink" component={Post} />
-      <Route path="/" component={Page} />
-    </Switch>
+    <div>
+      <Route component={logPageView} />
```
jm90m: Wasn't too sure about this one, but it seems to go into the `logPageView` function on each page

bonustrack: @jm90m Yes it's required to track when user change page to get this data on analytics.

Issue #1041, opened 2017-11-07
Bug Hunting : Problems with user profile images


nhj7:
<center>![image.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1510050501/szgq7hu42y2vfhwpviuv.png)</center>

**Problems with user profile images**

<hr />
<br />

When you switch to posting for another account.
I have found a problem that my profile image does not change.
This may seem like nothing, but it can lead to misunderstandings.

**This is the profile image of @clayop**
![캡처_2017_11_06_14_31_16_784.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1510049890/sr0poqntnz8e1b7vkitw.png)

**But when I switch posts, I get a wrong profile image.**
![녹화_2017_11_06_14_30_49_649.gif](https://res.cloudinary.com/hpiynhbhq/image/upload/v1510049969/seb2zrqps1uwqj3hhqcp.gif)

I think it should be fixed as soon as possible.
Thank you for reading.
<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@nhj12311/problems-with-user-profile-images">Utopian.io</a></em><hr/>

jm90m:
Fixed in https://github.com/busyorg/busy/pull/1045

PR #1045, opened 2017-11-07
Add `componentWillReceiveProps` method to Avatar component 

jm90m:
Fixes # https://github.com/busyorg/busy/issues/1041.

Changes:
- update `Avatar` component imageUrl state property when username is passed to it

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -31,7 +31,7 @@ class Avatar extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.username) {
+    if (this.props.username !== nextProps.username) {
       this.setState({
         imageUrl: getImage(`@${nextProps.username}?s=${nextProps.size}`),
       });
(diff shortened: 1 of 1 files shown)


src/components/Avatar.js
```diff
@@ -30,6 +30,14 @@ class Avatar extends Component {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.username) {
```
Ne3l: Check `this.props.username !== nextProps.username` to avoid unnecesary setStates?

Ne3l: if `nextprops.username` is falsy, should the defaultImage been shown?

jm90m: I added the `this.props.username !== nextProps.username` check.
We shouldn't set the default image state here, even if `nextProps.username` is falsy. We want it to hit the `img.busy.org` and it will get the default avatar on there.. the `defaultImage` is only used when `img.busy.org` fails

Ne3l: Seems reasonable.

Then in `componentWillMount` is necessary the check? in terms of consistency seems strange that one checks it and other doesn't.

jm90m: nope, just removed it

PR #1047, opened 2017-11-08
Update common.less

### LINKED ISSUE #1009: Remove iframe borders on Vimeo embed
Vimeo embed has border around `iframe`. It should be removed.
https://busy.org/design/@design-guy/graffiti-house-branding

Kennybll:
Fixes #1009.
Changes:
- Removes border around iframes.
- Adds a 20px margin on the top and bottom to create a unanimous feeling with images.

### DIFF
--- a/src/styles/common.less
+++ b/src/styles/common.less
@@ -52,4 +52,6 @@ video {
 
 iframe {
   border: none;
+  margin-top: 20px;
+  margin-bottom: 20px;
 }
--- a/src/styles/common.less
+++ b/src/styles/common.less
@@ -49,3 +49,7 @@ video {
 .iconfont {
   font-size: 26px;
 }
+
+iframe {
+  border: none;
+}


src/styles/common.less
```diff
@@ -49,3 +49,9 @@ video {
 .iconfont {
   font-size: 26px;
 }
+
+iframe {
+  border: none;
+  margin-top: 20px;
```
jm90m: I'm not sure if we should add these margins in, the author of the post might not have intended there to be margins around their iframe.

jm90m: @bonustrack what do you think?

bonustrack: The margin create an extra space in feed:
image
This PR is good for solve the issue about iframe border, for the margin we can do it in another PR, we should use a specific style class for it.

jm90m: @Kennybll Did you want to solve the issue for the margins, what @bonustrack talked about above in this PR or in another PR? If you want to do it in another PR, please close this one :)

Issue #1040, opened 2017-11-07
Comments number strings in Recommended posts sidebar should be same as in posts


espoem:
#### Expected behavior

StoryFull and PostRecommendation should use the same strings for comments count. That is `comments_count`.

StoryFull uses 

```javascript
<FormattedMessage
  id="comments_count"
  values={{ count: intl.formatNumber(commentCount) }}
  defaultMessage="{count} comments"
/>
```

PostRecommendation should use:

```javascript
<FormattedMessage
  id="comments_count"
  values={{ count: intl.formatNumber(post.children) }}
  defaultMessage="{count} comments"
/>
```

#### Actual behavior

The strings are different so it shows different words and capitalization. This happens in other languages too.

```javascript
const commentsText = post.children === 1
        ? intl.formatMessage({ id: 'comment', defaultMessage: 'Comment' })
        : intl.formatMessage({ id: 'comments', defaultMessage: 'Comments' });

<Link
  to={`/${post.category}/@${post.author}/${post.permlink}#comments`}
  onClick={() => this.navigateToPostComments()}
  className="PostRecommendation__comment-link"
>
  {post.children}
  {' '}
  {commentsText}
</Link>}
```

![image](https://user-images.githubusercontent.com/16621517/32485822-f6637d0e-c3a4-11e7-879f-c1ca74cf4260.png)

#### How to reproduce

1. Open any post

* Browser: Chrome
* Operating system: Windows 10

jm90m:
Thanks again for reporting this issue @espoem.

Fixed in https://github.com/busyorg/busy/pull/1048

Please re-open this issue, if you still see that it exists.

PR #1048, opened 2017-11-08
Update PostRecommendation & Story Full to both use same comments_count strings

jm90m:
Fixes https://github.com/busyorg/busy/issues/1040 .

Changes:
- Update PostRecommendation & Story Full to both use same comments_count / comment_count translations


bonustrack:
When there is 0 comment it render in plural it should be singular:
image

jm90m:
@bonustrack no it should be plurarl :P 

https://english.stackexchange.com/questions/1931/there-are-no-comments-vs-there-is-no-comment

PR #1053, opened 2017-11-08
Add Transfer & Claim Buttons to own users profile page

### LINKED ISSUE #1051: "Transfer" and "claim rewards" buttons missing in the wallet in user's profile
In Busy we have different ways how to display the wallet information. The first one is in the main page in the left navigation bar and the second is simply in the profile. Yet they are not the same. If you go to the wallet in your profile you will see that there is no button for transfers and you can't calim the rewards either.

### Expected behaviour

Both wallets should have the same functions for the logged in user. If they go to https://busy.org/wallet or https://busy.org/@espoem/transfers (change the username with yours) they should have the same functions.

### Steps to reproduce

1. Have rewards to claim
2. Go to https://busy.org/wallet
3. Go to https://busy.org/@espoem/transfers (with your username)
4. There is no transfer button and no claim rewards button

- Browser: Chrome
- OS: Windows 10

<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@espoem/transfer-and-claim-rewards-buttons-missing-in-the-wallet-in-user-s-profile">Utopian.io</a></em><hr/>

jm90m:
Fixes #1051.

Changes:
- Create WalletSidebar component to render steem trending charts, transfer, and claim rewards button for current users own user profile page


bonustrack:
The transfer button appear in left (1) sidebar or right sidebar (2) depending :
image
We should have one or another but not both to make it less confusing, i would suggest to remove the transfer button from left sidebar for any user, and show transfer button on wallet page on every user. When it's your own profile the transfer button should open transfer modal with empty "to".

jm90m:
@bonustrack done, removed transfer button from user info and added to all user wallet pages

bonustrack:
When i load a profile, for a small lapse of time i will see the main content with the post loading placeholder aligned in left like this:
image
It should never appear like this not aligned, sidebar left should always have same width.

jm90m:
it looks like now that the button is gone the `Affix` component sets the initial height to 0, so I just added a minimum height to it

PR #1056, opened 2017-11-08
Add usernameStyles property to common User component

### LINKED ISSUE #1030: More Interesting with Word Wrap - Bug Hunt
I rarely open Busy.org in my Laptop at home since I'm much more adapted to Steemit.com and its Night Mode now. I do use Busy.org in the office since it's accessible there, I use v1 only, v2 is still blocked by our servers. I just opened v2 now at home and I noticed that people having long names could cause something to be misaligned. Just look here:

Our good friend Terry aka @surpassinggoogle has been surpassing a lot of things including text areas, pushing the limits and shouts no - he's not 'Following' the rules, cause he has his own, better rules - in a very good way. Shoutout!

Ideally, word wrapping should be applied. I guess it's okay to cut off names in the middle and just add ellipsis after it. It should look more like this:

It looks much better right? Let's study this bug further and help resolve it. I made these mock-ups to help us see better with larger images.

Our first image shows the misalignment caused by some people who has longer names than usual:

As you can see, it pushes some of the elements out of their proper places. I added some guidelines here so that you can see what I mean:

Let's try moving the "Following" buttons to their proper places here:

Then let's check what's the proper margin for the text in this area is:

Then let's fix it by applying word wrap to the text area, cutting off the name and adding ellipses after it. We can also just cut off the names and not add ellipses, but in my opinion it's better to add ellipses after to signal that the name isn't complete at all.

It's still not done though - we have some alignment issues with the icons used. I looked for the center of the icon and made sure it's aligned with the other elements (user avatars) in its column (I replaced the icon with another one since I can't find anything more similar for presentation purposes):

There you have it. Alignment and Margins have been fixed:

Disclaimer: Codes are needed and we do not know if it is doable. However if someone can code this then I'm sure @busy.org will make their site in its best shape!

<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@deveerei/more-interesting-with-word-wrap-bug-hunt">Utopian.io</a></em><hr/>

jm90m:
Fixes #1030 .

Changes:
- Add usernameStyles property to common User component
- Add ellipsis to long usernames for InterestingPeople blocks
- Add title for browser hover effect so you can see full username

screen shot 2017-11-08 at 5 36 44 pm


src/components/Sidebar/InterestingPeople.js
```diff
@@ -19,7 +19,19 @@ const InterestingPeople = ({ users, onRefresh }) => (
...
       </h4>
       <div className="InterestingPeople__divider" />
-      {users && users.map(user => <User key={user.name} user={user} />)}
+      {users &&
+        users.map(user => (
+          <User
+            key={user.name}
+            user={user}
+            usernameStyles={{
+              width: '120px',
```
bonustrack: I'm not sure if we should have width hardcoded and if it's the best way to create usernameStyles props. Can't we fix the style inside <User> component directly?

bonustrack:
When i click follow button, the button is going out of the frame, it would be nice to fix this too:
image

src/components/Sidebar/InterestingPeople.js
jm90m: So User component is used in different places, the only way to add an ellipses via css is having a width on it for different cases.

Sekhmet:
This fix doesn't seem to work for me:
image
You probably can reproduce this problem by switching language to Polish or change button text by hand.

Sekhmet:
Is it possible to add a little bit of margin between username and Button?
image
image

jm90m:
@Sekhmet yeah just updated right now using flex, and made dynamic ellipses.. damn to be honest i spent longer on this than I would like to admit :rofl: , sometimes I hate writing styles hahaha, let me know what you think :)

src/components/Sidebar/User.less
```diff
@@ -4,22 +4,41 @@
 
 .User__top {
   display: flex;
+  flex-wrap: nowrap;
   justify-content: space-between;
   align-items: center;
   margin-bottom: 8px;
+  width: 100%;
+  white-space: nowrap;
+
```
Sekhmet: Looks like this line is not necessary.

jm90m: sure will remove this line.

PR #1066, opened 2017-11-11
Parent Node Bug and Margin On Iframes

Kennybll:
Fixes # (I did not create an issue. I found it myself and though to fix it right away).

Changes:
- As seen in https://i.imgur.com/Y9irhyK.png when someone puts an iframe into the editor without a parent element (such as a div) there is an error. This is because there is no parentNode when it is setup this way. So lines 146-151 make sure if there is no parentNode, then it will create a wrapping div. The problem lies at 158. Without the parentNode you can't replace the child. It took me five hours to find this out.
- Line 139 allows for all iframes to have an embedded thumbnail.
- Adding line 55 created a 20 pixel margin on top of an iframe. This makes it more in sync with images. Images have a 20px margin-top and it is added directly to the element style. So adding margin-top to the iframe element makes sense as well.

### DIFF
--- a/src/styles/common.less
+++ b/src/styles/common.less
@@ -52,5 +52,4 @@ video {
 
 iframe {
   border: none;
-  margin-top: 20px;
 }


Kennybll:
Umm circle ci failed because it could untar/unzip something...

Kennybll:
@bonustrack

src/styles/common.less
```diff
@@ -51,5 +51,6 @@ video {
 }
 
 iframe {
-  border: none; 
+  border: none;
+  margin-top: 20px;
```
bonustrack: Iframe tag should not have margin by default, iframe are not only visible in a single post page, they are also visible on the feed, this change create an extra margin on the feed.

bonustrack:
@Kennybll updated CircleCi

Kennybll:
@bonustrack I undid the commit that added margins to the iframes automatically.  Though the Parent Node Bug still needs attention.

bonustrack:
Thank you @Kennybll. I'm testing the heroku deployment here: https://busy-nd-pr-1066.herokuapp.com/ and i still can see the same error: 
image

bonustrack:
I'm closing this PR due to inactivity, you can reopen it if there is some changes.

Issue #1061, opened 2017-11-09
Menu and transfer modal overlap on mobile


espoem:
A user can transfer funds to another user if they go to the other user's profile. There is a menu that contains an option for transferring. If you click the transfer button, a modal shows up but the menu doesn't disappear and blocks the modal Cancel and Continue buttons. The menu can be dismissed if you click somewhere in the space.

![image.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1510242710/edfz3ogwm9j6n27qskfa.png)

### Expected behaviour

The menu should disappear when the menu item is clicked so the modal is not blocked.

![image.png](https://res.cloudinary.com/hpiynhbhq/image/upload/v1510242768/o41s7xuhcgasiq5a6e7f.png)

### Steps to reproduce

1. Go to another user's profile
2. Click the three dots and then the transfer button

- Browser: Brave
- OS: Android 7.1.2

/n/n<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@espoem/menu-and-transfer-modal-overlap-on-mobile">Utopian.io</a></em><hr/>

robaguilera:
I'd like to volunteer to tackle this issue.  I can work on this weekend and get a PR sometime within the next two days.

eastmaels:
Cause is also the same with #782.

PR #1079, opened 2017-11-15
Bugfix/transfer popup does not close on select

eastmaels:
Fixes # 782.

Changes:
- `src/user/User.js` 
    - added `popoverVisible` state; 
    - added `handleVisibleChange` function; 
    - in `handleUserMenuSelect`, set `popoverVisible` to false;
- `src/user/UserHero.js` 
    - added `isPopoverVisible` and `handleVisibleChange` properties
- `src/components/UserHeader.js` 
    - passed `isPopoverVisible` and `handleVisibleChange` props to as parameters to ant `Popover` component


eastmaels:
Also fixes #1061.

src/components/UserHeader.js
```diff
@@ -103,6 +107,8 @@ UserHeader.propTypes = {
   isSameUser: PropTypes.bool,
   hasCover: PropTypes.bool,
   onSelect: PropTypes.func,
+  isPopoverVisible: PropTypes.bool,
```
Sekhmet: Can we move this prop above functions? Same for defaultProps.

src/user/UserHero.js
```diff
@@ -65,12 +76,16 @@ UserHero.propTypes = {
   isSameUser: PropTypes.bool,
   hasCover: PropTypes.bool,
   onSelect: PropTypes.func,
+  isPopoverVisible: PropTypes.bool,
```
Sekhmet: Same thing there.

src/components/UserHeader.js
eastmaels: Copy. will move.

src/user/UserHero.js
eastmaels: Copy. will move.

Issue #839, opened 2017-10-18
Mobile responsive: Add search button


espoem:
<img src="https://user-images.githubusercontent.com/16621517/31723862-4c478ee8-b420-11e7-9f94-cbc1a690fe32.png" height="280px"/>

#### Expected behavior

There should be a search button on mobile layout. The input prompt should look similar to the image above.

#### Actual behavior

There is no search option.

#### How to reproduce

1. Open busy on mobile

* Browser: Brave
* Operating system: Android 7.1.2

PR #1085, opened 2017-11-16
Add search button to mobile top navbar

dfelczak:
Fixes: #839 

When search box is not active: 
screen shot 2017-11-17 at 00 15 34

and when search box is active: 
screen shot 2017-11-17 at 00 15 53


Sekhmet:
Hey, it looks great!
I noticed that toggling between states takes a while, that's because it causes entire Wrapper to re-render:
image
Can you move this logic and state to Topnav itself, so it won't update entire Wrapper but only Topnav?

dfelczak:
Nice catch! I will move state to Topnav :)

dfelczak:
Hello @bonustrack, thanks for good feedback. I've just applied changes and PR is ready for next review.

src/components/Navigation/Topnav.js
```diff
@@ -1,20 +1,58 @@
-import React from 'react';
+import React, { Component } from 'react';
```
Sekhmet: We don't import `Component` directly and use `React.Component` instead.

PR #1089, opened 2017-11-17
Server-Side Rendering

### LINKED ISSUE #877: Google Plus doesn't fetch post previews
#### Expected behavior

When I share Busy links on Google+, I want to see previews, like on Facebook (it works well there).
This issue is about the old Busy, for the new Busy see #823. 

#### Actual behavior

The image preview are not fetching from the post when I paste a link.

#### How to reproduce

Share a Busy post on Google+.

* Browser: Chromium v61.0.3163.100
* Operating system: Ubuntu Linux 14.04

### LINKED ISSUE #1023: Server-Side Rendering Checklist
Implementing Server-Side Rendering (SSR) that is fail-proof and reliable will require taking care of a few things first. Crashing app running in the user's browser is bad, but crashing production server is even worse, thus we have to make sure everything works as smoothly as possible.

### Improving overall performance.
I've started profiling Busy and noticed few performance issues that I will investigate further. Most of them are related to performing operations that are irrelevant in most cases (e.g. sorting list of votes for every post in feed even though most likely this won't be useful), but there are problems with code performance as well (in sorting votes as well).
Improving performance will improve user experience (especially on mobile or older devices), but will make SSR faster and use fewer resources.

### Find and fix memory leaks
Memory leaks are even worse on the server than browsers as usually clients have more RAM and have to do less work than the server that will have to handle hundreds of users. Exhausting RAM may lead to server crashes, so we have to make sure we find any memory leaks that might happen and solve them.

### Update Busy to React 16
React 16 is 2-3 times faster on the server than React 15. React Fiber focuses on SSR and adds new, useful features for that (e.g. hydration and streaming). We should migrate to React 16 before enabling SSR.
https://reactjs.org/blog/2017/09/26/react-v16.0.html#better-server-side-rendering

### Isomorphic code
We don't want to have any libraries failing on sever because they were designed to work only in the browser. We have to make sure that our code and libraries we use are isomorphic.

### Architectural changes and code splitting
Currently, we have many components that are only used in one place only (e.g. Story that is only used for displaying single post). We should move such components to feature directories. This will make code splitting more effective making features' chunks a little bigger, but common components chunk would be significantly smaller. We should split our bundle into multiple chunks (features like feed and post, vendor, and common) and load only required files.

After we got all of this we can start working on actual SSR. With those features, SSR should be working efficiently without crashes and bugs.

## SSR Checklist:
- [x] Improve overall performance
- [x] Find and fix memory leaks
- [x] Update Busy to React 16
- [x] Isomorphic code
- [x] Architectural changes

#### Update 20/11/2017
Removed code splitting from the checklist, it will be implemented later.

This post is just a draft and it is subject to discussion and changes.

Sekhmet:
Fixes #1023 
Fixes #877

Changes:
- [x] Separate `client` and `server` in `src` folder.
- [x] Add `common` folder.
- [x] Rewrite Express.js code
- [x] Add `fetchData` static methods. 
- [x] Remove legacy `needs` code.
- [x] Handle crashes.
- [x] Isomorphic analytics.
- [x] Fix UserMenu sidebars.
- [x] Authenticate and load user language before rendering page (prevent flickering of non-translated content).
- [x] Use Busy logo as placeholder


Sekhmet:
I haven't finished everything I wanted to do with SSR just yet, but everything so far works. I would like to merge those changes as quickly as possible, so we don't have two conflicting branches causing problems while merging.
I will continue working on new features on another PR, so we can have SSR on production already and make development easier.

jm90m:
Not too familiar with server side rendering stuff, but I've been testing performances for large data sets that get rendered, like in the Activity page. And there seems to be a performance increase, well done: 

Compare filtering here: https://busy-nd-pr-1089.herokuapp.com/@timsaid/activity

to filtering on here: https://busy.org/@timsaid/activity

bonustrack:
I have a small loading icon and these scrolls when i open a profile with french locale:
image

https://busy-nd-pr-1089.herokuapp.com/@busy.org
Can you reproduce these issues?

Sekhmet:
I have seen this issue (scrollbars) and it was on my todo list, I will fix this shortly.
I will have to take a look at an issue with small loading icon though.

PR #1114, opened 2017-11-25
Add invitation system

bonustrack:
Add invitation system with incentive of 20% beneficiaries for the referrer.

Changes:
- [x] Create invitation route `/i/@username`
- [x] Add cookie on user browser with username as referral
- [x] Add 20% beneficiaries to referral if user publish a root post
- [x] Add beneficiaries to post only if user account has been created after the referral invitation 
- [x] Make possible to invitation link to redirect to an user post like `/i/test/@hellosteem/hello-world`
- [x] Create a page invite for user to easy copy paste referral link
- [x] Add button for copy


jm90m:
screen shot 2017-11-26 at 9 27 35 pm

Invite looks kinda weird here without an icon and aligned with all others?

jm90m:
probably should add a clipboard icon that is clickable on the right of the input sidebar, you can make it copyable like here: https://codepen.io/shaikmaqsood/pen/XmydxJ/

jm90m:
Is this link `https://busy-master-pr-1114.herokuapp.com/i/@jm90mm` supposed to work? right now it redirects to `/@jm90mm`

bonustrack:
@jm90m Strange there should be an icon:
image this is a default icon, i didn't change the iconfont url

bonustrack:
The link is supposed to send you to the homepage, i will change that

src/client/invite/Invite.js
```diff
@@ -0,0 +1,67 @@
...
+    authenticatedUserName: PropTypes.string,
+  };
+
+  static defaultProps = {
+    authenticatedUserName: '',
+  };
+
+  render() {
+    const { authenticatedUserName } = this.props;
+    const port = window.location.port ? `:${window.location.port}` : '';
```
Sekhmet: Can we use `location` from react-router? `window` is undefined on the server.

bonustrack: react router `location` give me only `pathname`, i can't get `port`, `host`, or `protocol` from it. But i can check for window availability before generating the link.

src/client/invite/Invite.js
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  static defaultProps = {
+    authenticatedUserName: '',
+  };
+
+  render() {
+    const { authenticatedUserName } = this.props;
+    let inviteURL;
+    if (typeof window !== 'undefined') {
```
Sekhmet: If this page gets rendered on server `inviteURL` will be null and will stay this way (so there will be no invite URL at all).
`compontentDidMount` is not called on the server, only after mounting in a web browser. We could use this to create URL because `window` should be initialised there. We will save it in state (eslint is not going to like this, but we can create new method `createInviteURL` so eslint won't warn about using setState in componentDidMount.

bonustrack:
@jm90m can you see the icon in menu?

image

jm90m:
screen shot 2017-11-30 at 8 26 05 pm
Just tried it and I still can't see, can @Sekhmet see it?

Sekhmet:
Yeah, I can.
image

@jm90m did out style bundle not update for you or just iconfont file?

src/client/invite/Invite.js
```diff
@@ -0,0 +1,95 @@
...
+              <div className="">
+                <Form.Item>
+                  <Input
+                    className="Invite__input"
+                    value={this.state.inviteURL}
+                    readOnly
+                  />
+                  <CopyToClipboard
+                    text={this.state.inviteURL}
+                    onCopy={() => this.setState({ copied: true })}
```
Sekhmet: We should use handler for this action for performance reasons.

src/client/invite/Invite.js
```diff
@@ -0,0 +1,95 @@
...
+    authenticatedUserName: PropTypes.string,
+  };
+
+  static defaultProps = {
+    authenticatedUserName: '',
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
```
Sekhmet: We usually initialize state like this:
https://github.com/busyorg/busy/blob/master/src/client/components/AvatarLightbox.js#L20-L22

PR #1115, opened 2017-11-26
More accurate error messages for Steem/SBD transfers

Smittyvb:
Changes:
- If username being attempted to have money transfer to, has a username of less than 3 characters, say that the username is too short.
- Same thing for >16 characters.


Smittyvb:
Looks like tests failed because of:
```
npm ERR! Invalid tar header. Maybe the tar is corrupted or it needs to be gunzipped?
```
In the package installation stage. Not sure how this could cause this, though. Seems to be a problem with npm.

jm90m:
can we get a deploy on this @bonustrack ?

src/client/wallet/Transfer.js
```diff
@@ -119,6 +119,36 @@ export default class Transfer extends React.Component {
       return;
     }
 
+    if (value.length < 3) {
```
jm90m: probably could make both of these if statements into one seperate function above like `validateBalance`, but this one could be `validateUsernameLength`

jm90m: Also instead of having random numbers here, maybe instantiate them as variables like `minUsernameLength` & `maxUsernameLength`

jm90m:
This is missing `username_too_long` and `username_too_short` in en.json file, please add these 2 translations to `en.json`

PR #1111, opened 2017-11-24
Fixes post reading time large focus area issue

sirrius-a-b:
Fixes #1107 .

Changes:
- moved the `CommentFooter__bullet` outside of the `Tooltip`
- made small CSS changes to keep the existing style given the above change

PR #1122, opened 2017-11-28
Posted From Feature

Kennybll:
Show the app used to publish a post in the UI.

Changes:
- Adds a formatted message telling what app the post was posted by. Ex: Posted by busy
- There is a tooltip around that, which says the version of that app.


Kennybll:
From Workflow https://github.com/busyorg/busy/projects/6

Sekhmet:
`JSON.parse` can fail, and when it does it throws error making React go wild.
Can you make it fail-safe?
https://gist.github.com/gucheen/12b90451f04733078d7f

bonustrack:
@Kennybll here is my review:

- The tag should appear next to the reputation indicator, now there is this long space due to `Posted from steemit` text width see:
image
- The text `Posted from steemit` is too long et repetitive, i prefer just `Steemit` instead.
- This should also appear in single post
- We should save a list of apps in json file with name of app with correct case for example:
```
{
  steemit: 'Steemit',
  busy: 'Busy',
  esteem: 'eSteem',
  chainbb: 'chainBB',
}
```

Kennybll:
Also fixes issue #1111.

bonustrack:
@Kennybll in future please use more relevant name for your commits

Kennybll:
Okay.

Sekhmet:
@Kennybll 
1. Make sure that your commits do one thing and only one.
2. Don't integrate changes outside of base branch (`master`) in this case if it's not relevant to your PR.

src/client/helpers/apps.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  steemit: 'Steemit',
+  busy: 'Busy',
+  esteem: 'eSteem',
+  chainbb: 'chainBB',
+  utopian: 'Utopian',
```
bonustrack: Can you add also DTube, Steepshot, Zappl, Steemkr, dMania and DSound aswell?

bonustrack:
App name should appear in the blue rectangle area
image

Kennybll:
The tag won't stop moving... 
image

bonustrack:
Nothing is impossible

Kennybll:
Your right @bonustrack

Kennybll:
What do you think @bonustrack @Sekhmet

src/client/components/Story/Story.js
```diff
@@ -234,29 +273,34 @@ import './Story.less';
...
+                      <FormattedDate value={`${post.created}Z`} />{' '}
+                      <FormattedTime value={`${post.created}Z`} />
+                    </span>
+                  }
+                >
+                  <span className="Story__date">
+                    <FormattedRelative value={`${post.created}Z`} />
+                  </span>
+                </Tooltip>
+                { postedFrom }
```
Sekhmet: We tend not to use whitespaces inside curly braces.

src/client/components/Story/StoryFull.js
```diff
@@ -263,6 +264,43 @@ class StoryFull extends React.Component {
       );
     }
 
+    let postedFrom;
```
Sekhmet: This logic is exactly the same in `Story` as well. We can create a component to do just that, so we don't have duplicate code.

Kennybll:
Again sorry for s many commit, they all are different files that are changed because Github.com doesn't allow multiple file commits. Git on my Windows laptop isn't working so well.

Kennybll:
There @Sekhmet . Is there any other things you guys tend to not do, so I can be prepared in the future?
I remember you once said you put styles at the bottom of the imports.

src/client/components/Story/PostedFromEmbed.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import { Tooltip } from 'antd';
+import apps from '../../helpers/apps';
+import { jsonParse } from '../../helpers/formatter';
+import './PostedFromEmbed.less';
+
+const PostedFromEmbed = ({ post }) => (
+  <span>
```
Sekhmet: I think it would make sense to add a class name to a topmost element in this component (to respect http://getbem.com/introduction guidelines).

src/client/components/Story/Story.less
```diff
@@ -57,6 +57,10 @@
     padding: 16px;
     display: flex;
 
+    &__flex {
```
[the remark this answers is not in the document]
Kennybll: Yes. On line 238. I had to make that span display flex because the Topic tag would be on the next line next to the date posted.

Kennybll:
@Sekhmet , so I figured out what was wrong with git on my computer. Nothing! Lol. I thought the process was hanging, but nope, it was running the pre-commit tests. It just took a minute to output something.

Kennybll:
Is this good @Sekhmet @bonustrack ?

src/client/components/Story/StoryFull.less
```diff
@@ -54,7 +58,7 @@
     position: relative;
     display: flex;
     padding: 16px 0;
-
+
```
Sekhmet: There is additional whitespace there.

src/client/helpers/apps.js
```diff
@@ -0,0 +1,13 @@
...
+  dtube: 'dTube',
```
bonustrack: It's DTube, DSound and dMania (not dTube and dSound). Can you fix the case?

PR #1125, opened 2017-11-28
Add HeroBanner

### LINKED ISSUE #999: Add hero for not logged in user
If user isn't logged in we should have hero that will show up for him. I think user should be able to dismiss hero by clicking x icon, and it won't show up for them anymore (should be saved to local storage).

Sekhmet:
Fixes #999.

image

Changes:
- [x] Add HeroBanner component.
- [x] Support Affixes with dynamic content before sidebars. 
- [x] Enable HeroBanner for unlogged users only.
- [x] Allow closing HeroBanner.
- [x] Style buttons.


bonustrack:
Wasn't supposed to be full-width, something like this:
image

I think we dont need separator between the buttons and we need bigger text and biggers buttons.

jm90m:
idk if this is related to this PR, but on for some reason everytime I try to login and authenticate, I get hit with this:

screen shot 2017-11-29 at 7 32 54 pm

src/client/components/HeroBanner.less
```diff
@@ -0,0 +1,114 @@
...
+    display: flex;
+    height: 100%;
+    padding: 16px 0;
+
+    @media @small-max-width {
+      padding: 16px;
+    }
+
+    &__content {
+      margin: auto 0;
```
jm90m: I think you might be able to get rid of this if you are going to make the herobanner take up the entire screen width. And underneath `&__content` somewhere after around `line 87` you can add the following flex properties: 
```
align-items: center;
justify-contenter: center;
```
And then doing a few changes around like so:

in `HeroBanner.js`:
screen shot 2017-11-29 at 8 03 16 pm

in `Page.js`:
screen shot 2017-11-29 at 8 02 38 pm

in `HeroBanner.less`:
screen shot 2017-11-29 at 8 02 53 pm

Here's what I have in my local, after checking out your fork and branch, and testing it out.

screen shot 2017-11-29 at 7 56 55 pm

sry for the screenshots, but the git GUI that I use doesn't let me copy and paste lol.....

src/client/components/Utils/Affix.js
```diff
@@ -30,12 +30,22 @@ class Affix extends React.Component {
...
+      window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop;
 
     this.top = 0;
     this.bindedBottom = false;
     this.bindedTop = true;
 
+    this.baseTopPosition = this.relativeContainer.getBoundingClientRect().top + this.lastScroll;
+
+    if (this.relativeContainer.parentElement) {
+      this.mo = new MutationObserver(() => {
```
jm90m: I have never used `MutationObserver` before but I trust that this works :rofl:

src/client/feed/HeroBannerContainer.js
```diff
@@ -0,0 +1,13 @@
+import { connect } from 'react-redux';
+import HeroBanner from '../components/HeroBanner';
+import { getIsAuthenticated, getIsLoaded, getIsBannerClosed } from '../reducers';
+import { closeBanner } from '../app/appActions';
+
+export default connect(
```
jm90m: I really like this idea, having a suffix for a component that is `Container`, which handles all the state properties 👍

src/client/styles/modules/feed-layout.less
```diff
@@ -24,6 +24,7 @@
     float: left;
     display: none;
     width: @left-width;
+    min-height: 1px;
```
jm90m: won't need this anymore if banner is full width of the screen

jm90m:
had some small comments, but once @bonustrack approves it, its fine with me

Sekhmet:
Our error middleware even though it handles errors doesn't catch them so it always throws an error in the console. I updated login action so it rejects instead of resolving when there is no access token, so this is why it started showing this message.
Handling errors is on my TODO list so I would solve it (once and for all) there.

bonustrack:
Can we have bigger buttons? And can we center the content on small screen?

bonustrack:
The margin in top should not be bigger than margin in down
image

PR #1131, opened 2017-11-30
Update Calculation for Users Est Account Value

### LINKED ISSUE #1121: Est. Account Value in Wallet is wrong
#### Expected behavior
Should do the calculation at the latest rate instead of using steemit data

see busy v1: https://github.com/busyorg/busy/blob/v1/src/user/UserTransfers.js#L68

jm90m:
Fixes #1121 .

Changes:
- Remove steemit user est account value calculations and add our own using busy what busy v1 had: https://github.com/busyorg/busy/blob/v1/src/user/UserTransfers.js#L68


bonustrack:
We should never see `NaN`, now it appear when loading the wallet:
image

jm90m:
@bonustrack should be updated now, lmk if you still see the NAN

PR #1133, opened 2017-11-30
Small screens: made top menu on 2 columns and added replies item

### LINKED ISSUE #991: Replies are not accessible on mobile device anymore
The replies have been moved from the drop menu at the right top to the left navigation on the main page.

Replies are not accessible on mobile after this change.

### After the change

* Browser: Brave
* OS: Android 7.1.2

sirrius-a-b:
Fixes #991 .

Changes:
- `PopoverMenu` component can accept custom classes, but the default class it is still `PopoverMenu` so that will not break anything because it is used in multiple places

- added the `Replies` top nav menu item, but it is hidden for large screens

- made the top nav menu to be split on 2 columns on small screens (screenshots below). The reason for this is to not have the menu cover too much of the screen and to have space in case other items might be added as @jm90m suggested on #991.

- fixed a 1 pixel off media breakout point (antd popover) for `.Topnav__user`.This is not related, but I found this when I tested my solution and thought I should fix it too.

Note:
I know that I didn't asked you guys if you wanted to fix this issue like this or if it's ok to have two columns or if the `PopoverMenu` should be allowed to accept custom classes. I am of course open to suggestions about this solution or alternatively this can also be ignored. Just let me know when you get a chance.

Screenshots:
800px:
busy-2-columns-800

480px:
busy-2-columns-480

360px:
busy-2-columns-360


sirrius-a-b:
I think the it did not pass the checks because the `PopoverMenu` is not expected to accept a custom CSS class.

However, because it is used in many places, it is useful to be able to set an additional class for custom functionality such as make just one menu on two columns on small screens and not the others.

Or, maybe you do not like the solution of two columns there?

jm90m:
Hey thanks for the contribution! Really appreciate the time and effort put into this. 

But Idk about this one... @pareshUD @bonustrack @Sekhmet, what do you think about this design? I personally don't like the 2 column menu style.

bonustrack:
I'm not fan too about the 2 columns.

sirrius-a-b:
Hey.. thanks for getting back to me. So, it's not a problem that I made `PopoverMenu` to accept a custom class? It's just that you do not particularly like the 2 columns solution?

I can make it like it was, just one column, if that is the only issue that you have. With the custom class it is really easy to hide/show a specific element (like `replies`), so it would still satisfy #991.

Sekhmet:
Tests fail because you are using double quotes there.
https://github.com/sirrius-a-b/busy/blob/78a51cfa20f99efbdf118e7a3e68c248818ded7f/src/client/components/PopoverMenu/PopoverMenu.js#L35-L37
We use apostrophes in JS files.

sirrius-a-b:
I changed the double quotes into single quotes as suggested, and made the menu have just one column.

Sekhmet:
You have updated `TopicSelector` so snapshot doesn't match.
Can you run `npm run test -- -u` to update them?
You can run tests locally before pushing by running `npm run test`. This way you will know if everything is okay.

src/client/components/Navigation/Topnav.less
```diff
@@ -86,6 +86,18 @@
     }
   }
 
+  &__popover {
+    & li {
+      display: block;
+      
+      @media screen and (min-width: 991px) {
```
jm90m: You should try to use the constants from `busy/src/client/styles/modules/variables.less`

sirrius-a-b: I have thought of that. Actually, that's how I did it the first time, locally.

But those do not match the `antd` responsive break-points, and it does not behave good, as we might expect.

For instance, the `replies` link will show up on the top (normal version) dropdown between 991px and 998px, which was not intended.

sirrius-a-b: That same declaration is already found in that file on another occasion. I agree that it should not be hard coded like that.

Do you have any suggestions about this? Maybe make another media var for `antd`? Or, maybe should we change the `@medium` var to match the `antd` break-point?

jm90m: I think it would be best to make another breakpoint then for antd, maybe `@antd-medium`

sirrius-a-b:
I have updated the `TopicSelector` snap and created `@antd-medium` media var as you guys suggested. It seems that now there are no failed tests/checks :)

src/client/components/Navigation/Topnav.less
```diff
@@ -86,6 +86,18 @@
     }
   }
 
+  &__popover {
+    & li {
+      display: block;
+
+      @media @antd-medium {
+        &:nth-child(2) {
```
jm90m: I don't think this is good practice to display: none on the nth-child, needs to be more explicit imo

sirrius-a-b: I agree @jm90m . But to do that I would have to give that particular list item a class. I just made a commit to do just that.

jm90m:
Hey thanks for this PR, really appreciate the time and effort that you put into this, but we have decided to go with another solution (https://github.com/busyorg/busy/pull/1165) that included displaying all fields that were missing for mobile users.

I will be closing this now, @Sekhmet @bonustrack.

Next time when working on a feature like this please checkin with @Sekhmet or @bonustrack approval and for designs please check with @pareshUD or @bonustrack, we appreciate the time and efforts and we don't want to waste anyones time.

Issue #903, opened 2017-10-26
[Proposal] Development workflow


Sekhmet:
Busy gets more and more external and internal contributions. We should make sure we are capable of scaling from tens to hundreds of contributions weekly. Because of that, I would like to propose some changes to our development workflow:
1. Making sure our repository stays clean.
Our repository is already at 4600 commits and it makes it would be hard to revert to some point in time if needed. We can't (shouldn't) update old history, but we can make an effort to keep our future git history better. To do this we should enforce few rules:
    1. Don't EVER push to `dev` or `new-design` without pull request. It doesn't matter if it's a just simple fix or anything else. No code should go to the repository without review (see below).
    2. Don't use `busyorg/busy` for your local work. Our repository should only have branches that are relevant to our release cycle. We might want to have `dev` and `new-design` branches and additional branches for bigger pieces of work that could have multiple contributors working on them. It would be hard for external contributor to work on wallet page alone, but if we would create a branch `wallet-page` and allow multiple contributors to create PRs to that branch with small chunks of code (like fetching data, styles, displaying data, adding tooltips, adding minor functionality) it would make it easier for them to contribute. After we agree that `wallet-page` is finished we can approve and merge PR from `wallet-page` to `new-design`. Any work that won't require contributions from more than yourself (so usually some small changes) should take place on a fork. If there is branch inside busy repository you should never push to it directly, use pull requests instead.
    3. Use merge --squash or rebase if merging more commits than desired. If it makes sense to squash commit into one, do it. If you feel that one commit is not enough, rebase your work into multiple commits that have clear description and purpose. If you have a clean commits with a proper description, you can merge it just fine (but I don't think we should ever have bigger merges than 2-3 commits and it should be reserved only for big changes). 
    4. Limit number of people that can merge pull requests. It's better to wait some time to let other developers get familiar with changes and share their thoughts on those changes than merging improperly or with low quality, buggy or dangerous code.
    5. Be strict about what goes into our repository. Introducing bad code is easier than turning bad code into good code. If we let bad code in, this will require additional work to get it rewritten properly. Don't do this: https://twitter.com/iamdevloper/status/397664295875805184. If you are not 100% sure about something ask others for help.
2. Improve the quality of issues and pull requests.
[...]

PR #1134, opened 2017-11-30
Adds Witnesses Page

Kennybll:
Adds Witness Page.

Changes:
- Adds Witnesses Page in Routes
- Adds new functions to apiHelpers and sortHelpers
- Adds witnessesReducer to reducers


bonustrack:
@Kennybll You should ask a core member of Busy before working on feature, specially when we don't explicitly ask for this feature, there should be an issue or card in GitHub Project for it if it's something we approved. Your design proposal is very different from what we have on others pages on the sites. We can't approve your PR like that, we need a mockup for this otherwise you will spend lot of time in this PR.

Kennybll:
It is under the Github workflow, as Add an option to vote for a witness. What do you mean by mockup?

bonustrack:
I mean we should agree on a design with @pareshUD before doing the integration to avoid too much changes.

Kennybll:
So should I close this pull request, or should I contact pareshUD?

Sekhmet:
You shouldn't start writing code if that feature hasn't been discussed yet. First create an issue, let others give their opinions on the topic, make sure we have everything figured out (features, look and feel, the way it should be implemented) and then announce that you are willing to work on it, this way:
1. We don't waste time implementing things that will have to be changed (in some cases rewritten from ground up).
2. No two developers would work on the same feature making one's developer work useless.

You can take a look at those guidelines, they are not accepted yet, but I think it's good starting point and noone would get hurt by respecting those:
https://github.com/busyorg/busy/issues/903

jm90m:
Thanks for the contribution for this PR, but I'll be closing this PR since @Sekhmet or @bonustrack  has not approved of this. We can always reference this PR back when it is time to create a `Witness Page` with designs from @pareshUD.

PR #1141, opened 2017-12-01
Save Reward Settings In Metadata

Kennybll:
Fixes Save reward settings (50%/100% and upvote..) in metadata.

Changes:
 - Saves previous post's reward settings under SteemConnect.
 - Then fetched it when creating a new post/draft.


src/client/post/Write/editorActions.js
```diff
@@ -165,6 +166,8 @@ export function createPost(postData) {
 
     const newBody = isUpdating ? getBodyPatchIfSmaller(postData.originalBody, body) : body;
 
+    saveSettings({ upvoteSetting: upvote, rewardSetting: reward });
```
Sekhmet: This action isn't dispatched so it won't save anything, am I right?

jm90m: yup you're right... it needs to be `dispatch(saveSettings({ upvoteSetting: upvote, rewardSetting: reward })`

src/client/settings/settingsReducer.js
```diff
@@ -7,6 +7,8 @@ const initialState = {
   votePercent: 10000,
   showNSFWPosts: false,
   loading: false,
+  upvoteSetting: true,
+  rewardSetting: '50',
```
jm90m: I see this number '50' a lot throughout this PR, I feel like it should be a constant somewhere, like `export const INITIAL_REWARD_SETTING = '50'` then you can use that constant throughout where ever files need to be done

jm90m:
looking better and better! good work, one thing though instead of having the file named `constants` could you make it more specific and name it to `rewards.js` and move it over to this directory:

`/busy/src/common/constants/`

jm90m:
looks like there are conflicts on here with `src/client/post/Write/editorActions.js` please fix

Sekhmet:
I have just created new post (50% split, no upvote). When I go back to editor it looks like it has been saved properly, but if I refresh page unpvote is checked again.

Sekhmet:
I can't still get that upvote setting to work properly:
upvote

jm90m:
works for me on latest build ?

Where it upvotes itself if checked

busy-test-upvote-pr-12-13

Kennybll:
@jm90m that works, but trying to keep it unchecked is causing me problems. I can't seem to figure it out.

jm90m:
@Kennybll ohhh yeah i see.. gl with it, once you fix it and @Sekhmet approves it, then we can merge.

jm90m:
Just tested this recently, I still can't get this to work for unchecked.

Sekhmet:
I will take a look at this issue.

Sekhmet:
I have updated it there: #1233.

PR #1143, opened 2017-12-02
Update error parsing

### LINKED ISSUE #1140: Error messages from sc2-sdk are not visible anymore for user
We used to have an error message when for example we try to do a second vote before waiting 3sec. But now when something fail it's showing `124 on_message` for any error.
image

That's very confusing for end user.

Sekhmet:
Fixes #1140 

Changes:
- Error description format has been changed. Updated description parsing to properly retrieve human readable string


src/client/errorMiddleware.js
```diff
@@ -1,7 +1,7 @@
 import { notify } from './app/Notification/notificationActions';
 
 function parseBlockChainError(error) {
-  return error.split(':').slice(-1)[0];
+  return error.split(':')[2].split('\n')[0].substring(1);
```
jm90m: This is assuming error is always going to be a string, but hopefully it always is, I would suggest using lodash since it already comes with polyfills and errorchecks and maybe splitting this up into a few different lines that way they aren't reliant too much on each other. Since right now for the first `.split` to work then `error` needs to be a string. Then if that passes but the array results for the index `2` is undefined then it will throw an error because it expects the array result for the index `2` to be a string so you can split it again..

So I think you can do something like this 
```
const errorArray = _.split(":");
const fullErrorText = _.nth(errorArray, 2);
let errorMessage = "  " 
if (!_.isEmpty(fullErrorText)) {
  errorMessage = _.split(fullErrorText, '\n')
  errorTest = _.nth(errorMessage, 0) } || " ";
  return  errorTest.substring(1); // or maybe lodash has something
}
```

Or something like, idk though, your call.. I just quickly did that example so I could be wrong on that lol

PR #1144, opened 2017-12-02
Add feed fetched property

### LINKED ISSUE #1142: Message "Oops! This feed empty" appear even if content has not yet been loaded
Message "Oops! This feed empty" should not be visible if content has not been loaded yet. When i open busy.org, i can see "Oops! This feed empty" for few millisecond:
image
Then it disappear and i can see the loading story placeholder then the stories.
We should not show this message when there is actually something in the feed. That's confusing.

Sekhmet:
Fixes #1142 

Changes:
- Add `fetched` property to feed. If feed hasn't fetched yet don't display empty feed message.


bonustrack:
Can we solve the same issue in user feed too? https://busy-master-pr-1144.herokuapp.com/@webcoop

PR #1152, opened 2017-12-04
Consolidates Editor Components

### LINKED ISSUE #913: [Proposal] Consolidate Editor Components
The codebase currently contains 3 separate, but very similar, Editor components:
- `components/Editor/Editor.js`
- `components/Comments/EmbeddedCommentForm.js`
- `components/Comments/CommentForm.js`

All three of these components are rather long, in particular the Editor itself.

I think we should discuss a way of extracting common functionality. For example, the `insertImage` function between the three files is very similar, and is identical between CommentForm and EmbeddedCommentForm.

Two approaches that spring to mind:
1. Extract common functions like `insertImage` to helper methods, converting any values assumed to be on `this` into additional parameters.
or
2. Create a base class component (`BaseEditor`, but maybe with a better name), which contains the core functionality of an editor, and then all 3 implemented editors extend the base class, getting the benefit of common functionality implemented while being able to override / expand upon functionality.

Approach 1 is likely quicker. Approach 2 is likely more proper, but should probably only be done if there's enough overlap between the functions in the 3 components.

Curious to hear everyone's thoughts.

P.S. Please label this `pr: Discuss`.

Kennybll:
Fixes #913.

Changes:
- Adds EditorBase as a React High Order Component. This holds all overlapping code from Editor, CommentForm, and EmbeddedCommentForm.
- Editor, CommentForm and EmbeddedCommentForm is then changed to use EditorBase,


src/client/components/Comments/EmbeddedCommentForm.js
```diff
@@ -194,11 +120,11 @@ class EmbeddedCommentForm extends React.Component {
             accept="image/*"
             maxSize={MAXIMUM_UPLOAD_SIZE}
             onDropRejected={this.props.onImageInvalid}
-            onDrop={this.handleDrop}
-            onDragEnter={this.handleDragEnter}
-            onDragLeave={this.handleDragLeave}
+            onDrop={f => this.props.handleDrop(this.insertImage, f)}
```
jm90m: should probably call this `files` instead of `f` for more context

src/client/components/Editor/EditorBase.js
```diff
@@ -0,0 +1,147 @@
...
+  class editorBaseComponent extends React.Component {
+    static propTypes = {
+      onImageInserted: PropTypes.func,
+      onImageInvalid: PropTypes.func,
+    };
+
+    static defaultProps = {
+      onImageInserted: () => {},
+      onImageInvalid: () => {},
+    }
```
jm90m: missing semi colon here

src/client/components/Editor/EditorBase.js
```diff
@@ -0,0 +1,147 @@
...
+
+    state = {
+      isDisabledSubmit: false,
+      imageUploading: false,
+      dropzoneActive: false,
+    };
+
+    updateIsDisabledSubmit = (isDisabledSubmit) => {
```
jm90m: I think we should bind all these in the constructor instead of using arrow functions from now on.

https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

src/client/components/Editor/EditorBase.js
```diff
@@ -0,0 +1,147 @@
...
+      handleDragLeave: this.handleDragLeave,
+      handleDragEnter: this.handleDragEnter,
+    };
+
+    render() {
+      return <WrappedComponent {...this.props} {...this.newProps} {...this.state} />;
+    }
+  }
+
+  editorBaseComponent.displayName = `EditorBase(${getDisplayName(WrappedComponent)})`;
```
jm90m: whats the point of this?

src/client/components/Editor/EditorBase.js
Kennybll: I can't see to figure out how, could you help @jm90m

Kennybll: Nevermind, I figured it out.

Kennybll:
Is this good @Sekhmet ? Also does anything need to change in my code for the new img.busy.org?

src/client/components/Editor/EditorBase.js
```diff
@@ -0,0 +1,140 @@
...
+              this.setState({
+                imageUploading: false,
+              }),
+            );
+          }
+        });
+      }
+    }
+
+    newProps = {
```
jm90m: I would bind these in the constructor.... idk why you would bind them here....

Sekhmet:
I don't see much activity there from Kenny so I will close this PR and work on this myself.
Thanks for contributing.

PR #1163, opened 2017-12-06
Add missing strings to json file from page forbidden

espoem:
Changes:
- Add missing strings from the page forbidden that informs a user that he needs to log in.
- The strings were missing so they were not pushed to crowdin and could not be translated.


src/client/statics/Error401.js
```diff
@@ -18,16 +18,16 @@ const Error401 = () => (
...
-            id="@statics/need_login"
+            id="need_login"
             defaultMessage="You need to login to use this feature, please"
           />{' '}
           <a href={SteemConnect.getLoginURL()}>
             <FormattedMessage id="login" defaultMessage="Login" />
           </a>
           <br />
-          <FormattedMessage id="@statics/here_link" defaultMessage="Here's a link to the" />
+          <FormattedMessage id="here_link" defaultMessage="Here's a link to the" />
```
Sekhmet: I think there might be a case where this order wouldn't make sense (in some languages).
I think we can embed link inside `here_link`.

espoem:
Does it mean you want me to make that change? I'm not really sure. I was simply adding them to the json file.

Sekhmet:
If you can, you can do this yourself.

espoem:
@Sekhmet I'm sorry, I don't know how I would achieve a good solution. I am going to submit an issue instead, explaining which strings ids need to be fixed.

Sekhmet:
I can create diff for you so you can apply it and commit if you want.

espoem:
I will leave it to you. Working with git is not a problem but creating the formatted message with embedded link so that it could be processed as one whole string is hard for me. Working on the issue with all details

PR #1164, opened 2017-12-07
Feature/add follows you

### LINKED ISSUE #964: [Feature Proposal] "Follows You" indicator
Per our new development guidelines, I wanted to have discussion / get approval to begin working on a feature I proposed on Utopian.

You can read the full post here: [Utopian: Busy Feature Idea: "Follows You" indicator](https://utopian.io/utopian-io/@ryanbaer/busy-feature-idea-follows-you-indicator)

But here are some screenshots which convey what I had in mind:
image

image

eastmaels:
Fixes #964.

Changes:
- ` src/client/components/UserHeader.js` - add `isFollowingYou` flag
- `src/client/components/UserHeader.less` - added style for "Follows You"
- ` src/client/helpers/apiHelpers.js` - added `follows_you` result to `getAccountWithFollowingCount()`
- `src/client/locales/en.json` - added `follows_you` string variable
- `src/client/user/User.js` - added authenticated user's name as parameter to `getAccountWithFollowingCount()`
- `src/client/user/UserHero.js` - add `isFollowingYou` property


src/client/components/UserHeader.js
```diff
@@ -85,6 +86,11 @@ const UserHeader = ({
         </div>
         <div className="UserHeader__handle-rank-container">
           <div className="UserHeader__row UserHeader__handle">@{handle}</div>
+          {!isSameUser && isFollowingYou && (
+            <div className="UserHeader__followsyou">
```
jm90m: Correct me if I'm wrong but, I think this should be `UserHeader--follows-you`

As @Sekhmet mentioned before we use http://getbem.com/introduction/ for our class names

src/client/components/UserHeader.less
```diff
@@ -26,6 +26,7 @@
     .UserHeader__rank,
     .UserHeader__rank & > i,
     .UserHeader__handle,
+    .UserHeader__followsyou,
     .UserHeader__more,
     .UserHeader__user__username {
       color: #ffffff !important;
```
jm90m: this should be a constants.. `@white`

src/client/components/UserHeader.less
```diff
@@ -68,6 +69,11 @@
   }
 }
 
+.UserHeader__followsyou {
+  margin-top: -30px;
```
jm90m: I'm not a fan of large margin and positioning elements like this, I prefer using flex box but I think that would involve restructuring html elements around here and there, so if @Sekhmet thinks its fine then it's fine with me

jm90m:
@pareshUD could use your inputs on designs for this

eastmaels:
Thank you for the review @jm90m . I learn from this.

src/client/components/UserHeader.js
eastmaels: Changed to `UserHeader--follows-you`

src/client/components/UserHeader.less
eastmaels: Changed to `color: @white !important;`

src/client/components/UserHeader.less
eastmaels: Will wait for feedback.

bonustrack:
I would prefer a design like this:
image
Maybe @pareshUD you have better proposal.

Also it doesnt seem to work when i open a fresh new tab with a profile of someone that following me.

eastmaels:
Thank you for the feedback. I'll look into the issue why it doesn't display on a new tab.

eastmaels:
I'm guessing it's because of the timing of the return of the authenticated user's name. do you have any hints or tips on that?

Sekhmet:
I think it should fix that:
```diff
diff --git a/src/client/user/User.js b/src/client/user/User.js
index 8204699..6f402e6 100644
--- a/src/client/user/User.js
+++ b/src/client/user/User.js
@@ -51,7 +51,7 @@ export default class User extends React.Component {
   };
 
   componentDidMount() {
-    if (!this.props.user.name && this.props.authenticatedUser.name) {
+    if (this.props.user.name && this.props.authenticatedUser.name) {
       this.props.getAccountWithFollowingCount({ name: this.props.match.params.name,
         authUser: this.props.authenticatedUser.name });
     }
```

src/client/user/usersActions.js
```diff
@@ -5,9 +5,9 @@ export const GET_FOLLOWING_COUNT_START = '@users/GET_FOLLOWING_COUNT_START';
 export const GET_FOLLOWING_COUNT_SUCCESS = '@users/GET_FOLLOWING_COUNT_SUCCESS';
 export const GET_FOLLOWING_COUNT_ERROR = '@users/GET_FOLLOWING_COUNT_ERROR';
 
-export const getAccountWithFollowingCount = ({ name }) => dispatch =>
+export const getAccountWithFollowingCount = ({ name, authUser }) => dispatch =>
```
Sekhmet: We don't need to pass `authUser` as parameter. We can get access to state from there `getState()` method, and then use selector `getAuthenticatedUser()`.

src/client/user/User.js
```diff
@@ -90,6 +93,10 @@ export default class User extends React.Component {
     const title = `${displayedUsername} - Busy`;
 
     const isSameUser = authenticated && authenticatedUser.name === username;
+    let followsYou = false;
+    if (user.followsYou && user.followsYou.length > 0) {
+      followsYou = (user.followsYou[0] === authenticatedUser.name);
```
Sekhmet: We don't need those parentheses.

src/client/user/User.js
```diff
@@ -51,14 +51,17 @@ export default class User extends React.Component {
...
+        authUser: this.props.authenticatedUser.name });
     }
   }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.match.params.name !== this.props.match.params.name) {
-      this.props.getAccountWithFollowingCount({ name: this.props.match.params.name });
+    if (prevProps.match.params.name !== this.props.match.params.name
+      && this.props.authenticatedUser.name) {
+      this.props.getAccountWithFollowingCount({ name: this.props.match.params.name,
```
Sekhmet: It doesn't look to good on the same line, I would put this object properties on the new line each.

src/client/user/User.js
```diff
@@ -51,14 +51,17 @@ export default class User extends React.Component {
   };
 
   componentDidMount() {
-    if (!this.props.user.name) {
-      this.props.getAccountWithFollowingCount({ name: this.props.match.params.name });
+    if (!this.props.user.name && this.props.authenticatedUser.name) {
+      this.props.getAccountWithFollowingCount({ name: this.props.match.params.name,
```
Sekhmet: It doesn't look to good on the same line, I would put this object properties on the new line each.

src/client/user/usersActions.js
eastmaels: I still have to fix this. For the other review comments, i'll get to it when I've resolved this change of using `getState()`

src/client/components/UserHeader.js
```diff
@@ -83,7 +84,12 @@ const UserHeader = ({
             )}
           </div>
           <div className="UserHeader__handle-rank-container">
-            <div className="UserHeader__row UserHeader__handle">@{handle}</div>
+            <div className="UserHeader__row UserHeader__handle">
+              @{handle}&nbsp;
```
Sekhmet: Can we use `{' '}` instead of `&nbsp;`?

eastmaels: okay.

eastmaels: the `{' '}` didn't work as space. returned to `&nbsp;`

jm90m: Why cant you just use ``{`@${handle}   ` }`` ?

jm90m: oh yeah @Sekhmet that can work too

eastmaels: > Why cant you just use {`@${handle} ` }

I tried that too. The space gets trimmed.

eastmaels: > I feel like those could be separated using margin, instead of hard coded space.

Apply a style class separately for the `@handle` and `follows you` text?

eastmaels: Currently, the code follows the format `@handle follows you`.
On which style can the margin be applied?

Code:
```
            <div className="UserHeader__row UserHeader__handle">
              @{handle}&nbsp;
              {!isSameUser && isFollowingYou && (
                <FormattedMessage id="follows_you" defaultMessage="follows you" />
              )}
            </div>
```

eastmaels: Put the 'follows you' in a `span`?

src/client/components/UserHeader.js
```diff
@@ -83,7 +84,12 @@ const UserHeader = ({
...
+              {!isSameUser && isFollowingYou && (
```
jm90m: I don't really like this triple conditional expression. It makes the code look ugly. Can you put this up top and create a constant like `const displayFollowsYou = !isSameUser && isFollowingYou`

jm90m: his provides way more context on what the expression actually does

src/client/components/UserHeader.js
jm90m: you can do `.UserHeader__handle > span { margin-left: 5px; }` this targets the first span that gets generated by `<FormattedMessage/>` components

Sekhmet:
For me `@username follows you` is not a sentence, but two different things (user handle, and a follow indicator).
This is the same thing Twitter is doing:
image
I think we should separate handle from follow indicator, maybe add `UserHeader__follow-indicator` span and instead of using space (which implies it's one logical sentence) use margin-left on that element.

Sekhmet:
We can make changes later on, I think we can merge it now if you can resolve those conflicts @eastmaels.

eastmaels:
I'll incorporate the changes now.

eastmaels:
Hello @Sekhmet @jm90m , updated the codes to separate the `follows you` indicator.

I retained the class name as `UserHeader--follows-you-indicator` to follow BEM as mentioned by @jm90m in another comment:
> As @Sekhmet mentioned before we use http://getbem.com/introduction/ for our class names

eastmaels:
I'm not sure if you encounter this when you open a user in another tab. Here's an error in my console using local dev-server:

image

eastmaels:
Also, this is how it looks like when toggled as mobile device in Chrome:

image

jm90m:
Hey, I just took a look, and yeah on mobile it is messed up.

screen shot 2017-12-18 at 11 07 43 pm

You might need to do some rearranging of the different html elements there to make it look better.

`Follows you` should probnably be in its own line

@bonustrack has final say on design though

eastmaels:
Thanks @jm90m . Will wait for the final design.

bonustrack:
@eastmaels I agree with @jm90m we should try to have "Follow you" in a new line on small screen (but not on large screen)

bonustrack:
Unless @pareshUD have better idea?

eastmaels:
Something like this?

image

bonustrack:
Yes

eastmaels:
Okay. I hope @pareshUD confirm that this is okay.

eastmaels:
Okay, I'll just update the codes to put **Follows You** on a new line. And then if there will be changes, will address it in a separate issue. 

Regarding @jm90m's comment regarding using flex in exchange of large margins, will address it later since I'm not yet familiar with how flex works.

> I'm not a fan of large margin and positioning elements like this, I prefer using flex box but I think that would involve restructuring html elements around here and there, so if @Sekhmet thinks its fine then it's fine with me

Sekhmet:
@eastmaels 
https://user-images.githubusercontent.com/29425738/34134574-5545cab8-e496-11e7-9ec8-4a62c40221ae.png
This error is caused by the fact that we use https://reactjs.org/docs/react-dom.html#hydrate (to make it faster on server - it expects that DOM structure rendered on client is the same as the DOM structure that server returned). Nothing to worry about when running locally.

eastmaels:
Desktop layout:

image

Mobile:

image

src/client/components/UserHeader.js
```diff
@@ -87,11 +87,6 @@ const UserHeader = ({
           <div className="UserHeader__handle-rank-container">
             <div className="UserHeader__row UserHeader__handle">
               @{handle}
-              {displayFollowsYou && (
-                <span className="UserHeader--follows-you-indicator">
-                  {intl.formatMessage({ id: 'follows_you', defaultMessage: 'follows you' })}
```
Sekhmet: Can you use FormattedMessage instead of formatMessage?
If you changed text to Follows you, you should update defaultMessages as well.

jm90m:
This is broken for me... 
screen shot 2017-12-20 at 5 20 25 am

desktop layout

jm90m:
@bonustrack @Sekhmet is this broken for you in desktop also?

jm90m:
This is broken for me on multiple user pages

screen shot 2017-12-20 at 5 29 40 am
screen shot 2017-12-20 at 5 29 50 am

src/client/components/UserHeader.less
```diff
@@ -68,6 +69,10 @@
   }
 }
 
+.UserHeader--follows-you-indicator {
+  margin-top: -30px;
```
jm90m: You really should not be using this... to position elements, this gives the element almost a fixed position

jm90m: Probably why it's breaking on my screen

Sekhmet:
For me this PR is broken, when I go to someone profile it doesn't seem to load properly:
image
I'm the only one with this problem? I deleted cache/tried on incognito mode. It happens for also on FF (in Browserstack) so I don't think it's problem with my setup.

jm90m:
@Sekhmet I see it with some profiles too...

jm90m:
adding label `Blocked` to this until these issues are resolved.

src/client/components/UserHeader.less
eastmaels: Helo @jm90m. Can you advice on how to position? i'm no css expeet and need to learn.

jm90m: look up css flex box, lots of resources online for responsive layouts.

src/client/components/UserHeader.js
eastmaels: okay.

src/client/components/UserHeader.less
eastmaels: ok

Sekhmet:
@eastmaels I might take a look at why it does brake if you want, or if you need any help just let me know.

jm90m:
If you can't figure this out, lmk... I will take over this PR and update the styles.

jm90m:
Closing this PR... I will take over and update it in here: https://github.com/busyorg/busy/pull/1271

PR #1165, opened 2017-12-07
Display all menu navigation for loggedin mobile users

### LINKED ISSUE #991: Replies are not accessible on mobile device anymore

jm90m:
screen shot 2017-12-07 at 5 57 28 am

* Only left sidebar menu items appear in popover menu when in mobile view

Fixes #991 

Changes:
- add My Profile, Feed, News, Replies, Wallet links to pop over menu in top bar for mobile users
- add fullScreenHidden param option to PopOverMenu to hide option when full screen


src/client/Wrapper.js
```diff
@@ -38,6 +39,7 @@ export default class Wrapper extends React.PureComponent {
     route: PropTypes.shape().isRequired,
     user: PropTypes.shape().isRequired,
     locale: PropTypes.string.isRequired,
+    username: PropTypes.string,
```
Sekhmet: I think that would be nicer to put it below required prop types.

jm90m: will do!

PR #1189, opened 2017-12-12
Add market price on feed for every crypto

### LINKED ISSUE #1170: Add market price on feed for every crypto
When someone visit a tag feed that is related to another crypto for example:
https://busy.org/trending/bitcoin
https://busy.org/trending/ethereum
https://busy.org/trending/eos
...
We could show on right sidebar the market price for the specific crypto like this:
image
That give quick good insight for users visiting these pages.

This was suggested by `@misix`

jm90m:
Fixes #1170

Changes:
- Add FeedSidebar component & CryptoTrendingCharts component
- add cryptosHelper to search for cryptos within the top 100 crypto constants

screen shot 2017-12-12 at 3 29 34 pm
screen shot 2017-12-12 at 3 29 49 pm

### DIFF
--- a/src/client/components/Sidebar/CryptoTrendingCharts.js
+++ b/src/client/components/Sidebar/CryptoTrendingCharts.js
@@ -1,0 +1,116 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Trend from 'react-trend';
+import fetch from 'isomorphic-fetch';
+import _ from 'lodash';
+import { FormattedMessage } from 'react-intl';
+import Loading from '../Icon/Loading';
+import { getCryptoDetails } from '../../helpers/cryptosHelper';
+import USDDisplay from '../../components/Utils/USDDisplay';
+import './SteemTrendingCharts.less';
+
+const fetchCryptoPriceHistory = symbol =>
+  fetch(
+    `https://min-api.cryptocompare.com/data/histoday?fsym=${symbol}&tsym=USD&limit=7`,
+  ).then(res => res.json());
+
+class CryptoTrendingCharts extends React.Component {
+  static propTypes = {
+    crypto: PropTypes.string,
+  };
+
+  static defaultProps = {
+    crypto: '',
+  };
+
+  constructor(props) {
+    super(props);
+    const currentCrypto = getCryptoDetails(props.crypto);
+
+    this.state = {
+      currentCrypto,
+      currentCryptoPriceHistory: [],
+      loading: false,
+    };
+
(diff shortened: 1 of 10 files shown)


bonustrack:
Can we have the a margin-bottom of 10px on the market block?
Like this: https://user-images.githubusercontent.com/16245250/33907326-5d03506a-dfb7-11e7-91b1-682aaffa7fef.png
Not like this https://user-images.githubusercontent.com/16245250/33907345-71995f60-dfb7-11e7-8d5b-f938c13f242e.png

jm90m:
@bonustrack yup just updated it, please check again, thx

src/client/components/Sidebar/CryptoTrendingCharts.js
```diff
@@ -0,0 +1,116 @@
...
+import { getCryptoDetails } from '../../helpers/cryptosHelper';
+import USDDisplay from '../../components/Utils/USDDisplay';
+import './SteemTrendingCharts.less';
+
+const fetchCryptoPriceHistory = symbol =>
+  fetch(
+    `https://min-api.cryptocompare.com/data/histoday?fsym=${symbol}&tsym=USD&limit=7`,
+  ).then(res => res.json());
+
+class CryptoTrendingCharts extends React.Component {
```
bonustrack: We should avoid having duplicate code, i think this component should also be used on the Steem and SBD market block on wallet pages.

jm90m: Steem & SBD market blocks are both combined into one, thats why I kept the SteemTrendingCharts component... 

I can make it look like this if you want? 

screen shot 2017-12-12 at 4 18 10 pm

and it's not really duplicate code, since if you look at steem trending charts component, there is a few different logic in there.

bonustrack: Can't you just make that the component can take an array of symbol instead of only one?

jm90m: yeah I can but this would only be for one case? just for this specific scenario? Is that worth doing, when we are going to update the charts design anyways?

bonustrack: If we are not doing this we are going to update the charts in 2 different places isn't it?

jm90m: yeah thats true, I'll see what I can come up with .

jm90m:
@bonustrack just updated it so their is now a `CryptoChart` common component that is used. so less duplicate code now.

PR #1195, opened 2017-12-14
Added warning message about phishing links in transfer memo

4leem:
Changes:
- Updated 'UserWalletTransactions.js' file to include the message
- Updated locale/en.js to include the newly added message
- Updated snapshot test spec


src/client/wallet/UserWalletTransactions.less
```diff
@@ -2,7 +2,7 @@
 @import (reference) "../styles/modules/variables.less";
 
 .UserWalletTransactions {
-  background-color: @white;
+  /*background-color: @white;*/
```
jm90m: why is this commented out?

4leem: The newly added text also got the background which looks odd, hence moved the background to 
.UserWalletTransactions__transaction

4leem: With these change, the UI would something like this: 

image

Just in time, as you can see in screenshot, I got a suspicious link through memo.

jm90m: ahh ok, but instead of commenting out the styles, you should just remove it. makes no sense to comment out the styles if it will not be used.

Also @bonustrack has the final say on the design, so he should review this :)

webpack/webpack-dev-server.js
```diff
@@ -21,15 +21,15 @@ module.exports = {
         SENTRY_PUBLIC_DSN: null,
         STEEMCONNECT_CLIENT_ID: JSON.stringify(process.env.STEEMCONNECT_CLIENT_ID || 'busy.app'),
         STEEMCONNECT_REDIRECT_URL: JSON.stringify(
-          process.env.STEEMCONNECT_REDIRECT_URL || 'http://localhost:3000/callback',
+          process.env.STEEMCONNECT_REDIRECT_URL || 'http://localhost:3000/callback'
```
Sekhmet: I see that you removed trailing commas (to fix issue with node older than 8). I don't have anything against about doing this locally, but I would rather not put it in PR (I highly recommend updating to node 8 though). Can you remove those changes from this PR?

src/client/wallet/UserWalletTransactions.less
4leem: Yes, that was my initial thought to remove it instead of commenting out. Anyway I'll remove it, it makes the code look much cleaner.

webpack/webpack-dev-server.js
4leem: yes, you got it right. I will revert it back.

bonustrack:
I dont think we should have a message like this by default on the wallet page. Link in memo are not clickable anyway.

Sekhmet:
@4leem it is always better to create issue first and discuss some feature before starting work.
If you remove that comment in styles it looks good to me, but I'm not sure whether is it something we want or not, so I would let others decide on that.

4leem:
@Sekhmet Agree with you, probably next time it is better to create an issue then work on dev. I have now removed the commented line from style.

jm90m:
Closing this issue since @bonustrack does not approve, he is our CTO

PR #1196, opened 2017-12-14
Use Dtube Embed Player for both Post Preview & Post Full

### LINKED ISSUE #1106: Use the new DTube player instead of default browser player
For post published from DTube we should use their new player embed instead of default browser player. Here is the `src` we need for get the player:
`https://emb.d.tube/#!/salut/1nuxiyu0`
We need an iframe for it

jm90m:
Fixes #1106.

Changes:
- use new d.tube embed url with iframe

go to url `/trending/dtube` and look at dtube videos from there using the new dtube embed player


bonustrack:
We can go forward with this feature. Can you fix conflicts @jm90m ?

jm90m:
yeah will fix conflicts @bonustrack

Issue #1198, opened 2017-12-14
Update For All Translations


Kennybll:
#### Expected behavior
We have 51 translation files under [/src/client/locales/*.json](https://github.com/busyorg/busy/tree/master/src/client/locales). But on [Settings.js](https://github.com/busyorg/busy/blob/master/src/client/settings/Settings.js) and [translations/index.js](https://github.com/busyorg/busy/blob/master/src/client/translations/index.js) there is support for  40 of the translation files.

#### Actual behavior
We should have support for all 51 files.

#### How to reproduce
Look at the links above, and count.

* Browser: Chrome
* Operating system: Windows 10

Sekhmet:
I don't think we should enable incomplete translations. Seeing only 40% of page translated would be annoying.

Sekhmet:
We will only make complete (usable) translations available to users.

PR #1200, opened 2017-12-15
Code Splitting React Intl Locales

### LINKED ISSUE #1197: Code Splitting React Intl Locales
#### Expected behavior
We should split the react-intl locales to make the bundle smaller. I figure to use webpacks dynamic inport(). This will load it separately. I tried this and it saves around 400k.

#### Actual behavior
Bundles all translations to the main bundles.

#### How to reproduce
[https://busy.org/js/statistics.html](https://busy.org/js/statistics.html) shows translations take up 37.6%.

I started to work on it. I am about to add a pull request. Just thought to add it here first. I am willing to be assigned.

* Browser: 
* Operating system:

Kennybll:
Fixes #1197.

Changes:
- Uses webpack's dynamic import function to load translations as bundles.
- Removes translations/index.js because it would no longer be needed.
- This adds support to all translations for when settings is updated (#1198)
- Removes about 400k from the main bundle. [https://kennybll.herokuapp.com/js/statistics.html](https://kennybll.herokuapp.com/js/statistics.html)


src/client/Wrapper.js
```diff
@@ -73,6 +78,42 @@ export default class Wrapper extends React.PureComponent {
...
+    let locale = appLocale || 'auto';
+
+    if (locale === 'auto') {
+      locale = this.getBrowserLocale() || 'en';
+    }
+
+    return locale;
+  };
+
+  getBrowserLocale = () => {
```
jm90m: bind functions in constructor please

src/client/Wrapper.js
```diff
@@ -73,6 +78,42 @@ export default class Wrapper extends React.PureComponent {
...
+      import(`react-intl/locale-data/${this.getAvailableLocale(nextProps.locale)}`).then((localeData) => {
+        addLocaleData(localeData);
+        import(`./locales/${this.getAvailableLocale(nextProps.locale)}.json`).then((data) => {
+          this.setState({ translations: data, loading: false });
+        });
+      });
+    }
+  }
+
+  getAvailableLocale = (appLocale) => {
```
jm90m: bind functions in constructor please

src/client/Wrapper.js
```diff
@@ -73,6 +78,42 @@ export default class Wrapper extends React.PureComponent {
...
+    }
+
+    return locale;
+  };
+
+  getBrowserLocale = () => {
+    let detectedLocale;
+    if (typeof navigator !== 'undefined') {
+      detectedLocale =
+        navigator.userLanguage ||
```
jm90m: Lots of logic happening here, you need to split this up and break it down to make it more readable. You should not have multiple ternary operations like this, im surprised eslint didn't pick this up

src/client/Wrapper.js
```diff
@@ -73,6 +78,42 @@ export default class Wrapper extends React.PureComponent {
     this.props.getTrendingTopics();
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.locale !== this.props.locale) {
+      this.setState({ locale: this.getAvailableLocale(nextProps.locale) });
```
jm90m: Does this actually work for anyone @Sekhmet @bonustrack ? I took the time and ran this on my local..

screen shot 2017-12-15 at 5 26 28 am

and keep seeing these errors when Im switching languages

src/client/Wrapper.js
```diff
@@ -73,6 +78,42 @@ export default class Wrapper extends React.PureComponent {
...
+      import(`react-intl/locale-data/${this.getAvailableLocale(nextProps.locale)}`).then((localeData) => {
```
jm90m: I know dynamic imports were not recommended in the past before by the creator of redux - https://github.com/facebookincubator/create-react-app/issues/2176

and he works for facebook, so idk about this one and I'm not familiar with these

jm90m:
I added some comments, but @bonustrack / @Sekhmet might know more about dynamic importing then me :)

src/client/Wrapper.js
Sekhmet: I don't think that's too bad here. To me it looks decent, except that `(navigator.languages && navigator.languages[0] ? navigator.languages[0] : undefined);` could be replaced just with `navigator.languages && navigator.languages[0]` and it does exactly the same thing.

eslint didn't complain, because there is are no nested ternary operators (which causes eslint errors).

jm90m:
hey @Kennybll, @Sekhmet deployed this.... on here https://busy-master-pr-1200.herokuapp.com/

And it doesn't render anything.... so idk if this will work for us. thanks for the contribution though

Sekhmet:
It works only locally (running `dev-server`). It doesn't work on production build (with SSR) rendering.

Sekhmet:
This PR doesn't seem to be updated, so I created another one: #1237.

PR #1203, opened 2017-12-15
Add Crypto Market Price to each Post

### LINKED ISSUE #1201: Add crypto market price on single post
We can show market price on single post when a post is using a tag related to a crypto symbol: for example when a post has these tags: `bitcoin`, `eth`, `trading`, `eos` we would show on right sidebar of the single post page the block "Market" with price of BitCoin Ethereum and EOS.

See:
image

jm90m:
Fixes #1201 

Changes:
- Make CryptoTrendingCharts component take an array of tags to render
- Remove SteemTrendingCharts component
- Add CryptoTrendingCharts to posts with crypto tags


src/client/components/Sidebar/CryptoChart.js
```diff
@@ -61,18 +69,27 @@ class CryptoChart extends React.Component {
       loading: true,
     });
     fetchCryptoPriceHistory(symbol).then((response) => {
```
Sekhmet: We should handle catch as well to prevent unhandled promise error.

jm90m: good point will add

jm90m:
@Sekhmet added catch block

Sekhmet:
I feel like the way we use for refreshing is a little bit too much hacky.
Personally, I would move chart data to redux store. I would create an action `loadPrices(SYM)` and dispatch it for all needed currencies on `handleOnClickRefresh`.
I would create `CryptoTrendingChartsContainer` that adds `loadPrices action`, and `CryptoChartContainer` that subscribes to its state in redux.

jm90m:
@Sekhmet I guess moving it to the redux store would be alot better, and passing it through to reducers probably in the `appReducer` where we store rate there.

PR #1205, opened 2017-12-16
Add AMP

### LINKED ISSUE #1146: Implement AMP
Can we implement AMP? https://www.ampproject.org/ That should improve our Google ranking

Sekhmet:
Fixes #1146

Changes:
- [x] Add AMP template.
- [x] Add `ampRender`.
- [x] Prepare content for AMP (replace images, remove `head`, select content inside `body`).
- [x] Link pages using canonical URLs.
- [x] Pass images to schema.
- [x] Extract renderers.
- [x] Extract handlers.


bonustrack:
I'm not able to see preview on this post for example: https://busy-master-pr-1205.herokuapp.com/steemit/@aggroed/steem-works-a-funding-mechanism-for-a-worker-proposal-system/amp . I'm testing using this https://search.google.com/test/amp?view=search-preview&id=XB9lHG-FbRpPhSka3bUspQ

bonustrack:
image

Sekhmet:
I've clicked on the link you provided and I can see preview:
image
And that's what I see for: https://busy-master-pr-1205.herokuapp.com/utopian-io/@sekhmet/my-busy-org-october-report-voting-slider-avatar-lightbox-and-a-lot-more/amp
image

bonustrack:
@jm90m is the image preview visible for you?

Sekhmet:
@jm90m 
**About handlebars:**
Currently we use normal HTML pages and HTML comments for inserting content on server. It's fine for SSR, but not very effective for AMP template (which needs a lot more data). With handlebars we can manipulate AMP page without even touching JS code, because handlebars handles logic on template level. You can test it the same way as HTML templates (and it both would be pretty boring, because both .hbs and .html would not have dynamic content.

**About cheerio:**
we need virtual DOM manipulator on server because we need something to parse post contents. We have really three options there (we can't use native DOM manipulation as we don't have access to it on the server):
1. Parsing it manually (Regex or parsing it in other way).
2. JSDOM - which is strict about HTML correctness (not a good thing for dealing with HTML content coming from a lot of users and a lot of different clients), slower and heavier.
3. cheerio - which is faster and handles not-so-correct HTML pages better. It only uses jQuery syntax (which I think is still great).
I picked cheerio because I was already familiar with it and it seems to fit our needs well.

jm90m:
@Sekhmet thanks for the in-depth explanations, makes sense, looks like cheerio was a good choice 👍

PR #1215, opened 2017-12-18
Create SidebarContentBlock.less styles and apply it to all sidebar

### LINKED ISSUE #1211: Sidebar right blocks should use same class
It doesn't make sense to have multiple classes for sidebar blocks like `PostRecommendation` or `CryptoTrendingCharts` or `InterestingPeople` etc while they sharing the same style. We should use a single class for them and avoid code duplicate, for the sidebar block but also sidebar block title. Sidebar right blocks should have margin in bottom of 10 pixels. Now some block dont have this bottom margin:
image

jm90m:
Fixes #1211.

Changes:
- Update all sidebar blocks with content (crypto charts, interesting people, post recommendation) to use same styles for class name for background, title, icon and divider
- Also update cryptosHelper - getCryptoDetails to find exact matches for posts tagged with crypto
- Also update loader styles to always be purple

### DIFF
--- a/src/client/components/Sidebar/InterestingPeopleWithAPI.js
+++ b/src/client/components/Sidebar/InterestingPeopleWithAPI.js
@@ -7,9 +7,9 @@ import User from './User';
 import Loading from '../../components/Icon/Loading';
 import steemAPI from '../../steemAPI';
 import './InterestingPeople.less';
+import './SidebarContentBlock.less';
 
-@withRouter
-class InterestingPeopleWithAPI extends Component {
+@withRouter class InterestingPeopleWithAPI extends Component {
   static propTypes = {
     authenticatedUser: PropTypes.shape({
       name: PropTypes.string,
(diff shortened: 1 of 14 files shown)


src/client/components/Sidebar/InterestingPeopleWithAPI.js
```diff
@@ -7,9 +7,9 @@ import User from './User';
 import Loading from '../../components/Icon/Loading';
 import steemAPI from '../../steemAPI';
 import './InterestingPeople.less';
+import './SidebarContentBlock.less';
 
-@withRouter
-class InterestingPeopleWithAPI extends Component {
+@withRouter class InterestingPeopleWithAPI extends Component {
```
Sekhmet: Can we have annotation on different line and extend from `React.Component`?

jm90m: yeah lol, forgot this was old stuff that got merged in, will update now

jm90m: We should have eslint for this stuff :P

bonustrack:
@jm90m can you use also the `SidebarContentBlock` and title class for `UserActivitySearch`?

jm90m:
@bonustrack will do forgot about that one, good catch

PR #1227, opened 2017-12-21
Display Login Modal for unlogged users when they try to Like/Upvote/Save/etc...

### LINKED ISSUE #315: Direct the logged out users to the login page when they hit Comment

### LINKED ISSUE #469: Show alert when clicking on button action: "need to login" to Like/Upvote/Bookmark

### LINKED ISSUE #803: Implementation details on handling unauthenticated user actions

### LINKED ISSUE #1226: Upvoting post while unlogged out, displays voting slider with NAN
#### Expected behavior
Should either display modal to request user to login or take user to login page directly

#### Actual behavior
clicking-on-vote-for-unlogged

#### How to reproduce

* Browser: Chrome
* Operating system: Max OSX El Capitan 

#### Related Issues
https://github.com/busyorg/busy/issues/469
https://github.com/busyorg/busy/issues/315
https://github.com/busyorg/busy/issues/803

jm90m:
Fixes #1226
Fixes #469
Fixes #315
Fixes #803

screen shot 2017-12-21 at 2 24 38 pm

Changes:
- Create LoginModal Component
- Add authenticated checks to Story/Comments Footer Buttons
- Add authenticated checks to FollowButton
- Add authenticated Story Preview component (by passing authenticated prop from Feed)


jm90m:
as discussed with @Sekhmet, just updated everything to use a HoC `withAuthActions`

Let me know if i did this right @Sekhmet :rofl:

bonustrack:
Can we have bigger buttons like the one in reblog?
image

jm90m:
@bonustrack  yeah we can let me update it to look like reblog buttons

jm90m:
@bonustrack just pushed, made it larger

src/client/components/LoginModal.js
```diff
@@ -0,0 +1,80 @@
...
+          <span className="LoginModal__login-title">
+            <FormattedMessage id="login_to_busy" defaultMessage="Login to Busy" />
+          </span>
+          <span className="LoginModal__login-description">
+            <FormattedMessage
+              id="login_modal_description"
+              defaultMessage="Login with your Steem account using SteemConnect to enjoy Busy at 100%"
+            />
+          </span>
+          <span role="presentation" className="LoginModal__login-button" onClick={this.handleLogin}>
```
Sekhmet: Can we use `ahref` for that? It would make it more accessible.

src/client/components/StoryFooter/Buttons.js
```diff
@@ -34,79 +37,100 @@ export default class Buttons extends React.Component {
...
-  handleCloseReactions = () =>
+  handleCloseReactions() {
     this.setState({
       reactionsModalVisible: false,
     });
+  }
 
-  handleCommentClick = () => {
+  // eslint-disable-next-line class-methods-use-this
+  handleCommentClick() {
```
Sekhmet: I would rather make it static method than disable the rule.

jm90m: sure we can make it a static method

src/client/components/LoginModal.js
jm90m: yeah will change now

jm90m:
@Sekhmet prettier runs every time i save a file so it removes the parens, but ill add it back if you think its helpful

jm90m:
@Sekhmet done, re-added parents & made requested changes, i think i got them all lmk if im missing some

PR #1228, opened 2017-12-22
Disable curation rewards on comments

### LINKED ISSUE #1213: Remove curation reward on comments
It's possible to remove curation reward when you publish a comment on Steem with the parameter `allow_curation_rewards: false,` for `comment_options` operation. This doesn't work on root post but we should enable it on comment that would give around 25% more for comment authors.

Sekhmet:
Fixes #1213 

Changes:
- Add `broadcastComment` function that broadcasts `comment` and `comment_options` operations

### DIFF
--- a/src/client/comments/commentsActions.js
+++ b/src/client/comments/commentsActions.js
@@ -61,11 +61,13 @@ export const getComments = (postId, reload = false, focusedComment = undefined) 
   dispatch({
     type: GET_COMMENTS,
     payload: {
-      promise: steemAPI.sendAsync('get_state', [`/${category}/@${author}/${permlink}`]).then(apiRes => ({
-        rootCommentsList: getRootCommentsList(apiRes),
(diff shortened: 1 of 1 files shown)


src/client/comments/commentsActions.js
```diff
@@ -102,24 +143,31 @@ export const sendComment = (parentPost, body, isUpdating = false, originalCommen
...
+        parentAuthor,
+        parentPermlink,
+        author,
+        permlink,
+        '',
+        newBody,
+        jsonMetadata,
+      ).then((resp) => {
+        const focusedComment = {
+          author: resp.result.operations[0][1].author,
```
jm90m: I wish there was a better way to retrieve the author rather than this line `resp.result.operations[0][1]`, looks like a lot of things can go wrong here since you assume that `result.operations` is an array and the first index is the one you want get then you get the 2nd index of that lol... oh well i guess thats what the response returns, can't be helped i guess.

PR #1229, opened 2017-12-22
Update posts state from feed

### LINKED ISSUE #1222: Post reloads after upvoting
#### Expected behavior

No visible action with the post except the stats change should occur.

ezgif com-video-to-gif 1

The gif is from the stable version.

#### Actual behavior

The post reloads if a user upvotes a post.

ezgif com-video-to-gif

#### How to reproduce

1. Open a post on staging.busy.org
2. Upvote the post

* Browser: Chrome

Sekhmet:
Fixes #1222 

Changes:
- Mark post as loaded from feed.

### DIFF
--- a/src/client/post/postsReducer.js
+++ b/src/client/post/postsReducer.js
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import * as feedTypes from '../feed/feedActions';
 import * as bookmarksActions from '../bookmarks/bookmarksActions';
 import * as postsActions from './postActions';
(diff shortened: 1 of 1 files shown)


src/client/post/postsReducer.js
```diff
@@ -57,16 +57,26 @@ const posts = (state = initialState, action) => {
...
     case bookmarksActions.GET_BOOKMARKS_SUCCESS: {
-      const postsTemp = {};
+      const list = {
+        ...state.list,
+      };
+      const postsStates = {
+        ...state.postsStates,
+      };
+
       action.payload.postsData.forEach((post) => {
```
jm90m: I think I remember seeing somewhere that LoDash's _.each is faster for some browsers than the normal native `.forEach` 

https://stackoverflow.com/a/19206409/2884655

Your call whether to use it or not :)

PR #1255, opened 2017-12-24
Add development scripts

Sekhmet:
This adds development scripts inspired by [React](https://github.com/facebook/react/tree/master/scripts) and this [blog post](https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html) about React's repository architecture.

Reasons:
1. Add faster alternative for prettier and eslint.
2. Solve issues with varying prettier behaviour in different editors. When this PR will be merged our own scripts should be used.
3. Prevent committing not properly formatted files to repository.

- [x] Retrieve updated files.
- [x] ESLint
  - [x] Run on all files.
  - [x] Run on updated files.
- [x] Prettier
  - [x] Run on all files.
  - [x] Run on updated files.
  - [x] Detect differences (should fail CI).
  - [x] Add `less` support. 

Integration (another PR).
  Apply [this patch](https://github.com/Sekhmet/busy/blob/c7c012dfe82a8f6d2ba3f46d1cbdb35ef8c18906/enable-scripts.patch) to enable scripts.
  - Format all files.
  - Add pretty check to CI.


enable-scripts.patch
```diff
@@ -0,0 +1,61 @@
+diff --git a/.eslintignore b/.eslintignore
```
Sekhmet: This file is just a patch that will enable those scripts when applied.
I don't want to do it in this PR, because I would have to run `yarn run prettier-all` to make tests pass, but I think it should be done in another commit made especially for reformatting.

PR #1261, opened 2017-12-26
Don't index feed pages

### LINKED ISSUE #1248: Don't index feed pages on Google
Feed page like `/trending/*`, `/created/*`, `/promoted/*`, `/hot/*` are not too much relevant in search engine result, we should tell Google to not index them. Maybe using a robots.txt file. We need Google to following these page and go through all post but we don't need these page on Google directly.

Sekhmet:
Fixes #1248 

Changes:
- Add `noindex` to feed pages.

### DIFF
--- a/src/client/feed/Page.js
+++ b/src/client/feed/Page.js
@@ -57,6 +57,7 @@ class Page extends React.Component {
       <div>
         <Helmet>
           <title>Busy</title>
+          <meta name="robots" content="noindex" />
         </Helmet>
         <ScrollToTop />
         <ScrollToTopOnMount />
--- a/src/client/feed/Page.js
+++ b/src/client/feed/Page.js
@@ -57,7 +57,7 @@ class Page extends React.Component {
       <div>
         <Helmet>
           <title>Busy</title>
-          <meta name="robots" content="noindex" />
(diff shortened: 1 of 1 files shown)


src/client/feed/Page.js
```diff
@@ -57,6 +57,7 @@ class Page extends React.Component {
       <div>
         <Helmet>
           <title>Busy</title>
+          <meta name="robots" content="noindex" />
```
bonustrack: Can you use `noindex,follow`, so Google go through sub links?

Sekhmet: There is no such rule as `follow`. If you want to disable following links you use `nofollow` - following links is default.

bonustrack: I always used `index,follow`, Medium is using it too `view-source:https://medium.com/byteball/byteball-grants-program-906a71b93d3c`. Unless Google said it doesn't matter at all, it's more safe to have it dont you think?

bonustrack:
Does this apply for the homepage aswell @Sekhmet ?

Sekhmet:
Yes, it does. It shouldn't I guess?

bonustrack:
It should not, we want the homepage indexed.

PR #1264, opened 2017-12-26
Update username style

### LINKED ISSUE #1250: Change username links color and font weight
We should use black color on every username links instead of purple and normal instead of bold
image
Similar than medium or hashnode:
image

Sekhmet:
Fixes #1250

Changes:
- Update username style to black colour and normal weight.
- Update Wallet and Activity page styles.


bonustrack:
Can we have reputation font bold in comment:
image

PR #1270, opened 2017-12-27
Use get_account_reputation instead of look_up accounts for search autocomplete

jm90m:
Changes:
- Use get_account_reputation instead of look_up accounts for autocomplete on search
- Sort by user reputation


src/client/helpers/apiHelpers.js
```diff
@@ -110,8 +110,9 @@ export const isWalletTransaction = actionType =>
   actionType === accountHistoryConstants.DELEGATE_VESTING_SHARES ||
   actionType === accountHistoryConstants.CLAIM_REWARD_BALANCE;
 
-export const getLookupAccountNames = (name, limit = 5) =>
-  SteemAPI.sendAsync('lookup_accounts', [name, limit]);
+export const getAccountReputation = (name, limit = 5) =>
```
bonustrack: I think it's safe to request for 20 accounts by default, that should give better results

jm90m: yeah ill update it to request 20

Sekhmet:
I think we are displaying too much users (in my opinion 5 would be enough) - we can load more to get better results, but limit it to 5 on UI.
https://busy-master-pr-1270.herokuapp.com/search?q=sekhmet
Now it's cumbersome if you want to search for something else than users.

bonustrack:
Yes @Sekhmet . My meaning was to load more results from Steem API to get better quality but not showing more result on the UI.

PR #1275, opened 2017-12-27
Add vote value to user profile

### LINKED ISSUE #1272: Show voting value on user profile

Sekhmet:
Fixes #1272

image

Changes:
- Add vote value to user profile.


bonustrack:
When i load the page, `$NaN` appear for few miliseconds, it should not be the case.
image

Sekhmet:
Should it show 0 as placeholder or hide entire line?

bonustrack:
The best would be to have a small loading icon, similar than this 
image

Sekhmet:
Loading icon instead of entire line or to replace $ icon? If the latter what should the value be?

bonustrack:
I would see the line loading like this:
`$ Vote value: [loading-icon]`

PR #1276, opened 2017-12-27
Updates styles

Sekhmet:
Changes:
- Update text styles (username + post).
- Update logo styles.
- Add shadow to Topnav.


src/client/styles/custom.less
```diff
@@ -2,9 +2,11 @@
 
 // Start with assigning color names to specific hex values.
 @white: #fff;
+@white-shadow: rgba(0, 0, 0, 0.11);
```
bonustrack: Why `0.11`? I prefer `0.05` did you tried?

src/client/styles/modules/feed-layout.less
```diff
@@ -2,6 +2,8 @@
 
 .feed-layout {
   position: relative;
+  min-height: 100%;
```
bonustrack: Is this supposed to fix the view being cut when feed page is loading? It doesn't seem to work for me

src/client/components/Navigation/Topnav.less
```diff
@@ -6,17 +6,20 @@
 .Topnav {
   background-color: @white;
   border-bottom: 1px solid #e9e7e7;
+  box-shadow: 0 0 4px @white-shadow;
   height: 56px;
   line-height: 56px;
 
   &__brand {
-    font-size: 28px;
+    font-size: 32px;
```
bonustrack: `30px` is enougth

src/client/styles/modules/feed-layout.less
Sekhmet: I'm not sure how to solve this one. I think JM might know how to do this.

bonustrack: This fix doesnt need to happen in this PR

PR #1278, opened 2017-12-28
Fix "Search bar results are fixed to the top of the site" #1259 

### LINKED ISSUE #853: 3 dots menu should be fixed under the header
#### Menu from the right should stay fixed with the header under the 3 dots and should disappear after clicking on an option

The 3 dots menu should relate to the header and should stay fixed under the 3 dots button from the header.
It should also hide after choosing an option from the list
The same thing goes for the notifications

#### Actual behavior

Now if I click the 3 dots to bring up the menu and then decide to scroll the page before I chose one of the options that menu scrolls up with the page.
Secondly, if I bring up the menu by clicking on the 3 dots and then selecting an option the page updates the content but the menu remains visible until I click outside it
Here's a screenshot
busy

#### How to reproduce
I'm browsing with Chrome on Windows 10 and noticed the behaviour on the homepage but I reckon it relates to all the other pages too

* Browser: Chrome

### LINKED ISSUE #1259: Search bar results are fixed to the top of the site
Busy released version 2.2 of their platform. One of the new features that can be found in the [stable version](busy.org) is searching within the Busy website.

There is a minor issue with the feature. The floating window with the results is fixed to the top of the site. If you write a query to the search bar and scroll down, the window will scroll too and you will not be able to see the results. Or if you are in the middle of the page and want to search anything, you will not be able to see the results because you have to scroll to the top.

## Expected behaviour

The floating window with the results should be fixed under the top navbar, which is always visible. Scrolling down should not cause that the user is not able to pick the suggested search results.

## Steps to reproduce

Case 1

- open busy.org
- write a query in the search bar
- scroll down

Case 2

- open busy.org
- scroll down
- write a query in the search bar

The described behaviour also occurs in staging.busy.org

---
- Browser: Chrome
- OS: Windows 10

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@espoem/search-bar-results-are-fixed-to-the-top-of-the-site">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

sweever:
Fixes #1259
Fixes #853

Changes:
- dir : **busy/src/client/components/Navigation/Topnav.less**
- fixing search result with ` position: fixed;`

Changes:
- dir : **busy/src/client/components/Navigation/Topnav.js**
- fixing user menu with `  overlayStyle={{ position: 'fixed' }}`


Sekhmet:
Hey, thanks for contributing!

Looks like if u did the same for `ant-popover` it would solve this issue as well: #853

sweever:
@Sekhmet it work but i don't think its the same **class** here : 
**busy/src/client/styles/modules/popover-layout.less**

Sekhmet:
Yeah,
It would work if you made this change:
```diff
diff --git a/src/client/styles/modules/popover-layout.less b/src/client/styles/modules/popover-layout.less
index 8977ae0..11411c8 100644
--- a/src/client/styles/modules/popover-layout.less
+++ b/src/client/styles/modules/popover-layout.less
@@ -1,3 +1,7 @@
+.ant-popover {
+  position: fixed;
+}
+
 @media screen and (max-width: 990px) {
   .ant-popover {
     width: 100%;
```

If you want to solve those two issues at once, feel free to apply this patch (or add it by hand).

Sekhmet:
Hey, thanks for updating. I just noticed that this fix (one I suggested) causes another issue - PopoverMenus on feed now remain fixed when scrolling which is not desired result:



Actually, you can revert last change and use [overlayStyle](https://ant.design/components/tooltip/#Common-API) here to set position instead:
https://github.com/busyorg/busy/blob/c7685c3361e2fc0fc7039f306234b638b2e8d541/src/client/components/Navigation/Topnav.js#L115-L118

I think that it should fix this problem and not cause additional bugs.

Let me know if you need help with something.

Sekhmet:
It's working well now.
Note that you don't have to push every change to see if it works.
If you run `yarn run dev-server` you will have server that you can use to verify it everything works before committing and pushing.
If you use normal git client test will run locally before committing.
If you for some reason don't want to have this feature make sure to run `yarn run linc` and `yarn run prettier-check` to check that you didn't broke anything.

sweever:
@Sekhmet i'm in work now and i don't have dev app on this comp , so sorry , just want to complet da work.

PR #1280, opened 2017-12-28
Profile settings

### LINKED ISSUE #1256: Edit profile page with inputs for social accounts

Sekhmet:
Fixes #1256 

image

image

Changes:
- [x] Add UI.
- [x] Add translations.
- [x] Use antd form.
- [x] Generate SteemConnect link.
- [x] Link to `ProfileSettings`.
- [x] Load values into form.
- [x] Use only updated values.
- [x] Update `UserInfo`.


bonustrack:
@Sekhmet 
- Can we have the button "Save" not full-width? We need the same one than settings page
- I think social links should not include the website, i would see another group for it

bonustrack:
Also it may be good to allow only alphanumeric + `-_.` characters to avoid having user who type the full url of their social profile

Sekhmet:
So social links shouldn't be URLs but only usernames? How to handle it if users doesn't have username specified (e.g. Facebook)? What do we want to do with Bitcoin/Ethereum? Would it be just address? How do we display it?

bonustrack:
On Facebook if the username is not specified, the id should work.

bonustrack:
Bitcoin Ethereum can send user to their most popular explorer:
https://blockchain.info/address/1FVKW4rp5rN23dqFVk2tYGY4niAXMB8eZC
https://etherscan.io/address/0x1396df943ec26d1455a53bf2f4bd67ebd139907c

Sekhmet:
Updated, example profile there:
https://busy-master-pr-1280.herokuapp.com/@sekhmet

bonustrack:
I'm thinking that the official BitCoin icon come with the circle around like this:
image
Same thing goes for Facebook with the square:
image

Is it possible that you use these instead?

src/client/components/SocialLink.js
```diff
@@ -0,0 +1,32 @@
...
+    />
+    <a target="_blank" rel="noopener noreferrer" href={url}>
+      {profile.name}
+    </a>
+  </div>
+);
+
+SocialLink.propTypes = {
+  profile: PropTypes.shape({
+    id: PropTypes.string.isRequired,
```
jm90m: should we start defining the properties for certain shapes now for propTypes?

src/client/components/SocialLinks.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import SocialLink from './SocialLink';
+
+export const socialProfiles = [
```
jm90m: might be good a idea to keep `socialProfiles`, `socialTransformers` and `transform` in a helper / constants file instead of a components file

src/client/components/SocialLinks.js
```diff
@@ -0,0 +1,53 @@
...
+  { id: 'facebook', icon: 'facebook', color: '#3b5998', name: 'Facebook' },
```
jm90m: So I noticed we have a lot of colors floating around in our codebase in our javascript files...
would be nice to be keep them in a `constants/styles.js` file and import it based on below see jist here:

https://gist.github.com/jm90m/933417aaf8f1259804957be4df090a91

so `import { COLORS } ` then you get the color from there, i figure out the color name by using this website http://www.color-blindness.com/color-name-hue/

src/client/components/SocialLink.js
Sekhmet: If we know shape ahead of time, and it doesn't require much work to define properties it might be useful.

src/client/components/SocialLinks.js
Sekhmet: I don't think we have many colour definitions in our JS files. Those are only useful in context of social profiles.
Things is we already have file with colours (less) and I don't think we can combine those two.

PR #1290, opened 2018-01-01
Unify style of Wallet and Activity

### LINKED ISSUE #1289: Double border on activity row
There is a double border on activity row see: 
image
https://busy.org/@sekhmet/activity

Sekhmet:
Fixes #1289 

Changes:
- Use same style for transactions and activities.


Session.vim
```diff
@@ -0,0 +1,60 @@
+let SessionLoad = 1
```
bonustrack: Are you sure about this file? 😄

Sekhmet: 😅

PR #1303, opened 2018-01-05
Update Topnav and tooltip z-index

### LINKED ISSUE #1293: Post PopoverMenu overflow at NavigationBar
3xwkdw7xqx

#### Expected behavior

The **Popover** should appear behind the NavigationBar.

#### Actual behavior

The **Popover** appear in front of NavigationBar.

* Browser: chrome
* Operating system: win7

Sekhmet:
Fixes #1293 

Changes:
- Set Topnav z-index to 1050 (popover has z-index of 1030)
- Set tooltip z-index to 1060.

### DIFF
--- a/src/client/Wrapper.js
+++ b/src/client/Wrapper.js
@@ -167,7 +167,7 @@ export default class Wrapper extends React.PureComponent {
       <IntlProvider key={usedLocale} locale={usedLocale} messages={translations}>
         <LocaleProvider locale={enUS}>
           <Layout data-dir={getLocaleDirection(getAvailableLocale(locale))}>
-            <Layout.Header style={{ position: 'fixed', width: '100%', zIndex: 5 }}>
+            <Layout.Header style={{ position: 'fixed', width: '100%', zIndex: 1050 }}>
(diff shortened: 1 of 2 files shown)


bonustrack:
This make the header visible when a modal is open:
image

bonustrack:
The top nav should be under the black half transparent background of the modal aswell
image

jm90m:
@bonustrack @Sekhmet So I took a deep look into this issue and examined a few big sites.

I don't think that this is a real issue: https://github.com/busyorg/busy/issues/1293

If you go on https://medium.com/, and then look at their popover menu, the popover menu should always be behind the topnav..

screen shot 2018-01-06 at 7 52 58 am

If you open the popover menu to any post then scroll down you'll notice that topnav is always above it. Also setting z-index to things like this can overcomplicate the app, we should let antd handle the z-index, since it knows the right way of handling which elements go on top of each one.

Sekhmet:
@bonustrack what's your take on this?

bonustrack:
I don't like to see this top header over the grey background when opening a modal:
image
You can put the top nav over the popover menus but don't change the behavior when opening modal.

Sekhmet:
@bonustrack I updated those.

PR #1311, opened 2018-01-05
Add links to user profile in comments

### LINKED ISSUE #1308: Avatar on comments should link to user profile
Now the avatar on comments don't have link, there should be a link to the user profile
image

Sekhmet:
Fixes #1308 

Changes:
- Add link to user profile.

### DIFF
--- a/src/client/components/Comments/Comment.js
+++ b/src/client/components/Comments/Comment.js
@@ -222,6 +222,8 @@ class Comment extends React.Component {
       );
     }
 
+    const avatarSize = comment.depth === 1 ? 40 : 32;
+
     return (
       <div ref={this.setSelf} className="Comment" id={anchorId}>
         <span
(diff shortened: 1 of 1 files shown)


src/client/components/Comments/Comment.js
```diff
@@ -235,7 +237,9 @@ class Comment extends React.Component {
             <i className="iconfont icon-offline" />
           )}
         </span>
-        <Avatar username={comment.author} size={comment.depth === 1 ? 40 : 32} />
+        <Link to={`/@${comment.author}`} style={{ height: avatarSize }}>
```
Sekhmet: I believe there is better way to prevent this link from taking entire height than that.
@jm90m

jm90m: you did the right thing here with setting the height of this, there is no way to disable height with flex box, it will always be 100% of the container, unless you change the flex box direction to column in the parent container but I don't think you would want that in this situation. what you can do is set `max-height` to `avatarSize` as well but its essentially the same thing as setting `height` to `avatarSize` when dealing with flex box.

PR #1313, opened 2018-01-06
Make showNSFWPosts default to false

### LINKED ISSUE #1239: Handle showNSFWPosts properly
Currently unchecked `Show NSFW posts` option saves to null instead of false, so it causes prop types checks to fail.

1. It should be initialised to false by default instead of null:
https://github.com/busyorg/busy/blob/49a21a91faeb399cb0dfa8d9309e5c9cf1f989bc/src/client/settings/Settings.js#L68-L70
2. We should initialise it in `componentWillMount` as well:
https://github.com/busyorg/busy/blob/49a21a91faeb399cb0dfa8d9309e5c9cf1f989bc/src/client/settings/Settings.js#L72-L78

Sekhmet:
Fixes #1239 

Changes:
- Default `showNSFWPosts` to `false` instead of `null`.
- Set `showNSFWPosts` on mount.

### DIFF
--- a/src/client/settings/Settings.js
+++ b/src/client/settings/Settings.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
-import _ from 'lodash';
 import { injectIntl, FormattedMessage } from 'react-intl';
 import { Select, Radio, Checkbox } from 'antd';
 import {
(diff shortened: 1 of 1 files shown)


jm90m:
i forgot why, but the reason it was initially null was because of this line https://github.com/busyorg/busy/blob/e13009605c92e54a2b26887b7fe4eb90e3f63553/src/client/settings/Settings.js#L290

I'll have to look and try to remember why I had set it to null instead of false

PR #1314, opened 2018-01-06
add Scrollbars to CommentFooter

### LINKED ISSUE #1036: Edit option below a comment is not fully visible on mobile
#### Expected behavior

All elements and text in the line below a comment should be fully visible.

screenshot_20171107-070343

#### Actual behavior

The edit word in Czech "Upravit" is partly cut off if the comment is not at the top level of comment conversation.

screenshot_20171107-070331

#### How to reproduce

1. Set language to Czech
2. Find your comment that is not at the top level

* Browser: Brave
* Operating system: Android 7.1.2

sweever:
Fixes #1036

In my last PR it seem that breaking comment actions in two line take alot of space , so in this one i suggest this changement by adding antd scrollbars to comment actions . 
As the issue is in the mobile view , we can use this feature temporarily untill we have a better solution .

- Changes on ; busy/src/client/components/CommentFooter/CommentFooter.js
### Adding scrollbars module & implent it.
- Changes on ; busy/src/client/components/CommentFooter/CommentFooter.lss
### force full width to commentFooter

### DIFF
--- a/src/client/components/CommentFooter/CommentFooter.js
+++ b/src/client/components/CommentFooter/CommentFooter.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import find from 'lodash/find';
+import { Scrollbars } from 'react-custom-scrollbars';
 import { getHasDefaultSlider, getVoteValue } from '../../helpers/user';
 import Slider from '../Slider/Slider';
 import Buttons from './Buttons';
--- a/src/client/components/CommentFooter/CommentFooter.js
+++ b/src/client/components/CommentFooter/CommentFooter.js
@@ -143,16 +143,18 @@ export default class CommentFooter extends React.Component {
     }
 
     return (
-      <div className="CommentFooter">
-        {actionPanel}
-        {sliderVisible && (
-          <Slider
-            value={this.state.sliderValue}
-            voteWorth={this.state.voteWorth}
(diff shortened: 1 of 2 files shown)


jm90m:
hey thanks for the contribution, I only see the scrollbar when I actually hover over the comment footer and then try to scroll left to right, I dont think our users will understand this when they are on mobile and there isn't really a hover effect.

screen shot 2018-01-08 at 12 55 57 am

sweever:
@jm90m 

> I only see the scrollbar when I actually hover over the comment footer and then try to scroll left to right

#### yes true because you need to test it from mobile .

> I dont think our users will understand this when they are on mobile and there isn't really a hover effect.

#### what about this ??

jm90m:
yeah something like that would work but I'm using chrome simulator for iphone 5 (which is one of smallest screens and scrollbar doesn't show up)
screen shot 2018-01-08 at 4 48 07 pm

bonustrack:
I feel it's not so obvious for the user that this menu is scrollable. Wouln't be better to use icon instead of text for "Reply" and "Edit" ?

bonustrack:
What's your through @pareshUD?

pareshUD:
@bonustrack How about if we give 3 dots and on that allow user to edit delete.. Something facebook does on mobile site.
img_0888

bonustrack:
Closing this PR we will fix that issue in another way.

PR #1317, opened 2018-01-06
Refactor Editor and CommentForms

### LINKED ISSUE #913: [Proposal] Consolidate Editor Components

### LINKED ISSUE #946: Potentially Incorrect Usage of onChange with antd's Form
Upon reading AntDesign's documentation for using `Form`, I found the following note:
> After wrapped by getFieldDecorator, value(or other property defined by valuePropName) onChange(or other property defined by trigger) props will be added to form controls，the flow of form data will be handled by Form which will cause:
> 1. **You shouldn't use onChange to collect data**, but you still can listen to onChange(and so on) events.

Just to clarify, our usage of `Form`:
- Uses `getFieldDecorator`
- Uses `onChange` to collect data

I don't know the full history of this comment in `Editor.js`:
> // NOTE: antd doesn't update field value on Select before firing onChange
    // so we have to get value from event.

So I'm not clear if the author was aware of the note in the documentation or not. If this function was implemented with the knowledge of that documentation, then feel free to close the issue.

However, for the sake of this issue I'm going to operate under the assumption that it wasn't. If that's the case, it would explain why we had issues with collecting data from `onChange`.

In fact, in my fix for #924, I found that validators aren't invoked until *after* `onChange` has run, and when I triggered the validators from within `onChange`, I found that they were running against the previous data - i.e. I had just entered 6 tags, and the validator ran on the previous input which contained 5 tags, hence it did not fail.

So going with the assumption that we shouldn't be using `onChange` in this way, we should switch from using `onChange` to using their recommended function `onFieldsChange`. In theory, this should free us from our workaround functions like `onUpdate` and `getValues`.

If I'm completely wrong about this issue, please let me know what the story is and why we should in fact be using `onChange`.

Sekhmet:
Fixes #913 

Changes:
- [x] Add EditorInput
- [x] Add withEditor
- [x] Update Editor
- [x] Update CommentForm
- [x] Update EmbeddedCommentForm
- [x] Expose input as ref to allow focusing.

### DIFF
--- a/src/client/components/Comments/CommentForm.js
+++ b/src/client/components/Comments/CommentForm.js
@@ -16,6 +16,7 @@ class CommentForm extends React.Component {
   static propTypes = {
     parentPost: PropTypes.shape().isRequired,
     username: PropTypes.string.isRequired,
+    top: PropTypes.bool,
     isSmall: PropTypes.bool,
     isLoading: PropTypes.bool,
     submitted: PropTypes.bool,
@@ -26,7 +27,7 @@ class CommentForm extends React.Component {
   };
 
   static defaultProps = {
-    username: undefined,
+    top: false,
     isSmall: false,
     isLoading: false,
     submitted: false,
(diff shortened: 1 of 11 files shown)


bonustrack:
Copy paste an image does not seem to be working anymore. Can you check that @Sekhmet ?

bonustrack:
The textarea feel too too big now, can you use the same height than before?
image
image

jm90m:
Been going through the code, all looks fine to me, I was testing it a few days ago, and everything was also working fine 👍 , but for some reason now I can't login, anyways I guess once @bonustrack is done with his style inputs its good to merge

Sekhmet:
@bonustrack not sure if you have seen, but size has been updated already.

PR #1320, opened 2018-01-07
Generate accurate metadata and refactor image lightbox

### LINKED ISSUE #829: Gallery not working with images with class pull-left / pull-right
#### Expected behavior

Images within the div with the classes should show in the gallery.

#### Actual behavior

Images with the classes don't show in the gallery at all. They cannot be shown even though going left or right in the gallery.

#### How to reproduce

1. Go to `/cesky/@espoem/co-kdyby-byla-afrika-jednim-statem`
2. See the images next to each other. (They have the classes).
3. Click on them, the gallery shows.
4. None of the images are in the gallery.

* Browser: Chrome

### LINKED ISSUE #848: Don't open gallery on images wrapped in <a href></a>
#### Expected behavior

Gallery should not open when the image is wrapped in link tag. `<a href=""></a>`

#### Actual behavior

The image is actual link so a new tab opens. It can be any link but presumably it directs to the image origin. The gallery also opens with enlarged image.

#### How to reproduce

1. Go to `/design/@andrejcibik/my-graphic-design-litecoin-wallpaper-4k-res`
2. Click the image

* Browser: Chrome

### LINKED ISSUE #1366: Post images missing on json_metadata image
Some posts from busy.org don't have their images saved on `json_metadata.image` for example on this post, images urls are stored in `links`:
https://steemdb.com/steem/@jerrybanfield/steem-blockchain-frees-aggroed-from-the-corporate-world-1/data
Some apps don't fetch the whole body post from feeds and dont know the preview image if the url is not inside the `json_metadata.image`.

Sekhmet:
Fixes #848 
Fixes #829 
Fixes #1366

Changes:
- Don't open images that are direct child of link.
- When generating post metadata use regex to extract links and images (current solution ignores HTML links and images).
- Use regex to extract images from post instead of using metadata as it might be incomplete (e.g. this post is missing images in metadata so lightbox won't open: https://staging.busy.org/piston/@xeroc/piston-web-first-open-source-steem-gui---searching-for-alpha-testers)

### DIFF
--- a/src/client/helpers/parser.js
+++ b/src/client/helpers/parser.js
@@ -7,4 +7,27 @@ export function getFromMetadata(jsonMetadata, key) {
   return _.get(metadata, key);
 }
 
+const imgRegex = /<img[^>]+src="([^">]+)"/g;
+const hrefRegex = /<a[^>]+href="([^">]+)"/g;
+
+function extract(body, regex) {
+  const matches = [];
+
+  let match;
+  do {
+    match = regex.exec(body);
+    if (match) matches.push(match[1]);
+  } while (match);
+
+  return matches;
+}
+
+export function extractImages(body) {
+  return extract(body, imgRegex);
+}
+
+export function extractLinks(body) {
+  return extract(body, hrefRegex);
+}
+
 export default null;
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
     "lightrpc": "0.0.1",
     "lodash": "^4.0.0",
     "lodash-webpack-plugin": "^0.11.4",
-    "marked": "^0.3.6",
     "matchmedia": "^0.1.2",
     "morgan": "^1.8.2",
     "newrelic": "^1.40.0",
--- a/src/client/components/Story/StoryFull.js
+++ b/src/client/components/Story/StoryFull.js
@@ -18,7 +18,7 @@ import { getFromMetadata, extractImages } from '../../helpers/parser';
 import { isPostDeleted } from '../../helpers/postHelpers';
 import withAuthActions from '../../auth/withAuthActions';
 import { getProxyImageURL } from '../../helpers/image';
-import Body, { remarkable } from './Body';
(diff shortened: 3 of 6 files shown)


bonustrack:
Be careful when you introduce a new library to check the weight of the lib. `cheerio` seem to use `parse5` which seem to be heavy:
image

Before:
https://busy-master-pr-1320.herokuapp.com/js/statistics.html
After:
https://staging.busy.org/js/statistics.html

Do you have an alternative?

Sekhmet:
I've used regex instead of cheerio.
It decreases size of vendor bundle, but it might not be [the best option](https://stackoverflow.com/a/1732454/1780581) in the long run.

bonustrack:
Looking good, is it possible to have the cursor pointer when mouse over image same before?

src/client/helpers/parser.js
```diff
@@ -7,4 +7,27 @@ export function getFromMetadata(jsonMetadata, key) {
   return _.get(metadata, key);
 }
 
+const imgRegex = /<img[^>]+src="([^">]+)"/g;
```
jm90m: im not a regex expert but I think this matched this:

screen shot 2018-01-15 at 1 50 29 am

but I guess this would be a rare case, all code looks fine to me, not too sure on regex but if @bonustrack is fine with this then lgtm

Sekhmet: That's drawback of parsing HTML using Regexes. There is no simple regex that would handle all edge cases, it works as long as it's proper HTML. It breaks when body is just random content.

This would extract `src` properly though. I don't think it's worth it trying to make standard HTML parser using regex that handles non-standard "HTML".

jm90m: yeah makes sense, I've been trying to figure out ways on the side too, to try to extract html tags for other projects, its hard to do with just regex

Sekhmet:
@bonustrack I made extracting more accurate and added pointer cursor to images.

PR #1321, opened 2018-01-07
Changed instructions below 'topics' input box to make more accurate

debrayc:
I attempted to update the instructions ("Separate topics with spaces. Only lowercase letters, numbers and hyphen characters are permitted.") below the topics input box on the '/editor' route and found that the file change that I initially thought would fix the issue did not have the desired effect. 

I then searched the repo for the phrase only to find that there were 26 other instances of the phrase... I then replaced all instances with "Separate topics with spaces. Only lowercase letters, numbers and hyphen characters are permitted." I then saved the files and the text rendered on the screen as expected.

Here is a gif screen-capture from busy.org/editor:

_As you can see, commas ARE NOT NECESSARY!_

and here is a screenshot of the updated app running on my local machine:

Thank you for all that you guys do... Hope this is helpful...


bonustrack:
Thank you for your contribution @debrayc. Can you remove the file package-lock.json from you PR? We don't use it we use yarn.lock instead.

debrayc:
@Sekhmet I deleted the package-lock.json as requested. Let me know if you want me to change the other files as well, and thank you for all that you do!

Sekhmet:
The only translation file you should update is `default.json`. Can you revert changes to other files?

debrayc:
Will do and have it updated by tonight.

debrayc:
@Sekhmet it is done...

PR #1339, opened 2018-01-15
Update Design on Post Recommendations

### LINKED ISSUE #1252: Add image preview on recommended posts
It would be nice to have an image preview on the block "Suggested posts"
image
Can you work on a design for it @pareshUD ?

jm90m:
Fixes #1252

Changes:
- Update styles and layout of post links based on new design from https://github.com/busyorg/busy/issues/1252
- Display image preview

screen shot 2018-01-14 at 9 56 22 pm
screen shot 2018-01-14 at 9 56 26 pm


bonustrack:
- Now the image is stretched, it should keep aspect ratio. .
- This should not be clickable:
image
- Title should have bigger font size (16px)
image

bonustrack:
On mouse over comment link, the purple is ligther than the one we use see:
image
- Can you make it same color?
- Also remove the purple color on mouse over the title

bonustrack:
@pareshUD can you test recommended post here: https://busy-master-pr-1339.herokuapp.com/travel/@athome/oregon-coast what you think?

jm90m:
@bonustrack The purple color on hover is the default color for `hover` on all `a` / link tags, do you want me to change that for both comments & title? Comments will use the `active` link color rather than default link color for hover? And Title has no color?

screen shot 2018-01-18 at 4 59 14 pm

bonustrack:
- Color 1 and 2 are not same
image
- The title dont need color on mouse over (similar than a post title on a feed)

bonustrack:
The preview image should not use the same width/height than on the feed. Images are too big to load, the width/height should be like 40x40, not 600x800.

PR #1340, opened 2018-01-15
Update Design on Invites Page

### LINKED ISSUE #1216: Improve invite page design

jm90m:
Fixes #1216

Changes
- Add Share Buttons
- Complete Redesign of Invite Page based on design from https://github.com/busyorg/busy/issues/1216

screen shot 2018-01-15 at 1 28 05 am


bonustrack:
We will need to have a better text for invitation through email / twitter / facebook something like:
`Join me today on busy.org and get rewarded to blog {link}`
Also we need the original image for the letter icon with SVG, please contact @pareshUD to get the file.

jm90m:
@bonustrack Facebook doesn't let you share with text only URL using their share button - https://developers.facebook.com/docs/plugins/share-button, but for twitter / email, we can do. I can also add other social media like reddit, etc.. I used this to grab the SVG's, of the site icons so whatever is available on here - http://sharingbuttons.io/

bonustrack:
The social media icons should be bundled into iconfont aswell.

bonustrack:
@jm90m Can you change the title `Invite & Earn Rewards` to `Don't use Busy alone!`. FYI we should never use: "This Case On Title" but "This case on title". Icon is working for me now :)

PR #1348, opened 2018-01-17
Fix for 'User is not Authenticated' promise error

### LINKED ISSUE #1310: "User is not authenticated" error in the console for unlogged user
There is an error: "User is not authenticated" in the console when opening https://staging.busy.org/ as unlogged user
image

jm90m:
Fixes #1310

Changes
- Add check for successful login within login action call


bonustrack:
@jm90m We should not send any login request if the user doesnt have an `access_token` saved on his cookies

src/client/Wrapper.js
```diff
@@ -89,7 +88,10 @@ export default class Wrapper extends React.PureComponent {
   componentDidMount() {
     const { loaded, locale, usedLocale } = this.props;
 
-    this.props.login().then(() => this.props.getFollowing());
+    if (Cookie.get('access_token')) {
+      this.props.login().then(() => this.props.getFollowing());
```
Sekhmet: It was called for side-effects.
When login succeeds/fails we set `loaded` property in `authReducer`. This is used for verifying if we have checked if user is logged in or not.
Not calling `login()` when user has no cookie, causes that `loaded` will never be set potentially causing some functionality that relies on that to break (I'm not sure if it did, but it could potentially happen).
I though that this condition would be true if cookie is not set, but it turns out it doesn't.
https://github.com/busyorg/busy/blob/0d078e451ffecb61b12ab8a1cbf4fac999de0c8c/src/client/auth/authActions.js#L29

I think that more elegant way would be to just dispatch `GET_FOLLOWING_ERROR` action instead of throwing there:
https://github.com/busyorg/busy/blob/0d078e451ffecb61b12ab8a1cbf4fac999de0c8c/src/client/user/userActions.js#L118

Sekhmet: I noticed that it indeed causes some problems. One of them is that HeroBanner won't show up on dev-server because it can't detect if authentication has been completed.

Issue #1352, opened 2018-01-18
Some post are breaking the layout


bonustrack:
![image](https://user-images.githubusercontent.com/16245250/35098368-6c544bf0-fc86-11e7-9ae8-c471edfd4d45.png)
It seem related to this post: https://steemdb.com/bitshares/@heiditravels/blockchain-generations-from-bitcoin-to-smart-contracts-and-graphene/data
As the console lead to an error for this post on the var `window.__PRELOADED_STATE__ `
![image](https://user-images.githubusercontent.com/16245250/35098460-d1b09990-fc86-11e7-8cf5-94ad4e22d069.png)

gregory-latinier:
https://api.steemjs.com/getContent?author=heiditravels&permlink=blockchain-generations-from-bitcoin-to-smart-contracts-and-graphene

Using the data directly from the API we can see that a special character `LS` is present twice in the body post. This breaks the JSON.

jm90m:
Doesn't seem to be the problem with the post, since it works fine locally, might have to do something with a combination of bundling up process + react versions + returning null on components render - see here, https://github.com/busyorg/busy/pull/1357, also you can check on `busy-dev` discord channel for mine & @Sekhmet conversations on this matter

bonustrack:
@jm90m On localhost you need to run production code to have SSR and reproduce the issue. This fix #1353 was fixing this issue, the post of `heiditravels` has `LS` for line separator and was broking the variable.

jm90m:
I couldn't get #1353 to work for me... I think it should be like what I have in here:

https://github.com/busyorg/busy/blob/6d6c49ec0e8ad426a9bf9b0f79871656f16b31b8/src/server/renderers/ssrRenderer.js#L18

rather than what's in #1353 here: 
https://github.com/busyorg/busy/blob/6c3b3456bc5bea2b558ea5719ab747254cd08039/src/server/renderers/ssrRenderer.js#L15

jm90m:
I also don't think it should be replaced with `'\n'` it should be `'\\n'`

PR #1354, opened 2018-01-18
Replace the character that make the SSR crash

bonustrack:
Add fix #1353 to master for #1352. Do you have a better solution @Sekhmet ?


bonustrack:
It seem that this change break image preview on prod.

bonustrack:
Closing this since we have a better fix here #1360

PR #1357, opened 2018-01-19
Return div on hero banner instead of null

jm90m:
Return div on hero banner instead of null, as discussed with @Sekhmet on discord.

This should fix unlogged out user issue


bonustrack:
I'm getting logged, unlogged then logged again when opening https://busy-master-pr-1357.herokuapp.com/trending/bitshares
SSR is broken.
test

bonustrack:
We need to solve the root cause wich seem to be related to #1353

jm90m:
Closing this, resolved in #1360

PR #1361, opened 2018-01-20
Add click handler for Popovermenu on PostFooter for single page posts

jm90m:
Fixes:
- Single page post popover menu is not working
Changes
- Add click handler for Popovermenu on PostFooter for single page posts


bonustrack:
The redirection is working fine but we use to have a loading when click on "Edit post"
image
Now when i click on it, i dont have indication that it's loading.

PR #1363, opened 2018-01-20
Add error check incase asksteem fails for search

jm90m:
Fixes # 
Sometimes AskSteem is down, causing search to completely fail and not rely on steem for other results. See here https://staging.busy.org/search?q=jm90mm vs this PR

Changes:
- Add catch block for askSteem to return an empty array rather than throw an error


bonustrack:
When i open the search page i will see this for few milisecond while the search is not loaded yet:
image
This message should only appear when the search is done and there is no result, not before. Is it possible to solve this too on this PR?

jm90m:
yeah should be fixed in this PR

PR #1367, opened 2018-01-22
Add Check for empty content before making request to grab feed data

### LINKED ISSUE #1260: Busy.org Bug: Previous Page Jumping Issue
https://steemitimages.com/0x0/https://res.cloudinary.com/hpiynhbhq/image/upload/v1512514647/d2h2xi86qzqdwsuubfgp.gif

This is a bug report for Busy.org. It covers an issue with the __previous page__ function of the browser. The same bug appeared on Utopian.io (Busy fork) but was fixed with a recent update. 

## Issue
<hr>

When browsing through a category or trending/hot page and then clicking on a post to read it one of the common things to do afterwards is to return to the previous page via the according browser function. However, on Busy.org it seems to lock to a certain post and disregards the previous position within that list of posts. You can see a quick jump happening where the browser jumps away from the correct one to the one that is locked. 

For better understanding I made a video about that behaviour. As you can see it doesn't happen after the first post but after the second as it seems. I was able to reproduce that exact behaviour a few times.

https://youtu.be/wWishEAuht0

For comparison you can try the same on Utopian.io and see that it returns there to the correct post in the list. 

_This was tested on a Windows 10 PC with Firefox 57.0.2_

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@flauwy/busy-org-bug-previous-page-jumping-issue">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

jm90m:
Fixes #1260

Changes:
- add `_.isEmpty` check for contents before making request to grab feed data


bonustrack:
When i open https://busy-master-pr-1367.herokuapp.com/trending/photography and scroll on the page a bit while the topic are still loading in left, when the page get totaly loaded the scroll get reset and i comeback in top of the page. This behavior not seem to happen in staging and can be annoying.

jm90m:
@bonustrack  I can reproduce it on staging:

https://media.giphy.com/media/l0HUcFLQRzSqpuWA0/giphy.gif

jm90m:
@bonustrack can ya check this one? I think I found the issue a few days ago

PR #1370, opened 2018-01-23
Fix for page not fully visible on load

### LINKED ISSUE #1365: Page is not fully visible on loading
When the page is loading i'm not able to see it fully, the content stop at the red line mark:
image
The overflow should not be hidden.

jm90m:
Fixes #1365

Changes:
- Add minimum height to `container` class
- Remove `container` class from topnav and add previous container styles in `topnav-layout`


bonustrack:
Layout look broken to me:
image

jm90m:
should be fixed now @bonustrack, looks like I missed one place that still had `.container` styles and was in the topnav area. `.container` should be used for contents in the page

PR #1378, opened 2018-01-26
Set saving drafts to a few seconds

jm90m:
Fixes:
screen shot 2018-01-29 at 10 11 41 pm

Changes:
- Update saving drafts to 2 seconds


src/client/post/Write/Write.js
```diff
@@ -256,7 +256,7 @@ class Write extends React.Component {
     const redirect = id !== this.draftId;
 
     this.props.saveDraft({ postData: data, id: this.draftId }, redirect);
-  }, 400);
+  }, 2000);
```
Sekhmet: I don't think this is a good solution. It's just changing numbers, in few days it might still cause problem, so we will increase it even more. At some point we will notice that it takes too long and we will decrease it again.
We can use flag that tells us we we are saving post or not, no need to come up with magic numbers.

jm90m: yeah youre right, ill update it to use a flag

jm90m:
@Sekhmet just updated to check for Editor `saving` prop. Based on conversations with @gregory-latinier, I'm thinking of keeping the debounce at 2000 ms. Just to be extra safe.

jm90m:
@Sekhmet I also added screenshot of the problem in description of this PR, since I'm lazy :)

PR #1386, opened 2018-01-27
Add YouTube social profile

### LINKED ISSUE #1385: Add youtube field on edit profile page
We missed that one :)

Must be a value like `UCta3TYFhzfpPvOtKBDifYJA`, not the full youtube profile url. There should be a validator for it.

Sekhmet:
Fixes #1385 

Changes:
- Add YouTube entry to social profiles.

### DIFF
--- a/src/client/helpers/socialProfiles.js
+++ b/src/client/helpers/socialProfiles.js
@@ -1,6 +1,7 @@
 const socialTransformers = {
   facebook: id => `https://facebook.com/${id}`,
   twitter: id => `https://twitter.com/${id}`,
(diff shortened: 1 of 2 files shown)


verymilan:
but isn't `/user/` out of date? 😕 
i found no way to get such an id for youtube, instead it is `/channel/` and `/c/` (at least) now as it seems (as there are variants, i felt insecure otherwise i'd simply update the entry and made a pullrequest)

Sekhmet:
Now YouTube uses two different URLs that are not compatible with each other: https://github.com/busyorg/busy/issues/1489
To solve this we would have to allow for full URLs.

PR #1390, opened 2018-01-28
Always call login for side-effects

### LINKED ISSUE #1387: Restore side-effects on login action
#### Expected behavior

`login` action should always be called for side-effects.

#### Actual behavior

`login` is not called when cookie is not present. `loaded` is never set, so some functionality is not working properly (for example HeroBanner won't open on dev-server or potentially other problems.

See https://github.com/busyorg/busy/pull/1348#pullrequestreview-92022131 for explaination

#### How to reproduce

N/A

Sekhmet:
Fixes #1387 

Changes:
- Call `login` always for side-effects.
- Dispatch `GET_FOLLOWING_ERROR` instead of throwing.

### DIFF
--- a/src/client/Wrapper.js
+++ b/src/client/Wrapper.js
@@ -6,7 +6,6 @@ import { withRouter } from 'react-router-dom';
 import { renderRoutes } from 'react-router-config';
 import { LocaleProvider, Layout } from 'antd';
 import enUS from 'antd/lib/locale-provider/en_US';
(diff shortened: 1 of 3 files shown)


bonustrack:
Looking good, i dont see `/api/me` request without cookie. When i have a valid cookie, the access_token is verified on server side and when i open a page i see myself already logged but there is still a request to `/api/me`. This request is not necessary.

PR #1392, opened 2018-01-28
Add Busy beta opt-in

### LINKED ISSUE #1253: Add setting to be beta tester
We can add a setting to enable user to beta test the last version of busy (staging.busy.org). If user active it, he will get redirected to staging.busy.org automatically if he try to go on busy.org (until he disable it).

Sekhmet:
Fixes #1253 

Changes:
- Add beta checkbox.
- Redirect client-side.
- Redirect server-side.

### DIFF
--- a/src/client/settings/__tests__/settingsReducer.js
+++ b/src/client/settings/__tests__/settingsReducer.js
@@ -89,10 +89,7 @@ describe('settingsReducer', () => {
       },
     };
 
-    const res = settingsReducer(stateBefore, action);
-    console.log(res);
-
-    expect(res).to.eql(stateAfter);
+    expect(settingsReducer(stateBefore, action)).to.eql(stateAfter);
   });
 
   it('should set locale and voting power after login success', () => {
--- a/src/client/settings/__tests__/settingsReducer.js
+++ b/src/client/settings/__tests__/settingsReducer.js
@@ -14,6 +14,7 @@ describe('settingsReducer', () => {
     rewriteLinks: false,
     upvoteSetting: true,
     rewardSetting: rewardsValues.half,
+    useBeta: false,
   };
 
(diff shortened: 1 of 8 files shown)


bonustrack:
I have a small issue, when i open the page setting from a fresh tab, i will see the page loaded for few milisecond then it disappear and loading appear. The setting page should never be visible before been loaded. See: 

bug

Sekhmet:
This bug is not related to that PR (it happens on busy.org as well).

src/client/settings/settingsReducer.js
```diff
@@ -53,6 +56,7 @@ const settings = (state = initialState, action) => {
         showNSFWPosts: action.payload.showNSFWPosts,
         upvoteSetting: action.payload.upvoteSetting,
         rewardSetting: action.payload.rewardSetting,
+        useBeta: !!action.payload.useBeta,
```
jm90m: isn't `useBeta` always a boolean? so you won't need the `!!`

src/client/settings/__tests__/settingsReducer.js
```diff
@@ -85,7 +86,10 @@ describe('settingsReducer', () => {
       },
     };
 
-    expect(settingsReducer(stateBefore, action)).to.eql(stateAfter);
+    const res = settingsReducer(stateBefore, action);
+    console.log(res);
```
jm90m: random `console.log`?

src/client/settings/settingsReducer.js
Sekhmet: It comes from SC2 so it might be `undefined` as well. This way it won't safe as `undefined` in state.

jm90m: ahh I see, makes sense

PR #1399, opened 2018-01-29
Fetch followers/following on demand

### LINKED ISSUE #1389: Bug with center block height
The height of the center block is bigger than expected.
image
Test here: https://busy.org/@faberium/followed

### LINKED ISSUE #1397: Fetch followers/following on demand
#### Expected behavior

Followers/following should be fetched on demand as user scrolls down the list.

#### Actual behavior

Followers/following are fetched all at once and then they are loaded from memory in batches. 

#### How to reproduce

Visit profile with many followers: https://busy.org/@sweetsssj/followers
It takes a while (~30 seconds) to load all followers before they are displayed.

Sekhmet:
Fixes #1397 
Fixes #1389

Changes:
- Add `UserDynamicList` that handles dynamic loading.
- Use `UserDynamicList` in `UserFollowers` and `UserFollowing`.


src/client/user/UserDynamicList.less
```diff
@@ -1,11 +1,16 @@
 @import (reference) '../styles/custom.less';
 
-.UserFollowers {
+.UserDynamicList {
   background: @white;
   border: 1px solid #e9e7e7;
```
jm90m: Color here is `@white-gainsboro`

src/client/user/UserDynamicList.js
```diff
@@ -0,0 +1,63 @@
...
+        );
+      },
+    );
+  }
+
+  render() {
+    const { loading, hasMore, users } = this.state;
+    return (
+      <div className="UserDynamicList">
+        <div className="container UserDynamicList__container">
```
jm90m: You can remove `container` here, you don't need it, and this will fix (https://github.com/busyorg/busy/issues/1389)

Which I have a PR for, but you can fix it here.

jm90m:
You might want to add a check for when a user doesn't have any followers/following, because right now it just renders white space. I think it might be good to add the check in the render function of `UserDynamicList` component and just add a class to `UserDynamicList__container` like `UserDynamicList__empty` and in the styles make that `display: none`

Sekhmet:
I changed `@medium` to `@small` as it wasn't rendering borders on two column layout:
obraz

PR #1400, opened 2018-01-29
Detect errors based on their content

### LINKED ISSUE #1328: Bug Hunting: An Error of Warning and Post Size of busy.org
### The Issue

Days ago I [reported](https://steemit.com/utopian-io/@dapeng/bug-hunting-another-error-of-steem-post-size) a bug of steemit.com that an article with only about 400 lines cannot be posted on steemit.com, unless I reduced it to 150 lines. Yesterday I tested the same article on busy.org and found an error.

### The behavior of Busy.org

When I pasted this article to the default editor interface of busy.org and clicked  the post button, the button displayed 'submitting' for seconds,

Then a red cross appeared on the top of the editor:

The red cross was there for seconds, and disappeared. The post failed to post. Nothing else happened.

I removed some lines and tested many times, but the same error message appeared again and again. Finally, I posted the first 150 lines successfully.

### Expected behavior

busy.org should show a readable and understandable warning if there is something wrong with the post. Only a red cross is confusing, and the author does not  know what to do. If the problem is the text size limit, as I mentioned in the bug report of steemit.com, then I would suggest that  busy.org should check the size limit when the user types something in the default editor, or show a online counter displaying how many characters the user can type. 

### How to reproduce

Copy the text [here](https://raw.githubusercontent.com/pzhaonet/keller/master/mr/hope_bug.txt) and paste to the default online editor of busy.org, and click the post button.

### Environment

Windows 7 + Firefox 56

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@dapeng/bug-hunting-an-error-of-warning-and-post-size">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

Sekhmet:
Fixes #1328 

Error messages returned from APIs are inconsistent so they cannot be parsed reliably. This PR detects common error messages based on some kind of fingerprint and returns relevant error message.

Changes:
- Return error messages based on fingerprint.

### DIFF
--- a/src/client/errorMiddleware.js
+++ b/src/client/errorMiddleware.js
@@ -1,17 +1,25 @@
 import { notify } from './app/Notification/notificationActions';
 
 function parseBlockChainError(error) {
-  let parsedError;
-  try {
-    parsedError = error
-      .split(':')[2]
-      .split('\n')[0]
-      .substring(1);
-  } catch (e) {
-    parsedError = 'Unknown error';
+  if (error.indexOf('You may only post once every 5 minutes') !== -1) {
+    return 'You may only post once every 5 minutes.';
+  } else if (error.indexOf('You may only comment once every 20 seconds') !== -1) {
+    return 'You may only comment once every 20 seconds.';
+  } else if (error.indexOf('You have already voted in a similar way') !== -1) {
+    return 'You have already voted in a similar way.';
+  } else if (error.indexOf('Account has already reblogged this post') !== -1) {
+    return 'You have already reblogged this post.';
+  } else if (
+    error.indexOf('<= (get_dynamic_global_properties().maximum_block_size - 256)') !== -1
+  ) {
+    return 'Your post is too big.';
+  } else if (error.indexOf('bandwidth limit exceeded') !== -1) {
+    return 'Your bandwith has been exceeded. Please wait to transact or power up STEEM.';
   }
 
-  return parsedError;
+  // Log error to console for further investigation.
+  console.log('Unknown error', error);
+  return 'Unkown error has occured.';
 }
 
 export default function errorMiddleware({ dispatch }) {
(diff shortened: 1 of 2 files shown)


src/client/errorMiddleware.js
```diff
@@ -1,17 +1,25 @@
...
 function parseBlockChainError(error) {
-  let parsedError;
-  try {
-    parsedError = error
-      .split(':')[2]
-      .split('\n')[0]
-      .substring(1);
-  } catch (e) {
-    parsedError = 'Unknown error';
+  if (error.indexOf('You may only post once every 5 minutes') !== -1) {
```
jm90m: you might be able to use `_.includes(error, 'You may only post once every 5 minutes')`. Reads a bit better imo

src/client/errorMiddleware.js
```diff
@@ -1,17 +1,25 @@
...
+    return 'You may only post once every 5 minutes.';
+  } else if (error.indexOf('You may only comment once every 20 seconds') !== -1) {
+    return 'You may only comment once every 20 seconds.';
```
jm90m: Too bad we can't use `react-intl` here as it is in the middleware, might be good to create constants as some of these error messages are duplicates of what you are checking for. Like this shows up twice `You may only comment once every 20 seconds.`... But I did notice not all of them are duplicates, some have different message.. so might be good to make a map that contains ```{ ERROR_NAME: {error: '', message: ''} ``` (Something like that idk, you might be able to figure out different names). 

This is just a suggestion, PR looks fine to me either way.

Sekhmet: I didn't want to make them constants and use them both for fingerprinting and as error message, because we might want to change text that we display to users and error message from API might change as well.

src/client/errorMiddleware.js
Sekhmet: I like to use vanilla JS wherever possible (if it doesn't complicate code too much).

src/common/constants/errors.js
```diff
@@ -0,0 +1,26 @@
+export default {
+  POST_INTERVAL: {
```
jm90m: looks good 👍

PR #1417, opened 2018-01-30
Add LetsGetStarted Sidebar Block

### LINKED ISSUE #1254: Sidebar right block "Let's get started"
We should remove the right sidebar block "Never wrote a post? Start Now" and replace it with a list to complete with these steps:
image

jm90m:
Fixes #1254 

Changes:
- Add LetsGetStarted, HorizontalBarChart, LetsGetStartIcon components
- Hide LetsGetStarted Block only if user has completed all requirements & last created post is greater than 5 days ago. (subject to change based on feedback)


src/client/locales/en-US.json
```diff
@@ -287,5 +287,12 @@
   "login_modal_footer_text": "Don't have an account? Signup with {link}",
   "login_to_busy": "Login to Busy",
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
-  "login_with_steemconnect": "Login with SteemConnect"
-}
\ No newline at end of file
+  "login_with_steemconnect": "Login with SteemConnect",
```
Sekhmet: I don't think we should add translations there.

src/client/components/Sidebar/LetsGetStarted.js
```diff
@@ -0,0 +1,184 @@
...
+
+    console.log('DIFF DAYS', diffDays);
+    return diffDays > 5;
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      ...LetsGetStarted.getCurrentUserState(props.authenticatedUser, props.followingList),
```
Sekhmet: We don't have to spread this object. We can just assign it ti `this.state`.

src/client/components/Sidebar/LetsGetStarted.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  static displayContentsByPostRootDate(lastRootPost) {
+    if (lastRootPost === '1970-01-01T00:00:00') return true;
+    const lastRootPostDate = new Date(lastRootPost);
+    const currentDate = new Date();
+    const timeDiff = Math.abs(currentDate.getTime() - lastRootPostDate.getTime());
+    const diffDays = Math.ceil(timeDiff / (1000 * 3600 * 24));
+
+    console.log('DIFF DAYS', diffDays);
```
Sekhmet: Console log left there.

bonustrack:
Yes when all step completed it should be hidden.

src/client/components/Sidebar/LetsGetStarted.js
jm90m: good catch

src/client/locales/en-US.json
jm90m: to en-US?

jm90m:
@bonustrack yeah ill update this to hide when all steps are completed

src/client/locales/en-US.json
Sekhmet: Yeah, Crowdin uses `default.json` only.

jm90m: oh so we are gonna rely on crowdin for en-US translations?

bonustrack:
- Can you use these sentences:
> Complete your profile, Follow at least 5 steemians, Like some good posts, Write your first post
- Link the sentences to these pages:
Edit profile, Discover people, Trending, Editor
- Make the progress bar `.HorizontalBarChart__bar { height: 8px; }`

src/client/components/Sidebar/LetsGetStarted.js
```diff
@@ -0,0 +1,169 @@
...
+    followingList: PropTypes.arrayOf(PropTypes.string).isRequired,
+    isAuthFetching: PropTypes.bool.isRequired,
+    isFetchingFollowingList: PropTypes.bool.isRequired,
+    authenticatedUser: PropTypes.shape().isRequired,
+  };
+
+  static getCurrentUserState(authenticatedUser, followingList) {
+    const hasPost = authenticatedUser.last_root_post !== '1970-01-01T00:00:00';
+    const hasVoted = authenticatedUser.last_vote_time !== '1970-01-01T00:00:00';
+    const JsonMetadata = _.attempt(JSON.parse, authenticatedUser.json_metadata);
```
bonustrack: Can you use a small `j` `jsonMetadata`

src/client/components/Sidebar/LetsGetStarted.js
```diff
@@ -0,0 +1,169 @@
...
+    const hasProfile = _.has(JsonMetadata, 'profile');
```
bonustrack: We should check if he has a `name`, `about` and `profile_image`

jm90m:
@bonustrack should we keep it link color or change it to be black?

screen shot 2018-02-01 at 7 30 44 pm
screen shot 2018-02-01 at 7 30 41 pm

Also I added a tooltip to first one, since people might not know exactly what fields need to be filled out. What do you think?

jm90m:
For `Move the block to center (so its visible on mobile)`... you want me to display it on all feeds? Also I can have it on sidebar when not on mobile and display it on feed right underneath quick post editor for mobile.

bonustrack:
Keep it in black, dont show tooltip there, and show the block in center only in Feed page (homepage logged):
image

bonustrack:
We dont want border for this block on mobile:
image
Can't we have one class only to rule all these block?

jm90m:
@bonustrack yeah we can use SidebarContentBlock, since I retained those classes, just added to make sidebar content blocks more friendly on mobile

bonustrack:
The block appear when i load the page then disappear then appear again:
test
Can you make that it just appear once?

bonustrack:
And by the way activity block have right border on mobile:
image
Please use the same class for every center block border as i said, so this style bug will never happen anymore.

jm90m:
@bonustrack 
I don't think classnames are a good idea.. HTML should contain only the semantics, and styling or classes referring to styling should not be part of it. 

Also I noticed we are starting to have lots of `!important` around, but its understandable since their are a lot of cases where we do have to override the class. To solve this problem I suggest we go with using mixins (see latest commit), we can create mixins and customize them by passing params to them.

The goal should be to have the HTML describe the structure and the LESS/CSS describe how that structure should look. 

btw I kept the class `SidebarContentBlock` for the `LetsGetStarted` Component, since it uses same styling as the different sidebar contents (the padding, title styles, and content style). .can change though, but I figured it would be fine to keep it the same so we dont have duplicate CSS, can probably make `SidebarContentBlock.less` into a mixin too and just import it in all SidebarBlock components .less files.. but should probably save that for another PR

bonustrack:
I don't mean to use `classnames` lib. I mean to have only one class to define the style of every center block, a mixin could do that work. Also Lets get started block should not have a class `SidebarContentBlock` since its not in the sidebar anymore.

Sekhmet:
We use `!important` quite often, because not everything from antd is customizable, so we use it to change some styles.
However, I agree that we shouldn't be using `!important` to overwrite our own styles.

bonustrack:
I still having this issue https://github.com/busyorg/busy/pull/1417#issuecomment-362851911

jm90m:
@Sekhmet When you got time can ya review?

I added a HOC `DelayedComponent` so it delays the rendering of the specific component, since for some reason, I am still getting authenticated as `true` when it is clearly set to `false` but I think its an SSR Issue, I could be wrong though and could be something else, see gif, let me know if gif doesn't work.

https://giphy.com/gifs/3ohs4xmAKooN3hwUtG/fullscreen

bonustrack:
The block "Lets get started" use to appear earlier than this:
image
Can it appear earlier?

src/client/components/DelayComponent.js
```diff
@@ -0,0 +1,38 @@
...
+      super();
+      this.state = { ready: true };
+    }
+
+    componentWillMount() {
+      const { delay, onRender } = args;
+      const d = parseInt(delay, 10);
+      if (d && d > 0) {
+        this.setState({ ready: false });
+        this.timeout = setTimeout(() => {
```
bonustrack: There should be a better solution than adding timeout

jm90m: I'm pretty sure its this thats preserving the auth state - https://github.com/busyorg/busy/blob/ca3d908ad046461cf65263aace805aec92123209/src/server/handlers/createSsrHandler.js#L48

Look at my gif here: https://giphy.com/gifs/3ohs4xmAKooN3hwUtG/fullscreen

I was thinking of maybe preserving the state for everything else except `auth` for SSR

jm90m: @Sekhmet might have an idea, let me know your thoughts

pareshUD:
When the pages is loading I get this kind of error here 

screen shot 2018-02-08 at 1 58 57 am

bonustrack:
This issue is not related to this PR @pareshUD

PR #1419, opened 2018-01-30
Add TopNav Language Selection

### LINKED ISSUE #859: Show language selection for unlogged users
#### Expected behavior

There should be an option to select a language for users that are not logged in. It should be placed next to "Sign Up | Log in" at the top.

#### Actual behavior

User can't select a language if they aren't logged in.

jm90m:
Fixes #859.

Changes:
- Add LanguageSettings component to change between languages in the topnav


bonustrack:
By default there should be the language used, now when i load the page i see the interface in french but the dropdown with English
image

bonustrack:
For some reason, the hero never change language and stick in french for me.

jm90m:
@bonustrack should be fixed now, was testing on chrome and works fine for me when setting language to a different language for unlogged

src/client/components/Navigation/LanguageSettings.js
```diff
@@ -0,0 +1,144 @@
...
+import './LanguageSettings.less';
+
+@connect(
+  state => ({
+    authenticated: getIsAuthenticated(state),
+    locale: getLocale(state),
+  }),
+  {
+    saveSettings,
+    setLocale: locale => dispatch => dispatch(setLocale(locale)),
```
Sekhmet: AFAIK this would be the same as just writing `setLocale` instead of this whole line.

Sekhmet:
I think we can add some kind of loading state to language switcher. If user is logged in it can take a while to update.

mynameisek:
I think the purpose of this feature is to allow someone that turns to have the wrong language in homepage, to quickly switch to the good language. For that, it requires that the user quickly identify where he should click to switch language.

With this UX, this is not possible for him to understand where to click since he is not supposed to understand the initial language.

Imagine you are in internet café in China, and use Chinese computer, how you suppose to read chinese and click to modify into English ? You will not be able to do that. One solution could be the country Flag. If you have other ideas feel free to propose but that issue is real

jm90m:
@mynameisek Can't do country flag because that would signify that this one specific language belongs to this country, but I think it might be good to have it in the footer / sidebar like on FB, where they have English showing up all the time.

jm90m:
@Sekhmet thanks for the patch, ill try that

src/client/components/Navigation/LanguageSettings.js
jm90m: yeah looks like it is, seems to be working fine when I changed it

bonustrack:
Can we change the dropdown to a modal? I feel its too much result for a dropdown.
Also it does not fit on mobile, we need to figure out another design cc @pareshUD

jm90m:
yeah I'll update it to a modal, would be great to have another design

pareshUD:
Please check the previous designs where I have used the shorter names like EN or CN.. I also agree with Chris here as jjust using the language name will may confuse the user as if they can change the language or not.. 

@jm90m I think we can do something like facebook but I am not sure how often people change the language on busy and just one drop down will be enough I guess

jm90m:
@pareshUD, even if we change to short names, the list will still be the same length, the problem that @bonustrack is outlining, is the fact that the dropdown list is way too long and it doesn't all fit in mobile.. We could do a few things here.... 

1. create a modal with select dropdown
2. shorten the list to a few languages with the option to 'View more languages' at the bottom

lmk what you guys want to do @pareshUD @bonustrack

pareshUD:
@jm90m My only concern having the change language u footer is user will completely miss it that its there.. We can have that on sidebar but then again there are some pages which dont have sidebar or has different elements so changing position of the same also is not a good idea for that reason I think having that on top will be easy to access.. 

I agree with showing less languages and if you want to see whole list we can open up the popup modal for sure.

jm90m:
I think we should have languages also in a sidebar block like in FB, like the top 3-4 languages.. what do you think @bonustrack ? This way it solves @mynameisek issue

pareshUD:
I think its better not to use the English and use shorter one EN.. Which takes less space.. Also can we add search within the language dropdown for easy access may be?

jm90m:
@pareshUD idk talk to @bonustrack he decides, I personally like the long letters, and we still need design to address @mynameisek issue

bonustrack:
Menu should be aligned to red line. 
image
This is OBVIOUS @jm90m i wish you can figure out this thing yourself!

bonustrack:
And please remove tag "Need update" when you feel it's ready.

jm90m:
i was still working on this @bonustrack, there was a bug with this that was throwing lots of errors on my local. hence the `needs update` tag

bonustrack:
Ho i see my mistake sorry.

jm90m:
@bonustrack I am closing this and openign a new PR on different branch with same thing, since there's styling updates to the Topnav in here that are old and conflictign too much.

PR #1427, opened 2018-01-30
Update flagging UX

### LINKED ISSUE #1411: Change flag button color after the flag is recorded
https://busy.org/utopian-io/@littleboy/busy-org-change-flag-button-color-after-the-flag-is-recorded

Sekhmet:
Fixes #1411 

Changes:
- Add two states to flag button.
- Allow unflagging posts.


bonustrack:
Like icon should not be in loaded mode when flagging, i feel it's confusing now
image

PR #1440, opened 2018-01-31
Load more accounts for Steemd search

bonustrack:
Change number of results from `20` to `100` when query Steemd for accounts.


jm90m:
btw with this you'll see the usernames always first in the search so you're going to have to go through 1-100 names before you see posts. Unless we change sorting

Issue #1413, opened 2018-01-29
Render comments in chunks


Sekhmet:
#### Expected behavior

Comments shouldn't be rendered all at once. It should render around 20 comments in a tree and display `Show X more comments` link to render them.
If we link to comment directly its tree should be expanded by default.

#### Actual behavior

All comments are rendered at once and it takes quite some time to render them effectively freezing web page.

#### How to reproduce

Visit post with significant amount of comments and scroll to comments:
https://busy.org/steemstem/@steemstem/promoting-steemit-going-down-inside-the-large-hadron-collider-at-cern-with-steemstem

caleblogan:
I can work on this.

PR #1376, opened 2018-01-25
Add PostModal component & integrate it with Feed

jm90m:
Fixes #1331

Changes:
- Add PostModal Component and integrate it with SubFeed & UserProfileFeed

TODO:
- [x] navigating from topics, with follow buttons or hashes, component gets unmounted in React, but component still displays for some reason,
- [x] Loading comments is slow for large amount of comments (I guess this goes along with performance optimization
- [x] Image Modal appears in background of PostModa

PR #1444, opened 2018-02-01
Use ReactModal for PostModal

### LINKED ISSUE #1331: Add post modal

jm90m:
In an effort to figure out how to optimize the performance of this post modal, specifically with posts that have large comments `/dlive/@dlive/dlive-0-2-0-ui-now-optimized-for-mobile`.. I decided to try it out using ReactModal as well..

Let me know what you guys think is faster / better to use?

Fixes #1331

Changes:
- Use ReactModal instead of antd Modal


bonustrack:
On mobile it doesnt render well. Can you try on mobile and tell which method is faster? It feel similar speed at first glance too me

jm90m:
oh yeah I still need to update styles manually for mobile... forgot about that was just testing specifically for comments being rendered @bonustrack

Sekhmet:
It might get better after implementing #1413

bonustrack:
- The modal is not properly centered
- We should start from the top of the scroll
- There is a left margin on mobile we should remove that:
image

Do you feel it's better or faster than #1376 @Sekhmet @jm90m ?

jm90m:
@bonustrack I didn't do much styling on this, you have to manually do all the styles if you use ReactModal.. So I still need to do the mobile styles as well..

But I'm hoping @Sekhmet comment updates (https://github.com/busyorg/busy/issues/1413), can make it render faster.

Only was testing to see if ReactModal was faster than antd with comments for posts like the latest post on here: https://busy-master-pr-1444.herokuapp.com/@dlive

To me they both feel the same, depending on content

jm90m:
Closing this, will use antd modal.

PR #1448, opened 2018-02-01
Add border to avatars and make them circles

### LINKED ISSUE #1374: Use rounded (circle) avatars

### LINKED ISSUE #1433: Set default background on avatar or border
Avatar may be slow to load, we should be using a background so the user can see the avatar placeholder while the image is loading and avoid this:
image
If the background is not easy to set (because of default avatar), we can use a border instead.

Sekhmet:
Fixes #1374 
Fixes #1433 

We could add add background colour, but it doesn't look well when using avatars with transparency or round avatars. I created version of avatar that work with backgrounds, but with bigger number of comments (~100) it takes 6 seconds to dispatch and handle all the events).

Changes:
- Add border to Avatars.
- Make Avatars circles.

### DIFF
--- a/src/client/components/Avatar.less
+++ b/src/client/components/Avatar.less
@@ -1,7 +1,10 @@
+@import (reference) '../styles/custom.less';
+
 .Avatar {
   display: block !important;
   border-radius: 20%;
   background-size: cover;
   background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


bonustrack:
Can you use that PR to do #1374 too?

PR #1453, opened 2018-02-01
Update Transfer Modal & remove container min-height

### LINKED ISSUE #1438: Remove default max-height
Page come by default with a scrollbar, it should not be the case:
image
Probably because of recents changes https://github.com/busyorg/busy/pull/1416

jm90m:
Fixes #1438 

Changes:
- Remove unnecessary container class
- Remove container-height


bonustrack:
And... this issue is back
image

jm90m:
yeah @bonustrack i reopend issue for it, but for some reason im not seeing that issue

jm90m:
can you reproduce the issue that @bonustrack is talking about @Sekhmet ?

Sekhmet:
Nope.

bonustrack:
Did you tried to reproduce my issue logged @jm90m @Sekhmet ?

jm90m:
@bonustrack I see the issue rarely when I am logged in, its hard for me to reproduce and stop it, but I think I figured out the issue. It's the styling that was applied here - https://github.com/busyorg/busy/blame/master/src/client/styles/common.less#L40

Not too sure if this will cause any issues @Sekhmet, but I tested on Firefox and everything seems to be working fine.

src/client/styles/common.less
```diff
@@ -37,7 +37,7 @@ a:focus {
 }
 
 .ant-layout {
-  overflow: hidden;
+  overflow: visible;
```
Sekhmet: Isn't `visible` the default?

jm90m: Nah it's never specified, for `.ant-layout` but default for all html elements is `visible`. Thought it would be good to explicitly set it.

PR #1455, opened 2018-02-02
Add USD value in Transfer Modal

### LINKED ISSUE #1452: Transfer modal: show the equivalent amount in $ when doing a transfer
When i do a transfer i want to see how much the SBD or Steem that i want to send is worth in $

@pareshUD do you have a design for this?

jm90m:
Fixes #1452 

Changes:
- Display current USD amount of transfer

screen shot 2018-02-01 at 9 39 44 pm

### DIFF
--- a/src/client/locales/default.json
+++ b/src/client/locales/default.json
@@ -314,5 +314,4 @@
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
   "login_with_steemconnect": "Login with SteemConnect",
   "follows_you": "Follows you",
-  "transfer_current_usd_amount": "Current USD value of this transfer: ${amount}"
 }
--- a/src/client/locales/default.json
+++ b/src/client/locales/default.json
@@ -313,5 +313,5 @@
   "login_to_busy": "Login to Busy",
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
   "login_with_steemconnect": "Login with SteemConnect",
-  "follows_you": "Follows you",
+  "follows_you": "Follows you"
 }
--- a/src/client/locales/default.json
+++ b/src/client/locales/default.json
@@ -313,5 +313,6 @@
   "login_to_busy": "Login to Busy",
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
   "login_with_steemconnect": "Login with SteemConnect",
-  "follows_you": "Follows you"
+  "follows_you": "Follows you",
+  "transfer_current_usd_amount": "Current USD value of this transfer: ${amount}"
 }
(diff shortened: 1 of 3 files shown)


bonustrack:
It would be great to show the price like this: 
image
With text in grey color. If you cant do this, we need @pareshUD ideas

jm90m:
yeah thats fine I'll see what i can do

jm90m:
Should we also display the 'Current USD value of this transfer' text

jm90m:
@pareshUD would be good to have some ideas also

bonustrack:
We dont need to show `Current USD value of this transfer` message. Just showing the estimation is self explanatory

pareshUD:
transfer
transfer2

@jm90m @bonustrack Please try this.

Sekhmet:
@pareshUD Fabien suggested displaying value inside of input:
https://github.com/busyorg/busy/pull/1455#issuecomment-362841084

pareshUD:
yes @Sekhmet But I tried to display it in a better way.. Like having them on the far right.. But then I am finding the issue with mobile version when the screen resolution is smaller.. So I might reworked on this

bonustrack:
We can use the same position as @pareshUD suggested, should be easier to implement anyway.

jm90m:
yea this is way easier to implement

jm90m:
Updated with @pareshUD & @bonustrack  style

src/client/locales/default.json
```diff
@@ -313,5 +313,6 @@
   "login_to_busy": "Login to Busy",
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
   "login_with_steemconnect": "Login with SteemConnect",
-  "follows_you": "Follows you"
+  "follows_you": "Follows you",
+  "transfer_current_usd_amount": "Current USD value of this transfer: ${amount}"
```
Sekhmet: This isn't used anymore so I think we don't have to touch translations.

jm90m: yeah ill remove it

src/client/locales/default.json
```diff
@@ -313,5 +313,5 @@
   "login_to_busy": "Login to Busy",
   "login_modal_description": "Login with your Steem account using SteemConnect to enjoy Busy at 100%",
   "login_with_steemconnect": "Login with SteemConnect",
-  "follows_you": "Follows you"
+  "follows_you": "Follows you",
```
Sekhmet: Technically it's not correct JSON with trailing comma, but I think Crowdin will work with this anyway.

jm90m: no you're right, I'll remove it

jm90m: let me remove it

pareshUD:
Can we remove the - from $.. It looks like the balance is in negative at the moment. @jm90m @bonustrack 

screen shot 2018-02-08 at 1 52 35 am

bonustrack:
I think it's cool with the `~` it's not a `-`, it's an estimation, it make sense to have it, we have it also in vote value.

jm90m:
yeah I agree with @bonustrack I added the `~` to signify that its an estimation

pareshUD:
How about we use "=" instead?

jm90m:
you can't use `=` that would signify that it is that exact value, this is an estimation

Issue #1456, opened 2018-02-02
"Follow you" appear for unlogged user on mobile


bonustrack:
This account is not following me and i'm not logged but i can see "Follow you":
![screenshot_20180202-145633](https://user-images.githubusercontent.com/16245250/35722581-83a35a68-0829-11e8-95fe-8ee0e58be7c5.png)

jm90m:
should be fixed in #1457

PR #1457, opened 2018-02-02
Add authenticated check for `Follows you`

### LINKED ISSUE #1418: 'Follows you' displays for unlogged users
#### Expected behavior
'Follows you' text in User Profile page displays for unlogged user, when you go to the page directly.

#### Actual behavior
'Follows you' text in User Profile page should not appear for unlogged user.
screen shot 2018-01-29 at 10 23 43 pm
 

#### How to reproduce
Navigate to any user profile page directly as unlogged

* Browser: Chrome
* Operating system: OSX

jm90m:
Fixes #1418

Changes:
- add authenticated check for `Follows you`


Sekhmet:
I can't reproduce this problem. I'm not logged in and I can't make "Follows you" to display in any way (on production and staging as well).

bonustrack:
Did you tried on mobile @Sekhmet ?

Sekhmet:
Doesn't happen for me on mobile as well. Should I visit user profile page directly or by navigating there using links? Neither causes problem for me.

jm90m:
@Sekhmet Its happening for me right now on staging
screen shot 2018-02-02 at 4 44 20 am

it happens when you directly go to the page sometimes

src/client/components/UserHeader.js
```diff
@@ -103,7 +104,8 @@ const UserHeader = ({
               />
             </div>
           </div>
-          {isFollowing &&
+          {authenticated &&
+            isFollowing &&
```
Sekhmet: I think `isFollowing` should take whether user is logged in or not in account. Can we update how `isFollowing` is calculated, so `isFollowing` really means that user is following me.

jm90m:
@Sekhmet good idea, just updated

Sekhmet:
As I can't reproduce I'm just guessing, but it wouldn't be better to set `isFollowing` to `false` by default and never check if user is following you, if you are not authenticated?

jm90m:
So this happens when you login and logout, and move around, looks like @bonustrack is experiencing it too on mobile - https://github.com/busyorg/busy/issues/1456

I updated it, I forgot why I was using a null check there, but could've been with something in the past, that is now fixed. Refactored it.

PR #1458, opened 2018-02-02
Make EditorInput compatible with antd

### LINKED ISSUE #1450: Make EditorInput compatible with antd Forms
Currently there is a lot of magic happening when dealing with EditorInput. If we made it compatible with antd Forms, we could use `getFieldDecorator` like with everything else.

https://github.com/ant-design/ant-design/blob/master/components/input/Input.tsx

Sekhmet:
Fixes #1450 

Changes:
- [x] Make `EditorInput` compatible with antd.
- [x] Use new editor in forms.
- [x] Improve performance.
- [x] Restore previous selection behaviour.

### DIFF
--- a/src/client/components/Editor/Editor.js
+++ b/src/client/components/Editor/Editor.js
@@ -150,13 +150,14 @@ class Editor extends React.Component {
   }
 
   throttledUpdate() {
-    this.props.form.validateFieldsAndScroll((err, values) => {
-      if (err) this.props.onError();
-      else {
-        this.setBodyAndRender(values.body);
-        this.props.onUpdate(values);
-      }
-    });
+    const { form } = this.props;
+
+    if (Object.values(form.getFieldsError()).filter(e => e).length > 0) return;
+
+    const values = form.getFieldsValue();
+
+    this.setBodyAndRender(values.body);
+    this.props.onUpdate(values);
   }
 
   handleSubmit(e) {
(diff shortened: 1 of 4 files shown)


src/client/components/Editor/EditorInput.js
```diff
@@ -11,7 +11,7 @@ import './EditorInput.less';
 
 class EditorInput extends React.Component {
   static propTypes = {
-    initialValue: PropTypes.string,
+    value: PropTypes.string, // eslint-disable-line react/require-default-props
```
Sekhmet: We have to disable this rule. If we set `defaultProp` on it, antd will complain about updating value without `setFieldsValue`.

bonustrack:
Error should not fire if i dont unfocus the field, now its not possible to write a post without triggering the errors:
image

src/client/components/Editor/Editor.js
```diff
@@ -118,29 +117,15 @@ class Editor extends React.Component {
...
-        ...form.getFieldsValue(['title', 'topics', 'reward', 'upvote']),
-      };
-
-      this.props.onUpdate({
-        ...values,
-        topics: values.topics.slice(0, 5),
-        title: values.title.slice(0, 255),
-      });
-    }, 0);
+    _.throttle(this.throttledUpdate, 200, { leading: false, trailing: true })();
```
jm90m: looks better than setTimeout 👍

Sekhmet:
@bonustrack now it won't validate entire form on update (same as before), but also will prevent saving invalid data (they can be empty though).

PR #1459, opened 2018-02-02
Make Comments header responsive

### LINKED ISSUE #822: Comments sorting options don't fit the line in some languages
Czech language

English for comparison

#### Expected behavior

Comments sorting options don't overflow.

#### Actual behavior

With some languages, the sort options may overflow because the words are too long.

#### Steps to reproduce

1. Set Czech language. (It may occur with others too.)
1. Open a post on a mobile.
1. See comments.

- Browser: Brave
- Operating system: Android 7.1.2

Sekhmet:
Fixes #822 

image

Changes:
- Add `SortSelector` component.
- Use `SortSelector` for feed and comments sorting.

### DIFF
--- a/src/client/components/Comments/Comments.js
+++ b/src/client/components/Comments/Comments.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 import { injectIntl, FormattedMessage } from 'react-intl';
 import { message } from 'antd';
 import { MAXIMUM_UPLOAD_SIZE_HUMAN } from '../../helpers/image';
 import { sortComments } from '../../helpers/sortHelpers';
 import Loading from '../Icon/Loading';
+import SortSelector from '../SortSelector/SortSelector';
 import CommentForm from './CommentForm';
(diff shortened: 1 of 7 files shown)


bonustrack:
I think a dropdown would fit better in mobile

Sekhmet:
I'm not big fan of dropdowns where normal links would work just well. What do you think @pareshUD?

jm90m:
Icons would be better if people can differentiate, which one is which, imo, but I'll let @pareshUD handle design :)

bonustrack:
Dropdown fill into same line than `Comments` title, it's more ergonomic to use same line than creating a new line.

bonustrack:
Icon is confusing, user has to learn what is what..

Sekhmet:
I updated it to use selector:
image

However, I will mark it as "Needs update", because for some reason it's causing feed to render posts with the same keys.

Sekhmet:
It looks like it's not this PR that causes duplicate keys, so I will remove "Needs update" badge and leave it for review.
I will work on duplicate keys here: #1469

PR #1471, opened 2018-02-04
Add Signature tool to profile settings

### LINKED ISSUE #1436: Suggestion - Signature tool
#### Expected behavior

When opening the editor, a button should allow to paste a **predifined signature** at current cursor position. 
The predifined should be editable in the parameters page

#### Actual behavior

As feature is absent, i need to open a previous post, edit, copy the signature, cancel edition, paste it in the new post.

* Browser:  chrome / firefox
* Operating system: Ubuntu latest LTS

Sekhmet:
Fixes #1436 

Changes:
- Add Signature tool to user profile settings
- Include signature in user posts.

I couldn't really test it on real signature stored in profile settings, because I don't have active key for hellosteem, and I didn't want to post using my account.

### DIFF
--- a/src/client/components/Story/StoryFull.js
+++ b/src/client/components/Story/StoryFull.js
@@ -40,6 +40,7 @@ class StoryFull extends React.Component {
     rewardFund: PropTypes.shape().isRequired,
     defaultVotePercent: PropTypes.number.isRequired,
     onActionInitiated: PropTypes.func.isRequired,
+    signature: PropTypes.string,
     rewriteLinks: PropTypes.bool,
     pendingLike: PropTypes.bool,
     pendingFlag: PropTypes.bool,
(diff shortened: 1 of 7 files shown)


bonustrack:
The way i was seeing it, when user add his signature it would append in any of his post automatically, and we would not need to add on his post. Now the signature is simply being added on the post. We can't edit the signature for all posts in 1 time, and we can't add signature on old posts.

Sekhmet:
So it should not be stored in the post itself, but just be displayed if user has one?
It is nothing like it was proposed in #1436

bonustrack:
I think we can come across 2 features:

One is predefined template, which can be stored in user SC2 metadata and make easy to start new post with a predefined template.

Another feature is the signature, which should work IMO like phpBB where you can edit your signature anytime and it change it on all of your posts. For this it make more sense to use steemd account `json_metadata` to store it and not duplicate this content everytime you publish a new post.

Sekhmet:
Updated.
You can see signature below my posts:
https://busy-master-pr-1471.herokuapp.com/@sekhmet/hello-world-i-m-joining-busy-org-team

PR #1477, opened 2018-02-05
Add custom tag for latex

### LINKED ISSUE #1467: Math/Science Formulas
Some scientists are interesting to post Maths on Busy, but it does not display correctly, one solution is to use a library like KaTeX
https://khan.github.io/KaTeX/

image

Sekhmet:
Fixes #1467 

Changes:
- Add `improve` helper that we can use to add custom tags.
- Add `[+]latex[+]` tag to render LaTeX using Codecogs.
- Improve post preview and final contents when submitting.

image

To add LaTeX to post you can use custom tag:
```
[+]Latex there[+]
```

like this:
```
[+]
f(x) = \int_{-\infty}^\infty
    \hat f(\xi)\,e^{2 \pi i \xi x}
    \,d\xi
[+]
```

Post preview: http://localhost:3000/@hellosteem/testing-custom-latex

This is just custom tag, that will be replaced when adding post so it will work in other clients as well.

### DIFF
--- a/src/client/helpers/improve.js
+++ b/src/client/helpers/improve.js
@@ -1,0 +1,8 @@
+const latexRegex = /\[\+\]((\n|.)*?)\[\+\]/g;
+
+export default function improve(body) {
+  return body.replace(
+    latexRegex,
+    (match, p1) => `![${p1}](https://latex.codecogs.com/gif.latex?${encodeURI(p1)})`,
+  );
+}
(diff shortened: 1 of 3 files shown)


src/client/helpers/improve.js
```diff
@@ -0,0 +1,8 @@
+const latexRegex = /\[\+\]((\n|.)*?)\[\+\]/g;
```
Sekhmet: This could be in `regexHelpers` but I believe it makes sense to separate regexes used for improve engine from other ones. This might become it's own package in the future, so it makes sense to leave it there.

PR #1479, opened 2018-02-05
Add scrollbars to tags and update scrollbars styles

### LINKED ISSUE #1464: Sliding tag list on small screen
We should the list of tag in bottom of a post a slider on small screen. Similar than user menu.
image
image
Also the scrollbar should not appear on these sliding menus:
image

Sekhmet:
Fixes #1464 

Changes:
- [x] Add scrollbars to tags.
- [x] Hide scrollbars thumb.


src/client/components/Story/StoryFull.js
```diff
@@ -462,7 +463,18 @@ class StoryFull extends React.Component {
           />
         )}
         <div className="StoryFull__topics">
-          {tags && tags.map(tag => <Topic key={tag} name={tag} />)}
+          <Scrollbars
+            universal
+            autoHide
+            renderView={({ style, ...props }) => (
+              <div style={{ ...style, marginBottom: '-20px' }} {...props} />
```
Sekhmet: `react-custom-scrollbars` uses native track and thumb on mobile, so it is not customizable. This hides them.

bonustrack:
image
Can we have the tags continue to the red rectangle?

pareshUD:
@bonustrack Or can simply wrap it and put 2 more or something like this?

src/client/components/Story/StoryFull.js
jm90m: correct me if im wrong but isn't style the only thing being passed here for `renderView` ?

https://github.com/malte-wessel/react-custom-scrollbars/blob/master/src/Scrollbars/index.js#L574

So dont need the `{...props}` right?

Sekhmet:
I think that just wrapping tags would be much easier to understand for users (especially if we remove scroll thumb).

src/client/components/Story/StoryFull.js
Sekhmet: They use props here as well:
https://github.com/malte-wessel/react-custom-scrollbars/blob/master/examples/simple/components/ColoredScrollbars/ColoredScrollbars.js#L19

jm90m: ahh i see ok

bonustrack:
What you mean wrap?

jm90m:
Like how we currently have it @bonustrack

bonustrack:
Medium mobile app have a slider on tags list with scrollbar, i feel it's good UX

jm90m:
I use medium mobile app almost every day, since I used to write on there last year, I don't see tags on the iOS app, but mobile responsive medium wraps tags
screen shot 2018-02-08 at 7 36 39 pm

Sekhmet:
I don't think we should do something, because Medium does. Medium isn't always that great and worth to take inspiration from. They used to display fixed topnav and footer all the time that covered 30% of screen.

Sekhmet:
I tried making similar behaviour as Google tags, but can't get right padding to work when scrolled to the end of the list:
image
Do you mind taking a look @jm90m?

jm90m:
just saw this @Sekhmet, yeah I'll take a look!

jm90m:
@Sekhmet 
this is what you'll want to do:
```
diff --git a/src/client/components/Story/StoryFull.js b/src/client/components/Story/StoryFull.js
index 99dde68..dcbd00d 100644
--- a/src/client/components/Story/StoryFull.js
+++ b/src/client/components/Story/StoryFull.js
@@ -467,7 +467,11 @@ class StoryFull extends React.Component {
             universal
             autoHide
             renderView={({ style, ...props }) => (
-              <div style={{ ...style, marginBottom: '-20px', padding: '0 20px' }} {...props} />
+              <div
+                style={{ ...style, marginBottom: '-20px', padding: '0 20px' }}
+                {...props}
+                className="StoryFull__topics__container"
+              />
             )}
             style={{ width: '100%', height: 46 }}
           >
diff --git a/src/client/components/Story/StoryFull.less b/src/client/components/Story/StoryFull.less
index 34b2362..70d2a53 100644
--- a/src/client/components/Story/StoryFull.less
+++ b/src/client/components/Story/StoryFull.less
@@ -53,6 +53,11 @@
       margin: 16px 0;
     }

+    &__container {
+      margin-right: 20px !important;
+      overflow-y: hidden !important;
+    }
+
     &__content {
       display: flex;
       height: 46px;
@@ -112,6 +117,7 @@
     cursor: pointer;
   }
 }
+
 div[data-dir='rtl'] .StoryFull {
   &__header {
     &__text {
```

screen shot 2018-02-16 at 3 51 09 pm

Try that and let me know if thats what you guys wanted!

bonustrack:
I dont want this empty space, there should be tag continuing instead
image

Sekhmet:
slider

PR #1481, opened 2018-02-06
Add peer dependency so `npm install` works

kirkins:
I'm in the process of running this locally for the first time.

I had a dependency error when running `npm install`.

This change makes it so `npm install` will work right after cloning as described in the README.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,9 @@
     "webpack-dev-server": "~2.9.3",
     "webpack-hot-middleware": "^2.20.0"
   },
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


jm90m:
Did you try `yarn install`? README might be out of date

kirkins:
Maybe I'm wrong, I'm having trouble reproducing the error I was getting earlier.

I'm not sure what would have changed, but if I can't reproduce the error I will close PR.

kirkins:
Not sure why I was getting the message about peer dependency earlier. I've changed locations and `npm install` is working fine.

PR #1482, opened 2018-02-06
Access settings to see if should upvote on quickpost

### LINKED ISSUE #1480: Quick-post without liking?
I really enjoy the quick-post feature where you can make a post with title only like Twitter:

untitled

Whenever I make a post here it is auto-liked by my account.

Would it be possible to add a setting to user profiles to turn the auto-like off?

kirkins:
Fixes #1480

As title says fixes quick post always upvoting even when setting is set to false.

I have tested that the change works locally.

<s>I had to over-ride the pre-commit. I'm not sure about the prop-types.</s>

<s>Can someone advice what is needed:</s>

*edited: lint issue fixed*


jm90m:
Might also be good to actually display this setting in the `Settings` page, because right now this setting is kind of hidden, and only activated after you have created a post. what do you think @bonustrack

PR #1488, opened 2018-02-07
Update search placeholder & styles

### LINKED ISSUE #1466: Change input style for search
Can we have border on the search input in the topnav, use `45px` height and add `What are you looking for?` placeholder like this:
image

jm90m:
Fixes #1466

Changes:
- Update placeholder, height, and add border to search


bonustrack:
This
image

Does not look like this:
image

We need to remove this:
image

Also there should be a border-radius (similar radius than buttons)

Sekhmet:
I think that input should have a little bit of margin on the left on mobile:
image

pareshUD:
On mobile why icon is Purple and when you click on it the aligment is completely off

screen shot 2018-02-08 at 1 49 44 am
screen shot 2018-02-08 at 1 49 49 am

pareshUD:
Also lets make the size of the search box 40px and not 45px that way it will look good and have some padding around it.

jm90m:
@bonustrack @Sekhmet I added a 15px margin left for mobile search bar so its aligned more with the X

screen shot 2018-02-07 at 8 30 16 pm

Also do you guys really want it at 40px instead of 45px for the search bar?? This is what it looks like at 40px with it centered vs what is on this PR. LMK what you want @bonustrack / @pareshUD, you guys both talk it out.

screen shot 2018-02-07 at 8 31 24 pm

jm90m:
@Sekhmet yeah just updated to only reflect in default.json and no other text is formatted except the one i used

bonustrack:
- Border color of the input must be `#eae7e7` similar than the others borders. We should set this color only 1 time in our style for all the app.

- The autocomplete block should start at the red line
image
In a way that the autocomplete block is hiding the bottom border and border radius of the search input

bonustrack:
The autocomplete does not fit well:
image
Can we have same border than the input on the autocomplete block and also a padding top of 10px?

bonustrack:
We need bottom border aswell
image

src/client/components/Navigation/Topnav.less
```diff
@@ -180,11 +183,15 @@ div[data-dir='rtl'] .Topnav {
...
   position: fixed;
   border-radius: 0;
-  border-bottom-left-radius: 4px;
-  border-bottom-right-radius: 4px;
+  border-bottom-left-radius: @border-radius-base;
+  border-bottom-right-radius: @border-radius-base;
+  border-right: @base-border;
+  border-left: @base-border;
+  border-bottom: @base-border;
+  box-shadow: none;
```
bonustrack: Can you just add some `padding-top: 5px;` here to the input is not being cut with the autocomplete block.

PR #1505, opened 2018-02-09
Migrate to WebSockets

### LINKED ISSUE #1487: Go back to websocket for run all services?
In the long run we will run a WebSocket server to handle notifications. We are also going to store in a Redis database view count of post. Instead of cumulating connections (the json rpc api with api.steemit.com, the notification server, the API for get redis data) we could have a single WebSocket server at `busy-api` which proxy the json rpc api with api.steemit.com and may add additional data on responses like view count, or create new methods for the API, or cache responses, or send notifications to users.

What's your through on this @Sekhmet @jm90m @gregory-latinier?

Easiest way to go would be to create a small lib like `lightrpc` but with WS, kind of what is done here https://github.com/bonustrack/kbyte and use first a WS of Steemd node like `wss://steemd.privex.io` then switch to our proxy.

Sekhmet:
Fixes #1487 

Changes:
- Migrate to websockets (using `busyjs`).
- Update busyjs to be compatible with lightrpc (https://github.com/busyorg/busyjs/tree/dev)


src/client/steemAPI.js
```diff
@@ -1,9 +1,10 @@
-import { createClient } from 'lightrpc';
+const { Client } = require('busyjs');
+
+const client = new Client('wss://steemd.privex.io');
```
Sekhmet: We have env variables for this, but they are used for HTTPS endpoint. How do we handle this now, so there are no problems when deploying?

src/client/steemAPI.js
```diff
@@ -1,9 +1,10 @@
-import { createClient } from 'lightrpc';
+const { Client } = require('busyjs');
```
Sekhmet: We are probably doing something wrong when building busyjs. I made it use CommonJS modules only, but we can't use imports now.

src/client/steemAPI.js
bonustrack: We can create a config var like `API_URL` with `wss://steemd.privex.io` as default

bonustrack:
Heroku build is failing

package.json
```diff
@@ -55,6 +55,7 @@
     "babel-runtime": "^6.18.0",
     "body-parser": "~1.13.2",
     "bs58": "^4.0.0",
+    "busyjs": "https://github.com/busyorg/busyjs#dev",
```
bonustrack: This PR should remove `lightrpc`

Sekhmet:
It should be good for a review. I wonder what should we do with `STEEMJS_URL` environment variable. It currently points to HTTPS API. Should we use it still and change URL on the fly or use another one? Which node should we use by default?

bonustrack:
@Sekhmet we should change `STEEMJS_URL` to `API_URL` with `https://api.steemit.com` by default for now.

bonustrack:
We will not merge this PR yet. We will need to test it out with api.busy.org first.

PR #1506, opened 2018-02-09
Refactor Feed

### LINKED ISSUE #1492: Add exit page for external links
Hi,

I stumble upon a (minor) security issue that I would like to report you. 

Can you give me a email address to share the informations I have?

### LINKED ISSUE #1494: Prevent unnecessary feed re-renders
Currently we subscribe to lots of store parts inside of Feed:
https://github.com/busyorg/busy/blob/34a8497acb27062208abf2a31448e07d9a7d1d11/src/client/feed/Feed.js#L35-L50

Because of that when one of them changes it causes re-render of entire Feed which is slow. We used to use Storybooks, which made us create lots of stateless components, that required to have props passed to them. I think we should create smaller, but more focused on specific functionality container components.

In this issue I think we can improve rendering of feed by creating `StoryContainer` that would subscribe only to its own properties. This way we wouldn't need to subscribe to likes, pending likes, follows, bookmarks etc. inside of `Feed` - only relevant parts of application would re-render.

We can even create more container components, so we don't even re-render entire Story, but only StoryMenu or Slider.

Sekhmet:
Fixes #1494 

Changes:
- Added StoryContainer.
- Feed now subscribes to its own props.

**Tests:**
- [x] User feed
- [x] Trending feed
- [x] Comments
- [x] Replies
- [x] Bookmarks
- [x] SSR
- [x] Upvotes
- [x] Flags
- [x] Saving posts
- [x] Following user
- [x] Editing posts

### DIFF
--- a/src/client/feed/Feed.js
+++ b/src/client/feed/Feed.js
@@ -1,229 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import { push } from 'react-router-redux';
-import _ from 'lodash';
 import ReduxInfiniteScroll from '../vendor/ReduxInfiniteScroll';
-import { getHasDefaultSlider } from '../helpers/user';
-import { isPostDeleted } from '../helpers/postHelpers';
-import * as bookmarkActions from '../bookmarks/bookmarksActions';
-import * as reblogActions from '../app/Reblog/reblogActions';
-import * as postActions from '../post/postActions';
-import { followUser, unfollowUser } from '../user/userActions';
-import { editPost } from '../post/Write/editorActions';
-
-import {
-  getAuthenticatedUser,
-  getBookmarks,
-  getPendingBookmarks,
-  getPendingLikes,
-  getRebloggedList,
-  getPendingReblogs,
-  getFollowingList,
-  getPendingFollows,
-  getIsEditorSaving,
-  getVotingPower,
-  getRewardFund,
-  getVotePercent,
-  getShowNSFWPosts,
-} from '../reducers';
-
-import Story from '../components/Story/Story';
+import StoryContainer from './StoryContainer';
 import StoryLoading from '../components/Story/StoryLoading';
 import './Feed.less';
 
-@connect(
-  state => ({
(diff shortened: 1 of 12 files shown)


bonustrack:
How does this solve #1492?

Sekhmet:
Whoops, mentioned wrong issue.

src/client/feed/Feed.js
```diff
@@ -1,225 +1,36 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import { push } from 'react-router-redux';
-import _ from 'lodash';
 import ReduxInfiniteScroll from '../vendor/ReduxInfiniteScroll';
```
jm90m: This file looks way more readable 👍

src/client/feed/StoryContainer.js
```diff
@@ -0,0 +1,77 @@
...
+    isSaved: !!getBookmarks(state)[post.id],
+    isLiked: userVote.percent > 0,
+    isReported: userVote.percent < 0,
+    userFollowed: getFollowingList(state).includes(post.author),
+  };
+
+  const pendingLikes = getPendingLikes(state);
+
+  const pendingLike =
+    pendingLikes[post.id] &&
```
jm90m: `pendingLikes[post.id]` is repeated multiple times from here until line 50, maybe it would be a good idea to instantiate it like `const pendingLike = _.get(pendingLikes, post.id, {})`

src/client/feed/feedActions.js
```diff
@@ -43,13 +42,16 @@ export const getMoreFeedContent = ({ sortBy, category, limit = 20 }) => (
   getState,
   { steemAPI },
 ) => {
-  const { feed, posts } = getState();
-  const feedContent = getFeedContentFromState(sortBy, category, feed, posts.list);
+  const { feed, ...state } = getState();
```
jm90m: any reason for making a copy here? Couldn't it just be `const state = getState();`const { feed } = state;`?

Sekhmet: Right, I think we could even just use `state.feed`.

PR #1516, opened 2018-02-12
Configure scripts

Sekhmet:
Now editor should pick up prettier settings from code directly (so they don't have to be configured manually), and locales shouldn't be autoformatted (by prettier and editorconfig).

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
+  trailingComma: 'all',
+  printWidth: 100,
+};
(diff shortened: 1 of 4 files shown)


.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
```
jm90m: Awesome, these were the exact params I've been using for almost the past year using prettier 👍 

screen shot 2018-02-12 at 9 59 56 am

```
/usr/local/lib/node_modules/prettier/bin/prettier.js --write --single-quote=true --trailing-comma=all --print-width=100 src/client/app/appReducer.js
```

PR #1518, opened 2018-02-12
Check if feed is empty for EmptyFeed

### LINKED ISSUE #1510: "Ops this feed is empty" appear in end of feed
If you scroll this page for example https://busy.org/created/helloworld when you will reach the end of the feed the message `Ops this feed is empty` appear.
image
It should appear only when the feed has 0 results, not when there is results and we reach the end of the page.

Sekhmet:
Fixes #1510 

Changes:
- Check if the content is empty when displaying EmptyFeed.

### DIFF
--- a/src/client/feed/SubFeed.js
+++ b/src/client/feed/SubFeed.js
@@ -140,12 +140,13 @@ class SubFeed extends React.Component {
       loadMoreContent = () => this.props.getMoreFeedContent(sortBy, match.params.category);
     }
 
-    const loadScrollToTop = _.isEmpty(content);
+    const empty = _.isEmpty(content);
+    const displayEmptyFeed = empty && fetched && loaded && !isFetching;
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


src/client/feed/SubFeed.js
```diff
@@ -135,19 +135,19 @@ class SubFeed extends React.Component {
...
-        {loadScrollToTop && <ScrollToTop />}
+        {empty && <ScrollToTop />}
         <Feed
           content={content}
           isFetching={isFetching}
           hasMore={hasMore}
           loadMoreContent={loadMoreContent}
         />
-        {fetched && loaded && !isFetching && <EmptyFeed />}
+        {empty && fetched && loaded && !isFetching && <EmptyFeed />}
```
jm90m: might be good to put this at the top as `const displayEmptyFeed = empty && fetched && loaded && !isFetching;` for more readability/context.

Your choice :)

PR #1537, opened 2018-02-15
Update editor styles

### LINKED ISSUE #1461: Improve editor toolbar design
Instead of this:

image

We could have like this:
image

jm90m:
Fixes #1461

Changes:
- Use scroll for small screens for editor
- Update editor design based on mock


bonustrack:
Quick review:
- [ ] Add border primary color when focus on the input
- [ ] Remove border on editor tool icons on mouse over and other style, just make it grey by default and black on mouse over.
- [ ] Add left and right margin of 15px
image

Issue #1529, opened 2018-02-14
Content is not updated


kekkyojin:
#### Expected behavior

After navigating on the website -opening profile or visiting some categories-, new content should be shown.

#### Actual behavior

Clicking in any tab is showing always the last shown content.

#### How to reproduce

Open busy.org. Look at the content. Go to your wallet (also see issue #1174, it is the same as this one, but this is more generic) or your profile or some category, it doesn't matter.

If you do some transfer, the balance will not appear as updated, even navigating and coming back to the wallet.

Wait for some time for your followers to publish new content.

Navigate again to your main page. You will see the last shown content. After refreshing the browser then every new post will be shown now.

* Browser: Tested in Firefox 58 and Opera 51
* Operating system:  macOS 10.13.3

Sekhmet:
Duplicate of #1406

kekkyojin:
Yes, sorry, it seems to be duplicate of that bug. I didn't search deep enough.

PR #1538, opened 2018-02-16
Scroll back to top when user clicks profile link

kirkins:
This improves the usability for 2 situations I've noticed.

1. You scroll far down your history on profile and then click profile link expecting to go back to top but it doesn't.

2. If you are in another tab on your profile, for example activity, scroll far down, then click profile link, you will change tabs but still be scrolled down.

This seems to fix both of those issues.

_Edit: these problems are specific to viewing your own profile._


src/client/components/Navigation/Topnav.js
```diff
@@ -119,7 +119,7 @@ class Topnav extends React.Component {
             </Tooltip>
           </Menu.Item>
           <Menu.Item key="user" className="Topnav__item-user">
-            <Link className="Topnav__user" to={`/@${username}`}>
+            <Link className="Topnav__user" to={`/@${username}`} onClick={window.scrollTo(0, 0)}>
```
jm90m: We usually create `onClick` handlers for this sort of thing, to provide it with some context.

jm90m:
What do you think about this @bonustrack ?

kirkins:
There is also the question of would it better to refresh than scroll to top.

When you scroll to top content doesn't actually refresh.

I can put it in a handler I normally don't for something so short but however you usually do it is fine for me.

kirkins:
Then again it seems the refresh issue might be a bigger one outside the scope of this single link

https://github.com/busyorg/busy/issues/1529

A refresh work around here might conflict with changes needed for the issue above.

Sekhmet:
Is it solvable on the page level? If URL doesn't update there is no DOM update,

jm90m:
Thats true @Sekhmet there is no DOM update when URL doesn't change, I am not too sure if it is solvable on page level. Need to look into it.

kirkins:
@jm90m I found a fix that works for all the other cases, that is scroll to the top whenever the URL changes on the router.

Though I still have to keep the single instance of `onClick` reason being problem 1 in my original PR is a situation where a link is clicked but the URL stays the same.

Sekhmet:
It causes other problems though If you click tabs on user profile (Comments, Followers etc.) it scrolls back to top, which I don't think it desired.

kirkins:
@Sekhmet I see what you're saying but here is what @jm90m said:

> We should force the page to go to the top when we render the profiles pages (discussions, comments, wallet) and not when we click on the link. It's better to solve this in page level so we dont have to add onClick on every link of the website.

Then here is what you say:

> It causes other problems though If you click tabs on user profile (Comments, Followers etc.) it scrolls back to top, which I don't think it desired.

I'm trying to get approval from both you and @jm90m but you are asking for different things.

So please be patient with me on this. If you let me know the end-behavior from a user perspective I will do my best.

jm90m:
oh yeah sorry @kirkins we might have conflicting responses, lets make @bonustrack have the right response, I wrote my response based on my call with @bonustrack, and he wanted me to write it out, maybe I didn't convey our discussion more extensively. I'll tell him to respond to this.

Sekhmet:
@kirkins sorry, it felt unnatural to me so I pointed it out as a problem, but turns out it was the desired behaviour.
Sorry for the confusion.

PR #1541, opened 2018-02-16
Add auto-upvote setting to settings page

kirkins:
This is a follow up to https://github.com/busyorg/busy/pull/1482

I've added the upvote setting to the settings page.

It looks like these new fields will need translations. I tested locally and this change seems to work.

I had some problem with the test during commit so I had to use `-n`:

> Test Suites: 14 failed, 6 passed, 20 total
> Tests:       10 passed, 10 total
> Snapshots:   3 passed, 3 total
> Time:        22.483s
> Ran all test suites.

I'm not sure if the problem is with my setup of the commit. What is a test suite as opposed to a test?


jm90m:
screen shot 2018-02-18 at 9 17 42 pm
 btw this is what it looks like @bonustrack

src/client/settings/Settings.js
```diff
@@ -355,6 +366,26 @@ export default class Settings extends React.Component {
...
                   </div>
                 </div>
+                <div className="Settings__section">
+                  <h3>
+                    <FormattedMessage id="upvote_setting" defaultMessage="Automatically upvote" />
+                  </h3>
+                  <p>
+                    <FormattedMessage
+                      id="upvote_setting_details"
+                      defaultMessage="Enable this option to automatically upvote your own posts."
```
jm90m: This needs to be in `default.json` also as well as any new text. Check with @bonustrack for for text approval.

bonustrack: We should use `like` instead of `upvote`, similar than the others place in the app. Also i would change the title to `Like my posts`. `Automatically upvote` feel confusing, this title doesn't say that it's only in your own post.

src/client/settings/Settings.js
```diff
@@ -187,6 +197,7 @@ export default class Settings extends React.Component {
   handleShowNSFWPosts = event => this.setState({ showNSFWPosts: event.target.checked });
   handleRewriteLinksChange = event => this.setState({ rewriteLinks: event.target.checked });
   handleUseBetaChange = event => this.setState({ useBeta: event.target.checked });
+  handleUpvoteSettingChange = event => this.setState({ upvoteSetting: event.target.checked });
```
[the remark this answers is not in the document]
kirkins: Interesting read, relatively new to React but really like it so far. Let me if the change is what you had in mind.

I could convert the rest of the functions either in this PR or another, whatever is best.

jm90m:
looks and works fine for me @bonustrack 
thanks @kirkins  👍

Sekhmet:
`defaultMessages`'s and text in default.json now differ, but other than that it looks good to me. Thanks.
Can you update them, so they are the same in both files?

PR #1428, opened 2018-01-30
Support post links without category

Sekhmet:
Fixes #1410 

Changes:
- Support post links without category.
- Replace URL with link with category.

PR #1542, opened 2018-02-19
Remove unneeded history push

### LINKED ISSUE #1495: Navigation from Feed to go to full post by clicking the `comments` icon, and hitting back button is not working as intended
#### Expected behavior
When clicking on the comments icon, on the feed, you will be taken to the full post page and the comments section, then when you click the 'Back' button on your browser it should take you back to previous page.

#### Actual behavior

The url remains the same and you don't navigate to the previous page

#### How to reproduce
1. Go any feed and click on comments icon within the post (https://staging.busy.org/)

screen shot 2018-02-07 at 9 39 39 pm

2.  Try to click  back on browser to navigate back to previous page
feb-07-2018 21-43-11

* Browser: Chrome
* Operating system: El Capitan OSX

jm90m:
Fixes #1495

Changes:
- The window.history.pushState in render method of Post which was causing back button on browser to not work

I am not sure why this was needed, but it was causing back button to fail and I was trying to go through the conversations on here (https://github.com/busyorg/busy/pull/1428) to see if there were any edge cases that I was missing but it seems to be working.


Sekhmet:
It was there because we should use a short link (without category) by default.
Now there are two problems:
- We don't drop category for Twitter URL and direct link:
https://github.com/busyorg/busy/blob/97fcb77851729711587ea5bb7c14229ec7fd6164/src/client/post/PostModal.js#L84-L86
https://github.com/busyorg/busy/blob/97fcb77851729711587ea5bb7c14229ec7fd6164/src/client/post/PostModal.js#L115-L117
- We don't have a client-side redirect from long link to short one (we still should have this logic, but probably on componentDidMount).

jm90m:
ahh I see @Sekhmet, I updated it but I don't think its ideal, since when you navigate from the feed and click on the `comments` icon, you're able to press back on the page but you would need to press back twice to go to the current feed since it is modifying `window.history`.. 

I think the ideal solution would be to go around either change all links to drop category or have a better client side redirect logic (maybe in router? idk). what do you think?

Sekhmet:
Would it work with `replaceState` instead?

jm90m:
thanks @Sekhmet, replaceState seems to have worked.

PR #1544, opened 2018-02-19
Add Notifications on Client Side

### LINKED ISSUE #1120: Notifications UI
Create Notifications UI

jm90m:
Fixes #1120 .

Changes:
- Re-add notifications popover
- use busyjs for client


src/client/components/Navigation/Topnav.js
```diff
@@ -30,14 +32,62 @@ class Topnav extends React.Component {
...
+    notifications: PropTypes.arrayOf(PropTypes.shape()),
     searchAutoComplete: PropTypes.func.isRequired,
     onMenuItemClick: PropTypes.func,
+    onNotificationClick: PropTypes.func,
+    onSeeAllClick: PropTypes.func,
   };
 
   static defaultProps = {
     autoCompleteSearchResults: [],
+    notifications: [
```
bonustrack: Why we need this data here? Can we remove it?

jm90m: yeah was test date

bonustrack:
Quick review:
- Make the whole notification item clickable
- Add time ago
- Don't add permlink in the notification label
- Change follow text to `username started following you`
- Change mention texts to `username mentioned you in a comment` and  `username mentioned you in a post`
- Change reply texts to `username commented your post` and `username replied on your comment`
- Store in SC2 user metadata the last `timestamp` seen by the user **when he open the notification popover**. When user load the page, load the notifications aswell and calculate the unread count using the timestamp of last seen stored on SC2 metadata.
- Show unread notifications with light purple background (like currently), and the already seen with normal background.
- Make the popover appear over the topnav (so the arrow of the popover is not hidden)
image
- Remove space between notification icon and avatar in topnav
- Make this smaller:
image
- When there is more than 99 notifications, show `99+` counter
- Use black color for usernames

src/client/locales/default.json
```diff
@@ -32,10 +32,13 @@
   "bookmarks": "Bookmarks",
   "bookmarks_empty": "You don't have any story saved.",
   "notifications": "Notifications",
-  "notification_following_username": "{username} is now following you.",
-  "notification_mention_username_post": "{username} mentioned you on this post {post}.",
-  "notification_reply_username_post": "{username} replied you on your {post}.",
+  "notification_following_username": "{username} started following you.",
```
bonustrack: Don't use dot `.` in the end of these sentences.

bonustrack:
I dont see `username started following you` sentence

src/client/locales/default.json
bonustrack: Also I dont see this sentence been used
image

bonustrack:
- `See All` should be changed to `See all` and there should be a top border grey (on the red line level)
image
- The notification message should be grey color, and username in black and bold (similar than an activity label)

pareshUD:
When you click on see all notification it takes you no where..

src/client/locales/default.json
jm90m: you don't see those sentences being used because it goes to the en-US.json file, do you want me to change it there as well? I remember you guys telling me not to change it there. @bonustrack

kirkins:
Nice! Really looking forward to this feature.

bonustrack:
Looking better an better, we cant merge this PR yet we need to fix this issues on `busy-api`:
- Memory leak that make the server restart
- Dont store duplicate notification, when someone edit a post i may receive 2 time the same notification, for some reason a also see some duplicate follow notification.

bonustrack:
- [ ] Align all the center content to the red line not only the title!
image
- [ ] Dont show `You dont have notification` if it's loading, show loading icon instead
image
- [ ] If i open the notification popover fast before notifications get loaded it's buggy, i'm not able to see notification anymore after that.

jm90m:
@bonustrack definitely looks aligned to me lol, but I guess it can be moved over 1px?
screen shot 2018-02-21 at 9 16 35 pm

jm90m:
trying to reproduce your third issue @bonustrack, but I can't, I even slowed down my internet to average thailand speed...
screen shot 2018-02-21 at 9 41 30 pm

bonustrack:
@jm90m this is not aligned image

bonustrack:
Here is how to reproduce last bug:
bug

bonustrack:
You need to refresh the page a click on the notification icon very quick

jm90m:
thanks, @bonustrack, I cant reproduce it but I think I know the root cause, even tried it without console open + re-enable cache.

bonustrack:
Also, don't store new timestamp on SC2 metadata if user have a higher timestamp already saved

src/client/components/Navigation/Topnav.js
```diff
@@ -98,9 +135,14 @@ class Topnav extends React.Component {
...
   menuForLoggedIn = () => {
-    const { intl, username } = this.props;
-    const { searchBarActive } = this.state;
-    const { popoverVisible } = this.state;
+    const { intl, username, notifications, userSCMetaData, loadingNotifications } = this.props;
+    const { searchBarActive, notificationsPopoverVisible, popoverVisible } = this.state;
+    const lastSeenTimestamp = _.get(userSCMetaData, 'notifications_last_timestamp');
+    const notificationsCount = _.isUndefined(lastSeenTimestamp)
+      ? _.size(notifications)
+      : _.filter(notifications, notification => lastSeenTimestamp < notification.timestamp).length;
```
Sekhmet: We use two different things to get basically the same thing (`_.size`) and `.length`). I think we should use one of those methods for both cases.

jm90m: yeah you're right, I prefer to`_.size` since it has a bunch of polyfills + error checks that accompany it. will update now

bonustrack:
Approved @jm90m

PR #1547, opened 2018-02-21
Update version to 2.4

bonustrack:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "busy",
-  "version": "2.3.0",
+  "version": "2.4.0",
```
jm90m: whats the difference between 2.3.0 and 2.4.0?

Why is it not 2.3.1? Or 2.3.2 (go up by increments of .1)

We have 3 places for numbers and it can be incremented. Is there any special significance of it being incremented in the 2nd place rather than the 3rd?

jm90m: :trollface:

bonustrack: Because it look better :trollface:

PR #1570, opened 2018-02-26
Add ctrl/cmd + click check for when notifications linked is clicked

jm90m:
Add ctrl/cmd + click check for when notifications linked is clicked


bonustrack:
There isn't a more natural way to do that? Links should be openable by default without having to code keylogger.

jm90m:
they are openable on default ? Its just @espoem wanted it so notifications popover doesnt get closed when opened with new tab click

bonustrack:
Links are supposed to be openable with ctrl click by default. It's the case on every website, it's just the default behavior. Why do we need to handle it with JS?

jm90m:
So we can see if its a `cmd` or `ctrl` click so notifications popover doesn't close when they do click on it.

PR #1591, opened 2018-03-04
Fix Bug Where p and h2 Tags Are Rendered in Code Tags

### LINKED ISSUE #1579: HTML code <p> <h2> bug
https://busy.org/@bitcoin-journal/steem-vs-ripple-a-comparative-guide

image

caleblogan:
Fixes #1579 

Changes:
- Keep markdown renderer from adding code tags to html when tags are indented

The tags that were formatted wrong were indented and had an extra blank line so the renderer added code blocks around them. I took what condenser did to keep consistency. 

This may be considered a bug on condensers side? They allow indenting of html elements if you start the post with an html tag or p tag. When Busy renders the post that was made on condenser, it passes the indented tags to the renderer. This adds the code blocks, malforming the post.

I think how busy is rendering it is correct.
Here is where condenser skips remarkable if starting with p or html tag: https://github.com/steemit/condenser/blob/master/src/app/components/cards/MarkdownViewer.jsx#L103

## Busy Example:
busy_md_example
 
## Steemit Example:
steemit_md_example


Sekhmet:
Thanks for taking a look at this issue. I analysed your findings, and I think that opting-out of our rendering because of the fact that it's already HTML is not worth it.

I tested removing tabs before HTML tags and it seems to work fine with example post from the issue:
```diff
diff --git a/src/client/components/Story/Body.js b/src/client/components/Story/Body.js
index a04ad205..c39cbb9c 100644
--- a/src/client/components/Story/Body.js
+++ b/src/client/components/Story/Body.js
@@ -28,6 +28,8 @@ export function getHtml(body, jsonMetadata = {}, returnType = 'Object', options
 
   let parsedBody = body.replace(/<!--([\s\S]+?)(-->|$)/g, '(html comment removed: $1)');
 
+  parsedBody = parsedBody.replace(/\t+</g, '<');
+
   parsedBody.replace(imageRegex, img => {
     if (_.filter(parsedJsonMetadata.image, i => i.indexOf(img) !== -1).length === 0) {
       parsedJsonMetadata.image.push(img);
```

Do you think it would solve the problem?

caleblogan:
Yeah that should work. Looks a lot better. Should we also we also strip whitespace and not just tabs?

caleblogan:
If we don't strip spaces a user may write a post on condenser indenting with spaces. This will add code blocks on busy if we just strip tabs. Stripping spaces may affect old posts made on busy. Where a user used spaces to add code blocks to html. Not sure if anyone actually uses spaces instead of `.

caleblogan:
If we don't strip spaces and a user creates a post like the example above it will add code blocks around the html on busy. Only problem with stripping spaces is it might malform old posts if a user used spaces to add code blocks to html.

Sekhmet:
If it causes the same problem I think so.

caleblogan:
I made this test post with spaces on steemit: https://steemit.com/test-category/@okc/test
By stripping spaces it will appear correctly on busy.

Sekhmet:
I think it might remove spaces that are just spaces, like there: `Hello <b>World</b>`.
If I understand correctly, only whitespace that is at the beginning of the line causes this problem. If so something like this should handle this better:
`/^\s+</gm`

caleblogan:
Good catch. That regex works for me.

src/client/components/Editor/Editor.js
```diff
@@ -145,8 +145,9 @@ class Editor extends React.Component {
   }
 
   setBodyAndRender(body) {
+    const parsedBody = body.replace(/^\s+</gm, '<');
```
Sekhmet: Do we need it there? `bodyHTML` is passed to `Body` component that already strips that whitespace.

caleblogan: You're right. I must have left the old regex in the body when I was testing the editor preview. Didn't notice `Write` called `getHtml` before passing the body to the Editor.

PR #1594, opened 2018-03-04
Fixed Header from Moving When a Modal is Opened

### LINKED ISSUE #1554: Fix Top Navigation from shifting every time a modal is opened
#### Expected behavior
Every time a Modal is opened, the top navigation should remain in the same spot and not move.

#### Actual behavior
Top navigation header shifts awkwardly.

#### How to reproduce
- open any modal (Post modal, reblog modal, login modal...)

* Browser: Chrome
* Operating system:  OSX

caleblogan:
Fixes #1554 

Changes:
- Use vw instead of % on Layout.Header

## Explanation
Antd adds padding to the body when opening a modal. This keeps the body from shifting when scrollbars disappear. The header is fixed in this situation so it doesn't receive the padding from the body.

I'm not too sure if vw is bad to use for cross platform compatibility.


bonustrack:
The topnav is a bit off now, the red lines should be aligned to green ones:
image

caleblogan:
I added scrollbar to header to push it back over to the right positions. This does show when opening the modal. Its not really noticeable.

### Other possible solutions:
- Get the the scrollbar width with javascript and add it to the header
- Add the scrollbar to the body even when the modal is opened and override the padding antd adds to the body when a modal is opened.

Let me know what you think.

jm90m:
nice fix, does this work for you @bonustrack, everything is aligned

jm90m:
oops didnt mean to close but anyways, here's what I see

screen shot 2018-03-08 at 2 07 15 pm

Sekhmet:
Looks good to me.

bonustrack:
Now it's looking fine with this grey background with 3% of opacity but if we change background at some point it could look weird, for example:
image
Why not adding the scrollbar for the whole page?

Also the header is moving a little (1px) when i open a modal / close it

caleblogan:
I tried adding the scrollbar to the body permanently, but that was allowing the background page to scroll when a modal was open. I just added scroll bar to all modals and it seems to do the trick.

bonustrack:
Can you see the page moving 1pixel away multiple time when opening a modal? It's not so obvious but it here and it was not here before.
pixel

caleblogan:
That's a bit weird, its not happening for me. The only thing I changed was adding the scrollbar to the modal background overlay. Maybe @jm90m has an idea what could be causing it? Are you in landscape mode by any chance?

jm90m:
@caleblogan Not too sure what could be causing this, but I just took a look at this but it could have always been there from your original fix of changing the unit to 100vw or adding the scrollbar, it not really hard to notice but its there

Sekhmet:
JM worked on this issue in #1649.
It had some problems so I fixed it in #1700. I though that this PR has been already merged

PR #1597, opened 2018-03-05
Fix reply to comment with uploaded image bug

### LINKED ISSUE #1596: Replying to a comment with an image, inserts image on post comment form
If you try to use the upload an image button on a reply to a comment it will upload the image and then insert the markdown in the input form to reply to the original post.

The user would expect that the image should be inserted into the comment reply input box instead.

untitled

kirkins:
Fixes #1596

The problem here seemed to be that when you multiple instances of `EditorInput` on a page you end up with multiple elements with the same ID.

So when you click on the upload button it associates itself with the first instance of `EditorInput` on the page.

To fix this I removed the use of ID and instead embed the input within the label.


kirkins:
I guess the linter is complaining about having a label without an ID. When I Google it I get [this issue about partially disabling it](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/302).

As mentioned in this case having an ID seems to cause an issue, and using nested instead seems to fix it on my end.

Let me know what you think should be done about the lint issue.

Sekhmet:
I don't really like disabling rules that are somehow justified.
It might make sense to pass `id` prop to `EditorInput` and we can use it as part of an ID.

kirkins:
@Sekhmet that sounds like it would work too.

So If I add an `inputId` prop to the `EditorInput` file, I would then just pass an inputId value here:

https://github.com/busyorg/busy/blob/b21cfe38c8bb8dcbf17ce360aa48241c93f28d62/src/client/components/Comments/CommentForm.js#L110-L117

I'm not sure how I should generate a unique ID for this?

Sekhmet:
If it is reply or edit we can use something like this:
`edit-${postId}` or `reply-${postId}`. For others, we can use `post` or `post-footer`.

kirkins:
@Sekhmet thanks, that seemed to fix the lint issue.

Sekhmet:
Yeah, but now many forms don't have IDs (EditorInput is used in multiple places).

PR #1598, opened 2018-03-05
Keep PostRecommendation from Showing Posts from Negative Rep Authors

### LINKED ISSUE #1539: recommended posts include posts from low reputation users
#### Expected behavior

The recommended posts should not show posts from users with negative rep

#### Actual behavior

Posts are recommended from authors with negative rep

#### How to reproduce

Visit a post from a negative rep author who has multiple posts

caleblogan:
Fixes #1539 

Changes:
- Add filter to PostRecommendation for negative rep authors


src/client/components/Sidebar/PostRecommendation.js
```diff
@@ -73,6 +74,7 @@ class PostRecommendation extends Component {
   getFilteredPosts = () => {
     const { match } = this.props;
     return this.state.recommendedPosts
+      .filter(post => formatter.reputation(post.author_reputation) > -1)
       .filter(post => post.permlink !== match.params.permlink)
```
jm90m: you don't need to iterate through the list again, you can create 1 filter function above that handles both filtering cases here

jm90m:
I am not too sure if #1539, is what we want, @bonustrack has final say on these types of things.

caleblogan:
Do you mean not adding this feature? Or something more like if the author has negative rep, suggest posts from one of the interesting authors?

jm90m:
@caleblogan yeah not adding this feature, when people create issues for features like this, @bonustrack has to approve of it

caleblogan:
Yeah that makes sense.

jm90m:
Thanks for the contribution, me and @bonustrack talked about it, and we have approved this.

PR #1600, opened 2018-03-05
Remove busy-bot ad

### LINKED ISSUE #1560: Removing bot advertisement - Suggestion
#### Expected behavior
Remove or modify the text advertising the bot, the vote should be considered as a possible free bonus

#### Actual behavior

Seems to guarantee a vote if someone publish using busy, bringing complaints when vote does not occur

* Browser: All
* Operating system: All

### LINKED ISSUE #1692: Deleting drafts from editor doesn't work
#### Expected behavior

After deleting draft from editor draft should be gone.

#### Actual behavior

It still appears on drafts page (it's still stored).

#### How to reproduce

Open draft in editor, click delete draft and confirm. It's still there in drafts page.

Sekhmet:
Fixes #1560
Fixes #1692 

Changes:
- Remove busy-bot ad
- Add SingleLayout.


bonustrack:
There is a strange behavior, when you type and the draft get saved the editor input will get back to the saved version even if i added more text on it:
bug

Sekhmet:
I couldn't reproduce this problem. Do I need particular steps to reproduce this?

bonustrack:
@Sekhmet write something on the editor until the auto save is triggered, then add more text white the save is going on. When the save will be completed, the additional text will be removed and back after few seconds:
test

bonustrack:
Looks better now, i cant reproduce the previous bug anymore.

I have another small change. Can you make that we have a title even if user didnt set a title? Similar than the page drafts. This would avoid the case where we see only date, its a bit confusing
image

Sekhmet:
Don't merge it for now.
I've noticed few problems with current implementation of Write page (which for example makes loading saved upvote setting without ugly workarounds hard to implement). I'm updating this right now, so I will request another review when it's done.

Sekhmet:
I did some refactoring in 31e6b83f43290bebbef688dbaa6d3882b490e70f
Can we review it again?

src/client/components/Sidebar/LastDrafts.less
```diff
@@ -0,0 +1,29 @@
+@import (reference) '../../styles/custom.less';
+
+.LastDrafts {
+  &__title {
+    color: #99aab5;
```
jm90m: colors could be in constants

src/client/components/Sidebar/LastDrafts.js
```diff
@@ -0,0 +1,67 @@
...
+import { FormattedMessage, FormattedRelative } from 'react-intl';
+import { Link } from 'react-router-dom';
+import Loading from '../../components/Icon/Loading';
+import './LastDrafts.less';
+import './SidebarContentBlock.less';
+
+const Draft = ({ draft }) => (
+  <div className="LastDrafts__draft">
+    <Link to={{ pathname: '/editor', search: `?draft=${draft.id}` }}>
+      {draft.title.length ? (
```
jm90m: this might break if `title` is anything other than a string or an array

PR #1612, opened 2018-03-08
Render Comments in Chunks WIP

### LINKED ISSUE #1413: Render comments in chunks

caleblogan:
Fixes #1413 

Changes:
- Add component for show more comments `MoreCommentsLink`
- Creates comment helpers for finding nodes

I wanted to get some feedback to see if this is on the right path. I noticed that the biggest lag issue is that each comment takes awhile to load. Even just rendering 20 comments takes almost 1 second. This seems too long. 

Here's a graph of rendering 20 top level comments (not children). The `Buttons` for each comment take anywhere from 15ms up to 70ms each to render. I don't have much experience with performance issues so I'm not sure.

busy_comments_flame

### DIFF
--- a/src/client/comments/Comments.js
+++ b/src/client/comments/Comments.js
@@ -154,29 +154,30 @@ export default class Comments extends React.Component {
       rewriteLinks,
     } = this.props;
     const postId = post.id;
-    let fetchedCommentsList = [];
+    let rootLevelComments = [];
 
-    const rootNode = comments.childrenById[postId];
+    const parentNode = comments.childrenById[postId];
 
-    if (rootNode instanceof Array) {
-      fetchedCommentsList = rootNode.map(id => comments.comments[id]);
+    if (parentNode instanceof Array) {
+      rootLevelComments = parentNode.map(id => comments.comments[id]);
     }
 
     let commentsChildren = {};
 
-    if (fetchedCommentsList && fetchedCommentsList.length) {
+    if (rootLevelComments && rootLevelComments.length) {
       commentsChildren = this.getNestedComments(comments, comments.childrenById[postId], {});
     }
 
     return (
-      fetchedCommentsList && (
+      rootLevelComments && (
         <CommentsList
           user={user}
           parentPost={post}
-          comments={fetchedCommentsList}
+          comments={comments.comments}
+          rootLevelComments={rootLevelComments}
+          commentsChildren={commentsChildren}
           authenticated={this.props.authenticated}
           username={this.props.username}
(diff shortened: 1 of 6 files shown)


bonustrack:
This is how it look like now:
image
https://busy-master-pr-1612.herokuapp.com/@steem-plus/steemplus-2-5-integration-of-steemit-more-info-features
@pareshUD do you have a design for the button?

Sekhmet:
Thanks for taking look at this problem.
Buttons take a while to render because they are a little bit complicated and have logic inside.

Problem with testing performance in the development mode is that sometimes it's giving a wrong idea about what is wrong. I think that flame chart is mostly right, but when you are running lots of development code everything can take 10 times more time.

Despite that, current architecture doesn't scale well with posts that have lots of comments. To mitigate this issue we want to render at most X comments in one tree and ask if it should render more. This way it will render at most X comments once even though the post has 100 times more.

I think for such architecture it might make sense to use this endpoint instead: https://v2.steemconnect.com/docs/steemjs#api/get_replies_by_last_update
This way we don't fetch 500 comments if the user doesn't want to see them all.

bonustrack:
@pareshUD do you have a design for this? (second attempt)

bonustrack:
The problem with `get_replies_by_last_update` is that we would have only a sorting by date of comments and lose sorting by best, reputation etc. I think there is no ideal way atm to load comments in chunk from Steemd API, but at least we can display them in chunk.

bonustrack:
Can you use current primary color (purple) and use "sans serif" font @pareshUD ?
It would be good also to have a proposal on mouse over maybe with grey background.

bonustrack:
We should show only one button on bottom of the root comments with text "Load more comments" but not on nested reply. And for the style we can use @pareshUD last design with primary color for the text and sans serif font. Do you want to work on this @caleblogan ? Otherwise @jm90m @Sekhmet can take over the PR.

caleblogan:
Sorry, I currently don't have time.

bonustrack:
@Sekhmet will take over this PR

PR #1618, opened 2018-03-08
Fix bug where bullet was being displayed for unknown apps

### LINKED ISSUE #1617: Bullet should not appear next to post date if app is not recognized
This bullet:
image
Should not appear in single post or feed if there is nothing next to it (when application is not recognized on our apps.json list)

Go here to reproduce that issue: https://busy.org/@dodybireuen/busy-org-the-best-alternative-interface-for-exploring-the-steem-world

caleblogan:
Fixes #1617 

Changes:
- return null from component if app is not found
- extracted app parsing from `PostedFrom` component and put it in a helper

### DIFF
--- a/src/client/helpers/postHelpers.js
+++ b/src/client/helpers/postHelpers.js
@@ -23,21 +23,18 @@ export function dropCategory(url) {
  * Gets app data from a post.
  * Only Returns app info from apps whitelisted in apps.json
  * @param post
- * @returns {*}
+ * @returns An empty object if app is not valid otherwise an object with {appName: String, version: String}
  */
 export function getAppData(post) {
-  let appName;
-  let version;
-  try {
-    const app = jsonParse(post.json_metadata).app.split('/');
-    if (whiteListedApps[app[0]]) {
-      appName = app[0];
-      version = app[1];
-    }
-  } catch (e) {
-    return {};
+  const [appKey, version] = _.split(jsonParse(post.json_metadata).app, '/');
+
+  if (whiteListedApps[appKey]) {
+    return {
+      appName: whiteListedApps[appKey],
+      version: version || '',
+    };
   }
-  return { appName, version };
+  return {};
 }
 
 export default null;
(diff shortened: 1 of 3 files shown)


src/client/helpers/postHelpers.js
```diff
@@ -17,4 +19,25 @@ export function dropCategory(url) {
...
+ * Gets app data from a post.
+ * Only Returns app info from apps whitelisted in apps.json
+ * @param post
+ * @returns {*}
+ */
+export function getAppData(post) {
+  let appName;
+  let version;
+  try {
+    const app = jsonParse(post.json_metadata).app.split('/');
```
jm90m: this will fail if `.app` is undefined, and this will mostly be the case, since so many apps on steem blockchain are popping up and they might not include the app meta data, you should probably do `_.split()` here.

src/client/helpers/postHelpers.js
```diff
@@ -17,4 +19,25 @@ export function dropCategory(url) {
...
+    if (whiteListedApps[app[0]]) {
```
jm90m: you can do `const appName = _.get(app, 0, '')`

src/client/helpers/postHelpers.js
```diff
@@ -17,4 +19,25 @@ export function dropCategory(url) {
...
+      appName = app[0];
+      version = app[1];
```
jm90m: same with version here

src/client/helpers/postHelpers.js
caleblogan: I think in this case it would be ok, because of the try/catch.  Ill refactor using lodash as you suggested and remove the try/catch. Should make cleaner without the try/catch.

bonustrack:
It would be great also if we can remove the tooltip "Version *" if there is not version set, this happen with Steemhunt example: https://staging.busy.org/@urbangladiator/kittyhats-hats-apparel-and-accessories-for-your-cryptokitties

PR #1621, opened 2018-03-09
Usability with slow internet

### LINKED ISSUE #1576: troubleshoot problem, need help (busy slowness)
#### Expected behavior

Browsing fast on busy

#### Actual behavior

User has terrible delays while opening busy and staging, we cleared cache and rebooted machine, but issue remains. username : @dodybireuen 

#### How to reproduce
cannot reproduce, I do suspect backbone bottleneck between him and busy, o I think it will self resolve, but if you have an idea ... I welcome it
* Browser: all
* Operating system: all

kirkins:
I'm having an issue that sounds similar to #1576.

Unfortunately my internet is slow right now.

This has led to find that both Busy client and server have a maximum timeout of 5 seconds. 

screenshot from 2018-03-08 19-05-17

[Related issue for rpc.](https://github.com/busyorg/lightrpc/issues/6)

### DIFF
--- a/src/client/steemAPI.js
+++ b/src/client/steemAPI.js
@@ -1,6 +1,12 @@
 import { createClient } from 'lightrpc';
 
-const client = createClient(process.env.STEEMJS_URL || 'https://api.steemit.com');
+const options = {
+  timeout: 15000,
+};
+
+const steemUrl = process.env.STEEMJS_URL || 'https://api.steemit.com';
+
+const client = createClient(steemUrl, options);
 client.sendAsync = (message, params) =>
   new Promise((resolve, reject) => {
     client.send(message, params, (err, result) => {


Sekhmet:
We for sure don't want to increase the timeout on the server to 15 seconds (as it will cause a request to take up to 15 seconds to finish).

kirkins:
@Sekhmet if it is "up to 15 seconds" I think that is fine for anyone with slow Internet.

If it is minimum 15 seconds that will be a problem. Is this maximum 15 seconds or minimum?

The problem with 5 seconds is that busy.org becomes completely unusable for anyone with a slow internet connection. If I switch to Steemit.com instead on slow internet it will takes a while, over 5 seconds but it eventually loads.

With this current timeout Busy just says if you can't load a request in 5 seconds you can't use the site.

There are a lot of people using Steem in developing countries so I'm sure this issue is extremely common.

Sekhmet:
Let's say api.steemit.com has some problem and won't finish request that we make on the server -> user is waiting for 15 seconds for Busy.org to respond.
Client timeout can be changed, but I don't think we should keep hundreds of open connections on the server, just because we are waiting for API to suddenly come back to life.

kirkins:
@Sekhmet I see the server part is just concerning communication between Busy and Steem API?

I didn't realize that, I'll have to do some more testing. The PR as it is now doesn't improve the situation for me at all.

Sekhmet:
If you want to change timeout on the client only you can do it here:
https://github.com/busyorg/busy/blob/cd3b1a32e2921ae2cd763bae9fee541afe540ef4/src/client/steemAPI.js#L3

kirkins:
@Sekhmet thanks, that fixes it for me.

PR #1639, opened 2018-03-13
Add ... to meta description to avoid akward cut-offs

kirkins:
This is just a small change for those who share articles from Busy to Twitter/Facebook/ect.

It looks weird when the word gets cutoff. I was thinking I made a typo. Also I'm thinking it could be awkward if the last word gets cut-off and makes it seem like the article says something it doesn't. _(Example: Dickens being cut to Dick, or anything else like that)_

I wasn't actually able to fully test this as it doesn't render on my `localhost:3000` copy of the site.

untitled

_Also I noticed the Twitter card uses the Steemit account on Twitter. I've updated it to the busyorg account._


src/client/post/PostContent.js
```diff
@@ -230,7 +230,7 @@ class PostContent extends React.Component {
           <meta property="article:tag" content={category} />
           <meta property="article:published_time" content={created} />
           <meta property="twitter:card" content={image ? 'summary_large_image' : 'summary'} />
-          <meta property="twitter:site" content={'@steemit'} />
+          <meta property="twitter:site" content={'@busyorg'} />
```
jm90m: good catch on this

src/client/post/PostContent.js
```diff
@@ -207,7 +207,7 @@ class PostContent extends React.Component {
     const postMetaImage = postMetaData.image && postMetaData.image[0];
     const htmlBody = getHtml(body, {}, 'text');
     const bodyText = sanitize(htmlBody, { allowedTags: [] });
-    const desc = `${bodyText.substring(0, 140)} by ${author}`;
+    const desc = `${bodyText.substring(0, 140)}... by ${author}`;
```
jm90m: So it would be better if you created a helper function in one of our helper files that does string truncation. 

Right now if the bodyText is under 140 characters then it still adds an ellipsis, and that should not be the case.

kirkins:
Prettier is working locally for me on this, but fails on circle.

jm90m:
hey @kirkins, we had to revert the commit, since it was failing the build. Sorry my fault, I tested old build and it worked. But new build with your latest changes, fails the build.

kirkins:
@jm90m that is really strange.

How do I reproduce the build error, should I just run `npm run build:client`?

I seem to get the same results with that command for my branch and master.

kirkins:
@jm90m I can't reproduce because even when I run those commands with a fresh master I get:

```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: Unexpected server response: 503
    at ClientRequest._req.on (/tmp/busy/node_modules/busyjs/node_modules/ws/lib/websocket.js:645:21)
    at emitOne (events.js:116:13)
    at ClientRequest.emit (events.js:211:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:551:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:116:23)
    at TLSSocket.socketOnData (_http_client.js:440:20)
    at emitOne (events.js:116:13)
    at TLSSocket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
[nodemon] app crashed - waiting for file changes before starting...
```

Do you have any idea why it broke the build?

jm90m:
@kirkins probably this entire function - https://github.com/busyorg/busy/pull/1639/files#diff-3790f296954d7550bfd6609c20a66551R10

Try using lodash for some of the operations you're doing here.

PR #1659, opened 2018-03-20
Update antd popup messages z-index

### LINKED ISSUE #1658: Error message should be over post modal
When i open a post modal and i try to  comment twice in a small laps of time to trigger the error "You can comment only every 20sec", the error message will not be visible because the post modal go over it. We should bring the messages success or error over post modal.

jm90m:
Fixes #1658 .

Changes:
- update global ant-messages to always be over modals

### DIFF
--- a/src/client/styles/custom.less
+++ b/src/client/styles/custom.less
@@ -88,6 +88,7 @@
 @modal-background-color: rgba(242, 242, 242, 0.97);
 
 // z-indexs
+@ant-message-z-index: 1400;
 @react-modal-z-index: 1400; // used for post image gallery
 @ant-layout-header-z-index: 1050;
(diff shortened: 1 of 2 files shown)


src/client/styles/custom.less
```diff
@@ -88,6 +88,7 @@
 @modal-background-color: rgba(242, 242, 242, 0.97);
 
 // z-indexs
+@ant-message-z-index: 1400;
```
Sekhmet: What happens if we have both message and image gallery open at the same time?

jm90m: good question, I didn't know if image gallery should be behind or infront of the message, so i just set it the same

PR #1686, opened 2018-03-28
lodash for simpler version of commit #1639, which failed build

kirkins:
Hi, this is a follow up to the build issue from #1639 

Thanks to @fel1xw for providing solution in comments [here](https://busy.org/@kirkins/busy-org-improve-meta-description-for-sharing-on-social).


kirkins:
Just another small thing. I'm thinking it might be good to remove the ` - Busy` from the title.

Based on this example:

untitled

Anyone else think that would look slightly better with just the post title?

fel1xw:
It looks like a good an easy improvement to the social meta tags! Can you squash your commits so the git history stays clean? There is a great description to this here: https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

kirkins:
@fel1xw I use to squash on github but I don't anymore. This is for these reasons. 

1. Maintainer has option to squash all commits in a PR once they accept it, there is a button for this built into github.

2. When you delete the commits from the repo it changes the history on the pull-request. This can be confusing and if someone made inline comments on the code of a commit they get deleted from the site.

3. It re-triggers the build system pointlessly.

4. You can easily see the combination in the "Files changed" tab.

I agree these commits don't add anything but above are the reasons I don't squash in general without a good reason.

If the maintainer requests I will do it but otherwise I prefer not to.

jm90m:
deployment looks fine, and code looks fine though, what do you think @Sekhmet ?

PR #1697, opened 2018-04-02
filter object as tag edge case

kirkins:
I'm in the process of trying to [fix this issue](https://github.com/steemit/condenser/issues/2693) on Steemit.com.

Tested on Busy and the fix was a lot easier, just one filter.

Basically if a developer makes a post with an object as the tag it breaks the gui.

[Example of broken page](https://busy.org/@ikidnapmyself/my-awesome-title)


Sekhmet:
If we want to handle edge cases we shouldn't only handle arrays of objects.
What if tags is a number or a string. This will break page as well.

src/client/components/Story/StoryFull.js
```diff
@@ -476,7 +476,11 @@ class StoryFull extends React.Component {
             style={{ width: '100%', height: 46 }}
           >
             <div className="StoryFull__topics__content">
-              {tags && tags.map(tag => <Topic key={tag} name={tag} />)}
+              {tags.constructor === Array
```
jm90m: I think you can simplify this with lodash:

`_.map(tags, tag => _.isString(tag) ? <Topic key={tag} name={tag}  /> : null)`

Sekhmet: I think that it would be better to filter non string elements instead of returning array of `null` items.

jm90m: yeah good point might be better to use _.reduce and do something like this.....
```
    _.reduce(
      _.uniq(tags),
      (tagsArray, tag) => {
        if (_.isString(tag)) {
          return _.concat(tagsArray, <Topic key={tag} name={tag} />);
        }
      },
      [],
    );
```

(Added the _.uniq because I noticed some posts have duplicate tags lol)

Sekhmet: Wouldn't something like this work the same?

`_.uniq(tags).filter(_.isString).map(tag => <Topic key={tag} name {tag} />)`

kirkins: @jm90m thanks, tested and works but I'm having trouble with this lint error:

````
/tmp/busy/src/client/components/Story/StoryFull.js
  481:34  error  Expected to return a value at the end of arrow function  consistent-return
````

Sekhmet: It should work if you add `return tagsArray` after `if`.
I think that using regular filter instead of recreating it using reduce would be better solution.

jm90m: @Sekhmet yeah that would work the same, was trying to only iterate through the array once :P

PR #1710, opened 2018-04-06
Change whitelisted applications to Steemscript instead of internal list.

hernandev:
Changes:
- Added https://github.com/busyorg/steemscript as a dependency.

This dependency was added directly from GIT because the NPM release is 2 years old.

- Replaced `client/helpers/apps.js`

Instead of an internal list, use the one on the previous topic to build the user-friend application names.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     "secure-random": "^1.1.1",
     "socket.io": "^1.4.5",
     "speakingurl": "^11.0.0",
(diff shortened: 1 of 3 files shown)


Sekhmet:
Thank you.
I don't think those comments are that useful.
The code is pretty self-explanatory, so we probably should get rid of them.,

hernandev:
I can remove those comments @Sekhmet no problem

hernandev:
Done.

PR #1720, opened 2018-04-10
Add Last Drafts back to Editor Sidebar

jm90m:
- Add last drafts component to right sidebar on editor


bonustrack:
There is still an issue going on that make user lost draft when switching from one to another while draft being saved.

bonustrack:
Seem to be solved by now, i'm not able to lose a draft anymore. What do you think @Sekhmet ?

PR #1722, opened 2018-04-10
Don't render non-string value for app

### LINKED ISSUE #1721: Posts from new Parley site break Busy
Not sure what is causing this page to break:

https://busy.org/@zer0hedge/20180410t170926845z-post

Another example:

https://busy.org/@taskmaster4450/20180408t203636527z-post

Here is the console log when running the page locally:

```
The above error occurred in the <PostContent> component:
    in PostContent (created by Connect(PostContent))
    in Connect(PostContent) (created by Post)
    in div (created by Post)
    in div (created by Post)
    in div (created by Post)
    in div (created by Post)
    in Post (created by Connect(Post))
    in Connect(Post) (created by Route)
    in Route (created by Wrapper)
    in Switch (created by Wrapper)
    in div (created by Wrapper)
    in div (created by BasicLayout)
    in BasicLayout (created by Adapter)
    in Adapter (created by Wrapper)
    in LocaleProvider (created by Wrapper)
    in IntlProvider (created by Wrapper)
    in Wrapper (created by Connect(Wrapper))
    in Connect(Wrapper) (created by Route)
    in Route (created by withRouter(Connect(Wrapper)))
    in withRouter(Connect(Wrapper)) (created by Route)
    in Route
    in Switch
    in Switch
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by AppHost)
    in AppHost
    in AppContainer
    in Provider

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

kirkins:
Fixes #1721

I'm just filtering out non-string values for the app here.

What do you think of that as a solution?

Should we filter the name and version object values? I don't know if any other apps are using this format.


Sekhmet:
It's not a standard format, so we might ignore that as for now.

jm90m:
Might be better to use lodash indexOf, that way we aren't reassigning an object property

Just saw that you did this, but I made a PR for it but I closed it - https://github.com/busyorg/busy/pull/1734/files

kirkins:
@jm90m ok, I'm applying the changes from your closed pull request.

I'm not sure how to use lodash indexOf to check if an attribute is a string.

I see there is `if(_.isString(myVar))`.

Feel free to edit this PR directly, I will take a look at how you do it and try to remember for next time. Trying to use lodash more.

src/client/post/PostContent.js
```diff
@@ -181,11 +181,16 @@ class PostContent extends React.Component {
     const postMetaData = jsonParse(content.json_metadata);
     const busyHost = appUrl || 'https://busy.org';
     let canonicalHost = busyHost;
-    if (postMetaData.app && postMetaData.app.indexOf('steemit') === 0) {
+
+    if (_.isString(postMetaData.app)) {
```
jm90m: you don't need this check if you're using` _.indexOf`

src/client/post/PostContent.js
```diff
@@ -181,11 +181,16 @@ class PostContent extends React.Component {
...
+      postMetaData.app = '';
+    }
+
+    if (_.indexOf(postMetaData.app, 'steemit') === 0) {
```
jm90m: this should work just fine, I don't think you'll need the _.isString check

src/client/post/PostContent.js
Sekhmet: It looks like this condition should be negated.

It shouldn't matter though if indexOf from lodash can handle non-strings

Issue #1685, opened 2018-03-28
i cant edit my post.maybe post with too many text is occured problem


mlnl17:
#### Expected behavior
when i touched “edit post”,should move in to edit page.

#### Actual behavior
nothing happen.just load symbol spining around

#### How to reproduce
https://busy.org/@coolzero/3-22-coinmarketcap-cmc
other post are ok.just this one.
and other platforms are no problem ocuur.
in steemit.com,i can edit.
even in steemkr.com,i can edit.

maybe too many text in post is problem?

* Browser: safari
* Operating system: ios iphone6s plus

mlnl17:
https://imgur.com/gallery/KTymU
made gif file.just spinning,nothing happen

Sekhmet:
Problem is that you have limited space for your data on SteemConnect and if you have too many (or too big) posts it won't let you create new ones. We should display error message with description in that case.

dakeshi:
I have same issue on chrome browser. I wrote a posting which a quite long and then could not edit it.

![busy_org_edit](https://user-images.githubusercontent.com/8563047/38288135-b3f309e2-3809-11e8-8c18-5642a50ef836.gif)

bonustrack:
💯 that would be great to have a specific error message when this issue happen. I know others users who got the same problem.

Sekhmet:
It might be more common for Korean posts because every letter counts as three bytes (instead of 1 for English letters).

Sekhmet:
Fixed in #1725

PR #1725, opened 2018-04-11
Add error message when editor fails to save

### LINKED ISSUE #1717: Display error message if SC2 metadata storage is full
If an user has too much drafts the edit feature fail. We should catch the error when this happen and display an error message for the user.

SC2 give an error 403 for this see: https://github.com/steemit/steemconnect/blob/76942c760a4625455f74f9f1a84e801236bbc293/routes/api.js#L49-L52

The error message proposal is : "Oops! You hit the storage limit of 16mb, delete some drafts to go forward".

jm90m:
Fixes #1717 .

Changes:
- display error message notify popup when drafts failed to save


Sekhmet:
I think it fixes #1685 as well.

PR #1727, opened 2018-04-11
Prevent tooltip to go outside of screen

### LINKED ISSUE #818: Post tooltip overflows at top of the page
image

image

#### Expected behavior

The tooltip should switch to bottom position when near the top of the screen.

#### Actual behavior

The tooltip overflows so some information may not be visible.

#### How to reproduce

1. Go to `/trending`
2. Scroll down so that the line with likes, comments and reward is at the top.
3. Hover over likes count or the reward.

* Browser: Chrome 61.0.3163.100

Sekhmet:
Fixes #818

### Changes
- Add `BTooltip` component that switches position based on a position of an element on the screen.
- Use `BTooltip` for votes previews.

### Screenshots

Before:
pre

After:
post


src/client/components/BTooltip.js
```diff
@@ -0,0 +1,61 @@
...
+    const { title, ...props } = this.props;
+    const { bottom } = this.state;
+
+    return (
+      <Tooltip
+        {...props}
+        title={<div ref={this.setTitle}>{title}</div>}
+        placement={bottom ? 'bottom' : 'top'}
+        getPopupContainer={() => {
+          console.log('get');
```
jm90m: missed console log here

src/client/components/BTooltip.js
```diff
@@ -0,0 +1,61 @@
...
+  render() {
+    const { title, ...props } = this.props;
+    const { bottom } = this.state;
+
+    return (
+      <Tooltip
+        {...props}
+        title={<div ref={this.setTitle}>{title}</div>}
+        placement={bottom ? 'bottom' : 'top'}
+        getPopupContainer={() => {
```
jm90m: might be better to create a handler for this

bonustrack:
Isn't something antdesign can handle? I feel like it adding complexity to the app and it's not an issue we entirely solve. For example:
- The others tooltips for Like icon or in others place are not visible when they reach the top of the screen
- The vote list tooltip does not automatically move to bottom position when it reach to top if we keep the mouse on it
- The vote list tooltip does not seem to take in count the tooltip height properly. It stay in top position even if 25% of the tooltip is still not visible.

Sekhmet:
Antdesign should handle it with `autoAdjustOverflow` property, but it just doesn't work.

Ad. 1 - we can make it work by using BTooltip for those as well. I'm not sure if it makes sense though as they are just small tooltips.
Ad. 2 - how is it supposed to work? If it would switch position you would no longer have mouse hovering over it so it will disappear. Any complicated solution introduces more problems (tooltip disappearing without user intent, clipping of tooltip). This is not supposed to make a Tooltip that predicts everything, it's just a tweak to make things a little bit better.
Ad. 3 - We can't get the height of tooltip before it's shown (limitation of antd). We could solve that by hardcoding height or prerendering it just to get height, but it will make this component awfully complicated. Totally opposite of my intent to just display a tooltip on top or bottom depending on its position on the screen.

bonustrack:
Looks good now, if we can have this custom tooltip on every tooltip that would be perfect @Sekhmet

PR #1730, opened 2018-04-11
Disable autosize from EditorInput

### LINKED ISSUE #1500: Disable autosize in Editor
Collected a number of features I'd like you to consider, please
https://steemit.com/busy/@ash/dear-busy-org

#### Edited
There is hard limit on the number of rows that Editor can have. We should disable this limit.

Sekhmet:
Fixes #1500 

### Changes
- [x] EditorInput doesn't scale automatically.
- [x] It can be scaled in any way user wants.

### Test plan
- [x] Go to editor: https://busy-master-pr-1730.herokuapp.com/editor
- [x] Type something longer than 12 rows.
- [x] TextInput shouldn't scale.
- [x] Dragging EditorInput should resize input (its height).

### Screenshot

image

### DIFF
--- a/src/client/components/Comments/CommentForm.js
+++ b/src/client/components/Comments/CommentForm.js
@@ -111,8 +111,8 @@ class CommentForm extends React.Component {
         <div className="CommentForm__text">
           <Element name="commentFormInputScrollerElement">
             <EditorInput
+              rows={3}
               inputRef={this.setInput}
-              autosize={{ minRows: 3, maxRows: 6 }}
               value={body}
               onChange={this.handleBodyUpdate}
               onImageUpload={this.props.onImageUpload}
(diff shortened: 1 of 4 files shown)


bonustrack:
I wonder if is more easy to use this field when it take more space than the window height. Isn't making UX worst? I would be ok to have it bigger within the window height.

Sekhmet:
If the user wants to make it longer, bad UX would be not to let the user do it.

bonustrack:
But the thing is that you cant keep it smaller, ideally user should be able to set the height he want and we should keep this height for him.

Sekhmet:
Oh, I got what you mean. It resizes automatically as user types to the max size.
I think it might make sense to don't use `autoResize` feature at all and just let the user resize it themselves.

bonustrack:
Right @Sekhmet

PR #1735, opened 2018-04-12
Add error checks for 'constructor' username

### LINKED ISSUE #1723: Another JSON meta-data problem
Related to #1722 #1697 #1721

Found problem [here](https://github.com/steemit/condenser/issues/2693#issuecomment-379934098).

User profile breaks https://busy.org/@constructor

Not sure why, tried running page with the 2 PRs linked above didn't fix it.

Not sure how to look at user's posts doesn't work on steemit, busy, or steemnow.

jm90m:
Fixes #1723.
breaks - http://busy.org/@constructor 

Changes:
- was getting the `constructor` function in the user details objects rather than the actual username 'constructor' lol.. constructor is a reserved word


Sekhmet:
I would rather add prefix to the keys than to let users modify our objects.
There are many reserved keywords for Objects, and they can be overwritten by crafting account with specific user name, which can cause crashes or even vulnerabilities.

jm90m:
@Sekhmet good point, I added a helper in `stateHelpers` so we can use that to get the key when we store the user details in the state

Also updated activities / wallet page since they get affected by that username - https://busy.org/@constructor/transfers
https://busy.org/@constructor/activity

PR #1743, opened 2018-04-12
Add GitHub embed

### LINKED ISSUE #1523: Embed code on post with GitHub
Can we have something like this for markdown code
https://medium.com/javascript-scene/10-tips-for-better-redux-architecture-69250425af44
image

Sekhmet:
Fixes #1523 

Changes:
- [x] Add embed route on the server
- [ ] Handle dynamic embed height.

### DIFF
--- a/src/client/components/Story/Body.js
+++ b/src/client/components/Story/Body.js
@@ -44,7 +44,7 @@ export function getHtml(body, jsonMetadata = {}, returnType = 'Object', options 
   let parsedBody = body.replace(/<!--([\s\S]+?)(-->|$)/g, '(html comment removed: $1)');
 
   parsedBody = parsedBody.replace(
-    /https:\/\/gist\.github\.com\/\w+\/(\w+)/,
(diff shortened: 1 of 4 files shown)


Sekhmet:
There is a problem with handling dynamic height for embeds.
We can determine and set iframe height by using `onLoad` event. However it won't fire because we render post contents to HTML. I think we will need different rendering logic for that (something that parses post contents and renders React nodes).

bonustrack:
@Sekhmet we should check if its possible to find a fix to display embed with the script tag directly. Otherwise find another way to dynamicaly get iframe height if its not adding so much complexity or work.

Sekhmet:
I tried to use script tag directly, but we can't do this as it's using `document.write` which we can't handle, as we are rendering it asynchronously and browsers block this behaviour.
Potentially we could extract HTML content from GitHub and render it, but we would need different rendering of posts content so we can do this dynamically (but then iframe would work just fine as well).

PR #1745, opened 2018-04-13
Check that replies feed is empty before calling getReplies

### LINKED ISSUE #1715: Replies section keeps showing the same reply multiple times if you navigate away and come back
#### Expected behavior
When you click replies it should load all your replies. If you navigate to another page and come back, those same replies should still only show up once.

#### Actual behavior
When you first visit the replies page it shows the correct number of replies. If you go to another page and then back to the replies page, it now shows two. If you repeat, it then shows three and so on.

#### How to reproduce
Go to https://busy.org/replies whilst signed in and then click on news for example. When you click back on to replies the same one should now show up twice. It's easier to see in the video below

* Browser: Opera 52

#### Recording Of The Bug
https://media.giphy.com/media/yvqSzpaBO7YW3JbCC0/giphy.gif

Recording with more replies to show the issue more clearly

https://streamable.com/9z459

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@jrawsthorne/replies-section-keeps-showing-the-same-reply-multiple-times-if-you-navigate-away-and-come-back">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

jrawsthorne:
Fixes #1715.

Changes:
- Check that the replies feed is empty before calling getReplies


bonustrack:
Good catch thank you for this PR @jrawsthorne . CircleCi is failing, are you able to solve the issues: https://circleci.com/gh/busyorg/busy/4967 if not @jm90m can take a look on these

jrawsthorne:
The problem is a blank username in the test so getFeedFromState gives an undefined error.
Could check with 
`const content = !_.isEmpty(username) && getFeedFromState('replies', username, feed);`

Sekhmet:
Thanks for contributing.
It doesn't make much sense anymore to test this component as it became pretty connected.
Can you:
- rename it from `IReplies` to just `Replies` (and don't expect unconnected class)
- delete related test

We probably should add integration tests for those in the future.

Issue #1407, opened 2018-01-29
Add back to top button


Sekhmet:
https://ant.design/components/back-top/

Sekhmet:
@bonustrack do we want to implement this?

Sekhmet:
It should be pretty easy to implement (just few lines of code, so shouldn't take much time). There is already back to top component in antd (library we use for UI) so we just have to add it to Wrapper component.

https://ant.design/components/back-top/

Sekhmet:
Fixed in #1755.

PR #1755, opened 2018-04-14
Add back to the top button

th7nder:
Fixes Issue: #1407 

Adds back to the top button.
### Summary of Changes
* added BackTop component from antd
### Screenshots/Videos
gifffff


bonustrack:
@pareshUD do you have a better icon or suggestion for the design?

pareshUD:
two column editor

Please check this and let me know if any changes are there

pareshUD:
go to top

bonustrack:
We may need to change the behavior on mobile so the button don't appear all the time on top of the feed. Maybe on small screen show the icon only if the screen no move for 5sec. What do you think @pareshUD. Otherwise the icon is good, can you add it on iconfont @pareshUD ?

pareshUD:
@bonustrack Uploaded the back to top arrow to iconfont.cn

bonustrack:
@th7nder Can you update the iconfont url to `//at.alicdn.com/t/font_317125_wqz6u4c8idi8jjor.css` and use `icon-back-top` for the back to top icon?

th7nder:
Ok, will do. Thanks!

th7nder:
Ok! I think I've done the changes, icon is updated and button toggles.
Button shows up when you scroll down a bit, and then you start to scroll up.
It hides when you start scrolling down again.

PR #1760, opened 2018-04-16
Hide payout and up/down vote counts when equal to 0

### LINKED ISSUE #1752: Hide reward and like count if it's equal to 0
It's not necessary to show rewards or like counter or comment counter for post and comments if it's equal to 0.

Before
image
After
image

garroadran:
Fixes #1752.

Changes:
- Add check for values = 0 before rendering tooltips and FormattedNumbers


src/client/components/CommentFooter/Buttons.js
```diff
@@ -174,28 +176,30 @@ class Buttons extends React.Component {
...
-                {upVotesMore}
-                {upVotesPreview.length === 0 && (
-                  <FormattedMessage id="no_likes" defaultMessage="No likes yet" />
-                )}
-              </div>
-            }
+        {upVotes.length > 0 && (
+          <span
+            className={classNames('CommentFooter__count', {
+              'CommentFooter__count--clickable': upVotes.length > 0 || downVotes.length > 0,
```
th7nder: I think that upVotes.length > 0 condition makes this check not so useful.

src/client/components/CommentFooter/Buttons.js
```diff
@@ -174,28 +176,30 @@ class Buttons extends React.Component {
...
-            <FormattedNumber value={upVotes.length} />
-            <span />
-          </Tooltip>
-        </span>
+            <Tooltip
+              title={
+                <div>
+                  {upVotesPreview}
+                  {upVotesMore}
+                  {upVotesPreview.length === 0 && (
```
th7nder: I think that this goes like with the conditions before, after checking upVotes.length > 0 it is not required to check this - it will be always false I suppose.

src/client/components/CommentFooter/Buttons.js
```diff
@@ -211,37 +215,39 @@ class Buttons extends React.Component {
...
+                </div>
+              }
+            >
+              <FormattedNumber value={downVotes.length} />
+              <span />
+            </Tooltip>
+          </span>
+        )}
+        {payoutValue >= 0.01 && <span className="CommentFooter__bullet" />}
+        {payoutValue >= 0.01 && (
```
th7nder: How about wrapping this one up in additional <span> to avoid double condition checking?

garroadran: Wasn't sure what the best way to deal with this was, because wrapping it up in a single conditional throws linter errors (adjacent span tags need to be wrapped in another tag). If there's no issue with breaking the linter rule, or with wrapping the two spans up in another div, then yeah, I'd definitely make that change.

Sekhmet: There is a new feature in React that lets you group elements without need of additional div:
https://reactjs.org/docs/fragments.html

src/client/components/CommentFooter/Buttons.js
garroadran: Good points on both of these. Would you like me to go back and fix it?

Sekhmet: Sure, you can give it a go.

src/client/components/CommentFooter/Buttons.js
```diff
@@ -211,37 +212,38 @@ class Buttons extends React.Component {
...
+                </div>
+              }
+            >
+              <FormattedNumber value={downVotes.length} />
+              <span />
+            </Tooltip>
+          </span>
+        )}
+        {payoutValue >= 0.01 && (
+          <>
```
Sekhmet: This `<>` is special syntax and requires babel plugin to transform it into `<React.Fragment>`. We should just use `<React.Fragment>` instead.

bonustrack:
The space between element is inconsistant actually (see red and green line). Is it possible to have the same space between element (the green width would be find)?
image

garroadran:
Sure thing! Will have a look at that this weekend.

garroadran:
Spacing between elements should be a consistent 10px now.

bonustrack:
@garroadran These 2 icons are too much near, it may cause trouble on mobile. Can you add more spacing but still keeping the spacing uniform across other ones?
image

bonustrack:
Something like this:
image

garroadran:
Sure thing, but let me just double check that I understood correctly:
Upvote and downvote should have slightly more spacing between them (I'm thinking 16px) than all the other buttons (10px)?

bonustrack:
All space must be same, but now upvote and downvote seem to not have space between each others. Look at this 
image . I'm not able to put my cursor between upvote and downvote without selecting one of them.

garroadran:
Yikes, yeah, that's no good.
I think I have a fix, but give me another day or two to double-check it.

Sekhmet:
Sure, thanks a lot.

garroadran:
I think it's looking fairly balanced now, although I feel like the way I went about it was a bit hacky.
What do you guys think?

busy01
busy02
busy03
busy04

Sekhmet:
Great! Thanks a lot.

Issue #1754, opened 2018-04-14
Improve accessiblity


Sekhmet:
This is tracker for accessibility related issues.

These issues don't have to (and probably shouldn't) be solved in one take. Every PR is welcomed.

#### Non-clickable elements
- [ ] `src/client/components/Story/StoryFull.js` - PopoverMenu is not clickable.
- [ ] `src/client/components/UserMenu.js` - Menu elements are not clickable.
- [ ] `src/client/invite/Invite.js` - Copy link button is not clickable.
- [ ] `src/client/settings/ProfileSettings.js` - Submit button is not clickable.
- [ ] `src/client/activity/UserActivitySearch.js` - Collapse/Expand icon is not clickable.
- [ ] `src/client/components/Sidebar/CryptoTrendingCharts.js` - Refresh and collapse icons are not clickable.

_In progress_:
- [ ] `src/client/components/Sidebar/InterestingPeople.js` - Refresh icon is not clickable. assigned to @garroadran 

_Done_:
- [x] `src/client/components/Sidebar/Topics.js` - Show more/less buttons are not clickable - assigned to @th7nder

th7nder:
I would be delighted if I could try to solve this one!

garroadran:
I'd love to have a crack at
`src/client/components/Sidebar/InterestingPeople.js` - Refresh icon is not clickable
Although it might take me a few days to get around to it.

Sekhmet:
Great, thanks!

aiyumi:
I got here after [trying to suggest accessibility improvements through Utopian.io][a11y-ideas] and having the suggestions rejected because this issue already existed. I see the issue has been closed for a couple of months and many things are still unsolved. Are there any plans to tackle the accessibility problems again?

Aside from the elements already mentioned in the issue, there are many important buttons that cannot be accessed from the keyboard (cannot be focused by using the "Tab" key nor activated by the "Enter" key), such as the "Like" button, the "Reblog" button, and the button to open the popover menu (most of them also have no labels for screen readers). The files where they are located:

* src/client/components/StoryFooter/Buttons.js - except for the link with the number of comments, all buttons are inaccessible from the keyboard.
* src/client/components/CommentFooter/Buttons.js - all buttons are inaccessible from the keyboard.
* src/client/components/PopoverMenu/PopoverMenuItem.js - popover menu items have text but are inaccessible from the keyboard.
[...]

PR #1762, opened 2018-04-16
Topics show more/less accessibility fix

th7nder:
Related #1754 

Show more/less buttons on sidebar are now clickable. 

Summary of Changes
* make show more/less buttons clickable


src/client/components/Sidebar/Topics.less
```diff
@@ -8,7 +9,7 @@
     font-weight: 500;
   }
 
-  & > h5 {
+  & > a {
     color: @text-color;
     margin: 12px 0;
```
Sekhmet: Ahrefs are inline elements and can't have margin. Because of that link changed its placement:
image
vs
image

PR #1776, opened 2018-04-17
Check user blog state is empty before calling getFeedContent

### LINKED ISSUE #1775: Check that user blog feed is empty before calling getFeedContent
#### Expected behavior

When you visit a users page it should load all their posts. If you navigate to another user page (using search or clicking the user of a like) and come back, those same posts should still only show up once.

#### Actual behavior
When you first visit the user's page it shows the correct number of posts. If you go to another user page and then back to the user's page, it now shows the posts twice. If you repeat, it then shows them three times and so on.

#### How to reproduce

Go to a user page (https://busy.org/@jrawsthorne for example) and then click the likes of a post. From there go to another user's page and back again. All the posts will now be duplicated.

This is the same bug I reported previously with replies. I'm just about to submit a pr but wanted to make sure everyone could reproduce it if necessary.

* Browser: Opera 52

#### Recording Of The Bug

https://streamable.com/tqzas

jrawsthorne:
Fixes #1775 .

Changes:
- Check user blog state is empty before calling getFeedContent

### DIFF
--- a/src/client/user/UserProfile.js
+++ b/src/client/user/UserProfile.js
@@ -66,7 +66,11 @@ export default class UserProfile extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.match.url !== this.props.match.url) {
+    const { feed } = this.props;
+    const username = this.props.match.params.name;
+    const content = getFeedFromState('blog', username, feed);
+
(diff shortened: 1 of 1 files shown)


Sekhmet:
Looks good to me! Thanks.

PR #1791, opened 2018-04-26
use open instead of close price for SBD

### LINKED ISSUE #1788: SBD price is wrong
SBD price is wrong
on Busy:
image
on Coinmarketcap:
image

jm90m:
Fixes #1788

### Changes
Temporarily use open price instead of close since it is closer to the actual price for steem dollar


bonustrack:
BTC price for SBD is still wrong. And open doesn't give us latest price. We can remove get ride if the chart and use Bittrex API instead of CryptoCompare unless they answered your email.

Sekhmet:
Is this PR supposed to change `open` instead of `close` or it just changes ticker?

bonustrack:
@Sekhmet now just change the ticker

bonustrack:
There is still an issue with estimated account but this fix SBD rate

PR #1793, opened 2018-04-28
refactor: make SSR business logic agnostic

Sekhmet:
### Changes

Summary of changes you made.

* Move business logic from server to client.


bonustrack:
It seem that it bring 500ko weight more into busy.main because of the locales
image
https://busy-master-pr-1793.herokuapp.com/js/statistics.html

PR #1824, opened 2018-05-04
Fix displaying comment count element

adammankowski1:
### Changes

Summary of changes you made.

* Comment count element was visible event if there was no any comments. I've changed it to display it only when comments count is greater than 0.

### Test plan

You can use my [profile](https://staging.busy.org/@thequake) to check changes.

* Check and remember space between comment icon and share icon
* Run local version with this PR

### Demo

pre
busy_pre

after
busy_after


bonustrack:
Thank you for your contribution, we need the cursor pointer here too:
image

adammankowski1:
Okay, just doing it

PR #1831, opened 2018-05-07
chore: remove Block user menu

Sekhmet:
We had Block user menu entry which is not implemented yet and was confusing to users. After removing it we would have only one entry (Transfer) in popover menu which doesn't feel right. I added secondary Transfer button instead.

### Changes

Summary of changes you made.

* Remove `Block user` entry
* Replace PopoverMenu with secondary button.

### Test plan

* [X] Go to: https://busy-master-pr-1831.herokuapp.com/@slowwalker
* [X] Functional Transfer button should be displayed next to Follow button.

### Demo

Before:
zrzut ekranu 2018-05-07 o 20 46 51

After:
zrzut ekranu 2018-05-07 o 20 48 27


bonustrack:
It doesn't look good on small screen for me can you fix that?:
image

Sekhmet:
screen shot 2018-05-08 at 19 34 45

PR #1834, opened 2018-05-07
feat: add empty comments message

Sekhmet:
Comments section looks confusing right now if there are no comments. Users might think that comments failed to load.

### Changes

* add empty comments message

### Test plan

* [x] Go to: https://busy-master-pr-1834.herokuapp.com/@hellosteem/5ayq6b-test
* [x] There should be "no comments" message.

### Demo

Before:
zrzut ekranu 2018-05-07 o 23 53 14

After:
zrzut ekranu 2018-05-07 o 23 53 22


src/client/components/Comments/Comments.js
```diff
@@ -210,6 +210,8 @@ class Comments extends React.Component {
     const rootLinkedComment = findRootComment(comments, getLinkedComment(comments));
     const commentsToRender = this.commentsToRender(rootLevelComments, rootLinkedComment);
 
+    const empty = !loading && commentsToRender.length === 0;
```
bonustrack: Actually we dont need to load comments if the root post show that there is `comment_count: 0`. We can show this empty comment message directly without waiting for a loading.

Sekhmet:
Now comments are not fetched from API if `children` is 0.

PR #1835, opened 2018-05-07
Show powered-up username on power up transaction

### LINKED ISSUE #1820: Wallet Power up doesnt show the username
#### Expected behavior

In wallet, when power up operation to another account, it should display this other account username
image

#### Actual behavior

not displaying username, so it is confusing 
image

emre:
Fixes #1820 

### Changes

Wallet Power up transaction now shows the username if the transaction actor and the powered-up username is different. 

### Test plan

* [x] Find a power-up transaction includes a different ```to``` property compared to transaction actor. There should be an additional link to powered-up username in the activity.
* [x] Find a power-up transaction includes the same ```to``` property compared to transaction actor.There shouldn't be any link to the powered-up username in the activity. (Same with the old behaviour.)

### Demo 

*Before*

screen shot 2018-05-08 at 1 07 20 am

*After*

screen shot 2018-05-08 at 1 08 04 am


src/client/wallet/PowerUpTransaction.js
```diff
@@ -1,16 +1,23 @@
...
-const PowerUpTransaction = ({ timestamp, amount }) => (
+const PowerUpTransaction = ({ timestamp, amount, to, from }) => (
   <div className="UserWalletTransactions__transaction">
     <div className="UserWalletTransactions__icon-container">
       <i className="iconfont icon-flashlight_fill UserWalletTransactions__icon" />
     </div>
     <div className="UserWalletTransactions__content">
       <div className="UserWalletTransactions__content-recipient">
         <FormattedMessage id="powered_up" defaultMessage="Powered up " />
+        {to === from ? null : (
```
Sekhmet: We can use `to !== from &&` instead of ternary.

src/client/wallet/PowerUpTransaction.js
```diff
@@ -1,16 +1,23 @@
...
+          <Link to={`/@${to}`}>
+            &nbsp;<span className="username">{to}</span>
```
Sekhmet: I feel like we could have specific translation `power_up_to` with message `Power up {to}`, so it would work well with other languages as well.

src/client/wallet/PowerUpTransaction.js
emre: Yes, that's better. Squashed the commit with your suggestions. Thank you. :)

src/client/wallet/PowerUpTransaction.js
emre: Will do.

Sekhmet:
Hey,
do you want to update the translation for this, so we have separate `Powered up {username}` message?

emre:
hey @Sekhmet, I can add a new FormattedMessage with "Powered up {username}". Traveling in the last couple of days so couldn't update the pr yet. Will update the pr in 1-2 days though.

Sekhmet:
Sure, no rush. Thanks a lot.

emre:
Updated and rebased the commit.

Issue #1890, opened 2018-05-22
notifications transaction memo


tonimese:
Clicking on a notification about a transaction should open the wallet or the activity page to see the memo.
![image](https://user-images.githubusercontent.com/20820467/40365606-97ce5d9e-5ddd-11e8-9d34-cab431c6a69c.png)

Or It could display the memo when hovering over it.
![image](https://user-images.githubusercontent.com/20820467/40365975-aba3a24c-5dde-11e8-9635-d38b7ff2c282.png)

Or have buttons to choose the action.
![image](https://user-images.githubusercontent.com/20820467/40365316-f32496b4-5ddc-11e8-921b-4b0105ff1e75.png)

#### Actual behavior
Now clicking on notification about a transaction opens account profile.

Sekhmet:
Fixed in #1919

PR #1919, opened 2018-05-30
Transfer notification link to wallet instead of sender profile

kirkins:
This is the change suggested in #1890.

When user clicks on a received transfer notification send them to the wallet page instead of profile of user who sent the transfer.

### DIFF
--- a/src/client/components/Navigation/Notifications/NotificationTransfer.js
+++ b/src/client/components/Navigation/Notifications/NotificationTransfer.js
@@ -9,7 +9,7 @@ import './Notification.less';
 
 const NotificationTransfer = ({ notification, read, onClick }) => (
   <Link
-    to="wallet"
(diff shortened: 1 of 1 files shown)


src/client/components/Navigation/Notifications/NotificationTransfer.js
```diff
@@ -9,7 +9,7 @@ import './Notification.less';
 
 const NotificationTransfer = ({ notification, read, onClick }) => (
   <Link
-    to={`/@${notification.from}`}
+    to={`/wallet`}
```
Sekhmet: You can use regular `"wallet"` instead of string interpolation there.

Sekhmet:
Sorry, I messed up, forgot `/` in URL.
It should be `"/wallet"`.

Issue #1903, opened 2018-05-24
"Follow you" text indicator is missing text-shadow


bonustrack:
There should be a text-shadow for follow you indicator same than the handle or grade
![image](https://user-images.githubusercontent.com/16245250/40494735-c5215e4a-5f9f-11e8-83bf-76467219e861.png)

Sekhmet:
Fixed in #1920

PR #1920, opened 2018-05-30
Add text shadow to follows-you text on mobile

kirkins:
Fix for #1903, adds text shadow to follows-you text on mobile.

### DIFF
--- a/src/client/components/UserHeader.less
+++ b/src/client/components/UserHeader.less
@@ -50,6 +50,7 @@
     position: relative;
     top: -30px;
     text-align: center;
+    text-shadow: 0 0 4px rgba(0, 0, 0, 0.8);
 
     @media @small {
       display: none;
--- a/src/client/components/UserHeader.less
+++ b/src/client/components/UserHeader.less
@@ -50,7 +50,6 @@
     position: relative;
     top: -30px;
     text-align: center;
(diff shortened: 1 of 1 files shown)


Sekhmet:
It shouldn't be mobile only. Follows you should have shadow if user has cover photo, similar to this:
https://github.com/kirkins/busy/blob/528e61210a3b9809378cbc162defa1864c86e7fb/src/client/components/UserHeader.less#L23-L34

kirkins:
I'm not sure if this change is what you want.

I see these classes for that text.

    <span class="UserHeader__follows-you UserHeader__follows-you--mobile UserHeader__follows-you-cover-text-color"><span>Follows you</span></span>

The only one referencing cover is `UserHeader__follows-you-cover-text-color` so I'm targeting that now. New to using less.

Sekhmet:
You can just add `.UserHeader__follows-you` here (same as rank, handle, more etc.)

```less
  &--cover {
    background-color: darken(@primary-color, 25%) !important;


    .UserHeader__rank,
    .UserHeader__rank & > i,
    .UserHeader__handle,
    .UserHeader__more,
    .UserHeader__user__username {
      color: #ffffff !important;
      text-shadow: 0 0 4px rgba(0, 0, 0, 0.8);
    }
  }
```

Issue #1812, opened 2018-05-01
Remove js-logger


Sekhmet:
We currently use as dependency and initialise `js-logger`, but we don't really use it anywhere. We can just remove it entirely.

adammankowski1:
It would be my honor to delete it. Please assign me this issue.

PR #1921, opened 2018-05-30
Remove unused package js-logger

kirkins:
This removes an unused package `js-logger` was assigned #1812 to other developer but left open 29 days.


adammankowski1:
@kirkins woah, I've done it on my fork, but I forgot to send PR. Little tip, always remember to update your lock file. It's better to use npm/yarn commands to manage packages :)

PR #1922, opened 2018-05-30
Add busy version to settings page

### LINKED ISSUE #1811: Version number in settings
Showing Busy version number in settings page could be useful for debugging and referencing.

kirkins:
Fixes #1811 by adding the suggested version number to the settings page.

### DIFF
--- a/src/client/settings/Settings.js
+++ b/src/client/settings/Settings.js
@@ -362,8 +362,9 @@ export default class Settings extends React.Component {
                 <div className="Settings__version">
                   <p>
                     <FormattedMessage
-                      id="busy_version"
-                      defaultMessage={`Version: ${packageJson.version}`}
+                      id="version"
+                      defaultMessage="Version: {version}"
+                      values={{ version: packageJson.version }}
                     />
                   </p>
                 </div>
--- a/src/client/settings/Settings.js
+++ b/src/client/settings/Settings.js
@@ -363,7 +363,7 @@ export default class Settings extends React.Component {
                   <p>
                     <FormattedMessage
                       id="busy_version"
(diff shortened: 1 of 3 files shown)


kirkins:
Added some padding to the version, not sure what it should look like.

src/client/settings/Settings.js
```diff
@@ -27,6 +27,7 @@ import requiresLogin from '../auth/requiresLogin';
 import LANGUAGES from '../translations/languages';
 import { getLanguageText } from '../translations';
 import './Settings.less';
+import PackageJson from '../../../package.json';
```
Sekhmet: I would just call this variable `package`.

Sekhmet:
Thanks. You should add default translation to default.json as well:
https://github.com/busyorg/busy/blob/master/docs/CONTRIBUTING.md#adding-new-translations

This guide doesn't cover adding translations with variables yet, but you should create variable:
`id: "version"`
`defaultMessage: "Version: {version}".`

Otherwise we would have to translate each version to all languages.

See values prop for FormattedMessage: https://github.com/yahoo/react-intl/wiki/Components#formattedmessage

Sekhmet:
@bonustrack what text should we use?

bonustrack:
"Version: {version}" like this without "Busy" is fine.

kirkins:
Did I miss anything for adding the ID?

src/client/settings/Settings.js
kirkins: Seems the word is reserved. I'll make the p lowercase.

Sekhmet:
Formatted message should look like this
```js
<FormattedMessage id="version" defaultMessage="Version: {version}" values={{ version: package.version }} />
```

Can you rename translation ID to just `version`?

PR #1926, opened 2018-05-31
Display custom errors instead of Unkown error

### LINKED ISSUE #1918: Error messages from SC2 are not visible for user anymore
We used to have a detail of the error message visible for the user, but now we only get "Unknown error" for any error. This should not be the case, it's confusing for an user and make harder to debug.
image

gregory-latinier:
Fixes #1918 

### Changes
Fix the error displayed
Add an other custom message

### Test plan

Use an account with less than 1 SP and vote for a post or comment


bonustrack:
We should not show long error like this:
image

We could parse error from blockchain, it's something we have in steemconnect. We could use the parsed error from the blockchain answer as fallback if we dont have set our specific message.

Sekhmet:
We used to do this, but some errors use different format and it wasn't working with every error.
If you have solution on SteemConnect that works with everything I think we can parse it (we would lose ability to translate it though).

bonustrack:
We should try to first use custom error message if we can detect the error, then fallback to raw blockchain parsed error if we dont have custom error message for it, if we can't parse blockchain raw error we can just show "Unknown error"

Sekhmet:
@gregory-latinier 
We used to parse blockchain error to extract human readable error, instead of raw assertion. Fabien means we should do something similar, and if it fails for some reason we should display Unknown error.

1. Fingerprinted error message.
2. Parsed assertion.
3. Unknown error.

gregory-latinier:
Are there any specs or details on how the error message from the blockchain should be parsed ? What text is considered relevant ?

bonustrack:
Try this https://github.com/steemit/steemconnect/blob/5f78218b07486cb1a06a38c87587ac1033d11b3d/helpers/operation.js#L13-L25 @gregory-latinier

gregory-latinier:
I don't understand what you're expecting.
Here is an error example `info->abs_rshares > STEEM_VOTE_DUST_THRESHOLD || vote_weight == 0: Voting weight is too small, please accumulate more voting power or steem power.`
This is the raw text. If we don't use the fingerprint to display a custom error should we display this error as is or do something else ? I wouldn't mind a little bit more details.

Sekhmet:
We used to extract text from such message:
`info->abs_rshares > STEEM_VOTE_DUST_THRESHOLD || vote_weight == 0: Voting weight is too small, please accumulate more voting power or steem power.`
would be parsed to:
`Voting weight is too small, please accumulate more voting power or steem power.`

gregory-latinier:
Ok so I've generated a few errors to find a common denominator and for instance:
`_db.get_balance( o.from, o.amount.symbol ) >= o.amount: Account does not have sufficient funds for transfer.`
we can see that there is a technical explanation separated by a `:` so I decided to slice the string from the first `:` and display the next part

src/client/errorMiddleware.js
```diff
@@ -3,15 +3,16 @@ import errors from '../common/constants/errors';
...
   }
-
-  // Log error to console for further investigation.
-  console.log('Unknown error', error);
-  return 'Unkown error has occured.';
+  const idx = error.indexOf(':');
+  if (idx > 0) {
+    return error.slice(idx + 1).trim();
+  }
+  return error;
```
Sekhmet: I think it should return `Unknown error` there.

bonustrack: Yeah you can use "Unknown error" and console log the error so we could debug it for user.

PR #1934, opened 2018-06-01
Add activity indicator to avatar

### LINKED ISSUE #1684: Add activity green bullet indicator
Add activity indicator with a green bullet on bottom right of the avatar in profile page like this:
image
Show the green bullet when one of these account parameter update is less than 5min ago:
- `last_owner_update`
- `last_account_update`
- `last_vote_time`
- `last_account_recovery`
- `last_post`
- `last_root_post`

gregory-latinier:
Fixes #1684 

### Changes

A global properties reducer was added because I need the time of the blockchain as a reference point for date comparison.
The UserHero class was transformed from Stateless to Component in order to load the global properties
The AvatarLightbox was updated to display the activity bullet

### TODO on another PR
Refactor the wallet so that the global properties this component uses are moved to global properties reducer

### Test plan

Update your profile and reload the page. The bullet is displayed.

### Demo (optional)

Before update

image

Profile updated

image

5 min later

image


src/client/user/UserHero.js
```diff
@@ -29,64 +40,91 @@ class UserMenuWrapper extends React.Component {
...
-        )}
-      />
-      <Route render={() => (authenticated ? <Hero /> : <div />)} />
-    </Switch>
-  </div>
-);
+// eslint-disable-next-line react/no-multi-comp
+@connect(
+  state => ({
+    time: getGlobalPropertiesTime(state),
```
bonustrack: Why you are requesting the blockchain to know the time? It feel like adding a lot of complexity for nothing, cuz we have the time without it.

gregory-latinier: Last action must be base on blockchain time otherwise it won't work. For instance I got a 2 hours difference between my computer and the blockchain so if I use `new Date()` only the activity is already 2 hours old and the activity indicator never shows

src/client/components/UserHeader.less
```diff
@@ -69,6 +69,18 @@
...
+    position: relative;
+  }
+  &--active {
+    position: absolute;
+    left: 75px;
+    bottom: 0;
+    width: 25px;
+    height: 25px;
+    border-radius: 50%;
+    background: #a8f460;
```
bonustrack: We have a var set for the green color `@green` can you use it instead?

gregory-latinier: The variable `@green` is not the same color of the screenshot you provided. Is that normal / OK ?

bonustrack: Yes it's fine @gregory-latinier i've used a random color in the screenshot.

src/client/components/UserHeader.js
```diff
@@ -23,14 +23,15 @@ const UserHeader = ({
...
+  isActive,
 }) => {
   const style = hasCover
     ? { backgroundImage: `url("https://steemitimages.com/2048x512/${coverImage}")` }
     : {};
   return (
     <div className={classNames('UserHeader', { 'UserHeader--cover': hasCover })} style={style}>
       <div className="UserHeader__container">
-        <AvatarLightbox username={handle} size={100} />
+        <AvatarLightbox username={handle} size={100} active={isActive()} />
```
Sekhmet: Is there a reason for isActive being function instead of bool?

src/client/user/UserHero.js
```diff
@@ -58,6 +66,16 @@ const UserHero = ({
                 hasCover={hasCover}
                 isFollowing={isFollowing}
                 onTransferClick={onTransferClick}
+                isActive={() =>
```
Sekhmet: Quite a lot of code there. It probably could be extracted to variable (if using boolean is possible) or method.

src/client/components/UserHeader.js
```diff
@@ -123,6 +124,7 @@ UserHeader.propTypes = {
   hasCover: PropTypes.bool,
   isFollowing: PropTypes.bool,
   onTransferClick: PropTypes.func,
+  isActive: PropTypes.bool.isRequired,
```
Sekhmet: We have two different styles of naming booleans in our codebase (mostly because of some legacy code and lack of consensus on that). Now it looks weird to have `active` and `isActive` represent the same thing. I think we should pick one version for booleans and use it from now on, and update old code as well. Not sure which one we should prefer.

@gregory-latinier @jm90m 

nit: usually we have required props first, then optional props.

gregory-latinier: I'm used to write `isActive` and `setActive` from my past java coder life.

"Using the is prefix solves a common problem of choosing bad boolean names like status or flag. isStatus or isFlag simply doesn't fit,"

bonustrack: Should be fine with if we include the prefix "is". isFlag could be isFlagged, isStatus does not make sense i think we would need to change naming, "status" does not feel like a boolean to me.

gregory-latinier: it's the purpose of using `is` `flag` and `status` aren't good boolean name because `isStatus` doesn't make sense

Sekhmet: Let's try to use `isActive` or `hasChildren`, to avoid confusion with DOM attributes (which are single word). We would use this naming for all boolean variables (not only props) for consistency.

PR #1960, opened 2018-06-11
Define the white domains list in an helper

### LINKED ISSUE #1931: Whitelist domains like steemit.com from exit page
Now when an user click on external link there will be an exit page to show up for him. There is some websites that doesn't require this exit page, we should create a whitelist for it in a json file. We can start to whitelist these domains:

```
steemit.com
steem.io
d.tube
dlive.io
utopian.io
coinmarketcap.com
youtube.com
```

gregory-latinier:
Extract the domain from url and redirect directly

Fixes #1931 

### Changes

Add a white domains list in an helper
Extract the domain from an url and redirect immediately if it's in the whitelist

### Test plan

Open the post https://busy.org/@utopian-io/people-of-utopian-4-knowledges and click on the utopian.io link => you're redirected immediately

Open the post https://busy.org/@kaleem345/be-smart-no-1-cares-about-your-effort-only-the-results-zg1hbmlh-w3z9r => you're on the exit page

### DIFF
--- a/src/client/statics/ExitPage.js
+++ b/src/client/statics/ExitPage.js
@@ -4,6 +4,7 @@ import 'url-search-params-polyfill';
 import { injectIntl, FormattedMessage } from 'react-intl';
 import ActionLink from '../components/Button/ActionLink';
 import ActionButton from '../components/Button/Action';
+import { whitelistDomains } from '../helpers/exitPageHelper';
 import './ExitPage.less';
 
 @injectIntl
@@ -12,6 +13,14 @@ export default class ExitPage extends React.Component {
     intl: PropTypes.shape().isRequired,
     location: PropTypes.shape().isRequired,
   };
+
+  componentWillMount() {
+    const url = decodeURIComponent(new URLSearchParams(location.search).get('url'));
+    const hostname = new URL(url).hostname;
+    if (whitelistDomains.includes(hostname.replace(/^[^.]+\./g, ''))) {
+      window.location = url;
+    }
+  }
 
   closeWindow = () => {
     if (typeof window !== 'undefined') {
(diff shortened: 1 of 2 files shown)


src/client/statics/ExitPage.js
```diff
@@ -13,6 +14,14 @@ export default class ExitPage extends React.Component {
     location: PropTypes.shape().isRequired,
   };
 
+  componentWillMount() {
+    const url = decodeURIComponent(new URLSearchParams(location.search).get('url'));
+    const hostname = new URL(url).hostname;
+    if (whitelistDomains.includes(hostname.replace(/^[^.]+\./g, ''))) {
```
Sekhmet: `includes` doesn't work in IE11 so it will crash Busy.

gregory-latinier: no polyfills are used ?

gregory-latinier: instead of modifying the exitPage may be I should test the whitelist before and use the `secureLinks` options ?
Sorry I'm still discovering the app tbh :)

bonustrack:
We should keep icon for external links even if the domain is whitelisted already.

src/client/statics/ExitPage.js
bonustrack: Yes it would be better to see the link from the post directly and make direct link to it. Perhpas we can have a component "ExternalLink" where we put the logic.

Sekhmet: We should have polyfills (via `babel-polyfill`), but for some reason exit page crashes on IE11 on Windows 10 (works on staging). I'm not sure what is the issue as BrowserStack doesn't open dev tools for me, and I don't have any windows machine right now.

Sekhmet:
I will add whitelist in this PR #1965

PR #1965, opened 2018-06-12
Add exit page option to settings page + whitelist domains

### LINKED ISSUE #1931: Whitelist domains like steemit.com from exit page

### LINKED ISSUE #1932: Add setting option for disable exit page
We should give a way for user to enable or disable Exit page warn from the settings.

gregory-latinier:
Fixes #1931 
Fixes #1932 

### Changes

- Add `Enable exit page` setting.
- Add domain whitelist. 

### Test plan

- [ ] Visit: https://busy-master-pr-1965.herokuapp.com/@hellosteem/66fpvv-test-links
- [ ] Validate link behaviour with enabled exit page.
- [ ] Disable exit page.
- [ ] Check post links again.


src/client/components/Story/Body.js
```diff
@@ -109,13 +111,20 @@ Body.propTypes = {
...
 Body.defaultProps = {
   body: '',
   jsonMetadata: '',
   full: false,
   rewriteLinks: false,
+  exitPageSetting: true,
 };
 
-export default Body;
+export default connect(
```
Sekhmet: I would rather keep Body as dumb component and pass this prop instead (or create BodyContainer).

gregory-latinier: I think a BodyContainer would be best otherwise we would modify EmbeddedCommentForm, CommentForm for instance and connect them with the settings which doesn't make sense.
What is your final word ?

Sekhmet: I prefer having small container components that just wrap dumb component, so I think BodyContainer would work the best.

src/client/components/Story/Body.js
```diff
@@ -122,9 +122,8 @@ Body.defaultProps = {
   exitPageSetting: true,
 };
 
-export default connect(
-  state => ({
-    exitPageSetting: getExitPageSetting(state),
-  }),
-  null,
-)(Body);
+const BodyContainer = props => <Body {...props} />;
```
Sekhmet: This doesn't change anything, you can connect to `Body` directly.

src/client/components/Story/Body.js
```diff
@@ -122,9 +122,8 @@ Body.defaultProps = {
...
+
+export default connect(state => ({
```
Sekhmet: I think we should separate dumb and container components. Right now it works the same way as when we connected to Body directly. Maybe we should finally create `containers` directory.

gregory-latinier: Honestly I wouldn't call the Body component a dumb component. There are a lot of logic code for displaying data.

gregory-latinier: I agree that there should be containers that handle state and other app logic and presentational components. But that would require a lot of work and it's not the point of the PR. 
What should be done precisely to finalize this PR ?

Sekhmet: I'd say presentational components encapsulate UI logic, which I think current `Body` component does. Containers would just provide data to presentational components.

`components` folder should not contain containers, so I would extract `BodyContainer` to `containers` folder and use it in the app.

src/client/vendor/SanitizeConfig.js
```diff
@@ -160,9 +160,11 @@ export default ({ large = true, noImage = false, sanitizeErrors = [], secureLink
       if (!href) href = '#';
       href = href.trim();
       const attys = {};
-      // If it's not a (relative or absolute) steemit URL...
-      if (secureLinks && !href.match(/^^(\/|https:\/\/(staging\.)?busy\.org(?![\w\.]+))/)) {
+      if (!href.match(/^^(\/|https:\/\/(staging\.)?busy\.org(?![\w\.]+))/)) {
         attys.target = '_blank';
+      }
+      // If it's not a (relative or absolute) steemit URL...
```
Sekhmet: This comment belonged to different condition (regex match). I don't think it worth keeping this comment at all.

PR #1969, opened 2018-06-12
use originalComment.json_metadata when updating

mktcode:
to avoid removing existing metadata
if `isUpdating` use `originalComment.json_metadata`

- publish comment with custom metadata
- edit on busy.org
- check metadata again

https://busy.org/@mkt/bug-report-busy-org-overwriting-comment-jsonmetadata

I tested it locally in my fork of course.


gregory-latinier:
Your code looks good but could you provide a test plan please ?

mktcode:
Srsly?

- publish comment with custom metadata
- edit on busy.org
- check metadata again

mktcode:
Idk if that's what you mean but I updated my comment above.

src/client/comments/commentsActions.js
```diff
@@ -97,7 +97,9 @@ export const sendComment = (parentPost, body, isUpdating = false, originalCommen
   const permlink = isUpdating
     ? originalComment.permlink
     : createCommentPermlink(parentAuthor, parentPermlink);
-  const jsonMetadata = { tags: [category], community: 'busy', app: `busy/${version}` };
+  const jsonMetadata = isUpdating
+    ? JSON.parse(originalComment.json_metadata)
```
bonustrack: We should handle the case where json_metadata is not a real json. JSON.parse will fail if this happen with your implementation.

mktcode: You think https://github.com/busyorg/busy/blob/a2d948db624af33eb01c1237bfb3779df6539e0c/src/client/helpers/formatter.js is the right place to put a `isJson` helper function or something?

mktcode: I don't want to put it right in `commentsAction.js`

Sekhmet: We might want to extract method tryParseJSON that uses lodash's `try` and `isError` to return parsed JSON or empty object.

Sekhmet:
I'm not sure what happened to my comment there, but it would be nice to use this function after it's merged: #1981 instead of new `isJsonStr`, so we don't have to parse string twice, and make things more concise.

mktcode:
sure, I was just confused about this parse function somehow and didn't want to change it but your totally right... I just cannot make sense of this line https://github.com/busyorg/busy/pull/1981/files#diff-3790f296954d7550bfd6609c20a66551R2 isn't that covered by the catch?

mktcode:
updated everything and tested again... should be ready for merge

PR #1981, opened 2018-06-20
refactor: update jsonParse function

Sekhmet:
### Changes

* Update `jsonParse` function to return null in case of error.
* Update `getAppData` to use string JSON metadata to match real world data.
* Add `jsonParse` tests.


src/client/helpers/formatter.js
```diff
@@ -1,11 +1,11 @@
-export const jsonParse = str => {
+export function jsonParse(input) {
+  if (!input) return null;
```
mktcode: isn't that covered by the catch?

Sekhmet: This also makes function return `null` for input of `false` which is not desired.

mktcode: ok, didn't expect JSON.parse to behave like this... `JSON.parse(0) => 0` and `JSON.parse(false) => false` shouldn't this throw an exception? anyway...

PR #1982, opened 2018-06-20
fix: don't pass invalid props to DOM input

Sekhmet:
### Changes

* Prevent passing `inputId`, `inputRef`, `onImageUpload`, and `onImageInvalid` props to DOM node causing warnings.

### Test plan

* [x] Log in.
* [x] Go to some post: http://localhost:3000/@larrymorrison/8e0ad280-745b-11e8-97f4-27fc95e09096 (has to be local so warnings can be displayed).
* [x] No warnings should appear in the console.


jm90m:
haven't ran busy in a while, but just ran it, looks like `yarn dev` compiles the server now? So we don't see the debug errors, usually when you don't use variables within the current scope of the function, then react shows a warning in the terminal

Sekhmet:
It's not React thing, but eslint. You get that for free in tools like expo or create-react-app, but by default there is nothing like that.

Those variables are not unused. They are extracted using rest operator, this way they have their own purpose, even though they are not technically used anywhere in the code.

PR #2044, opened 2018-07-06
mobile wallet buttons

### LINKED ISSUE #1059: "Transfer" and "claim rewards" buttons are not accessible on mobile device
If a user uses a mobile device for browsing Busy, he is not able to transfer currency nor claim rewards in the wallet page or in the user's profile. If they go to another user profile, there is an option to send the funds to the user.

### Expected behaviour

It should be possible to transfer or claim rewards in the wallet page on mobile devices. This regards the users profile too.

### Steps to reproduce

1. Go to https://busy.org/wallet on mobile

- Browser: Brave
- OS: Android 7.1.2

/n/n<br /><hr/><em>Open Source Contribution posted via <a href="https://utopian.io/utopian-io/@espoem/transfer-and-claim-rewards-buttons-are-not-accessible-on-mobile-device">Utopian.io</a></em><hr/>

mktcode:
*I don't think this can be merged right away but I want to get this started... basically this is simply copied from the sidebar and I left out the market prices.*

Fixes #1059 

### Changes

* Added UserWalletButtons component
* Added it to UserWallet.js
* Show only on small screens

### Test plan

* [ ] Reduce screen size to below 768px
* [ ] use transfer, claim, power-up/-down buttons

### Demo

image

### DIFF
--- a/src/client/user/UserWallet.js
+++ b/src/client/user/UserWallet.js
@@ -4,6 +4,7 @@ import { withRouter } from 'react-router-dom';
 import { connect } from 'react-redux';
 import _ from 'lodash';
 import UserWalletSummary from '../wallet/UserWalletSummary';
+import UserWalletButtons from '../wallet/UserWalletButtons';
 import { SBD, STEEM } from '../../common/constants/cryptos';
 import { getUserDetailsKey } from '../helpers/stateHelpers';
 import UserWalletTransactions from '../wallet/UserWalletTransactions';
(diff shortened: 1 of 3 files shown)


Sekhmet:
I'm not really sure what ideal solution would be, but I think we should make it responsive, so everything is shown on every device.

Right now it's rendering sidebar twice, which one of them is visible and one is not.

Without well-thought responsive designs (and a desired solution for this problem) I wouldn't implement any half-solutions.

mktcode:
hmm... let me see if I can play around with the css to find a better solution.

mktcode:
I don't think it can be a general rule to put the right container on top of the center one in all cases... :/

PR #2071, opened 2018-07-16
allow mixed line endings locally

Sekhmet:
This will allow any mix of LF and CRLF locally, but will fail on CI if CRLF are commited. This basically requires users to configure Git to commit LF (`core.autocrlf true`).


Sekhmet:
Users can use crying emoji as line endings locally. As long as they commit LF we shouldn't make it harder for users to modify their environment just to make eslint pass.

PR #2094, opened 2018-07-31
Nightmode Styles Complete

### LINKED ISSUE #2065: Task Request - Busy.org Night Mode - 100 Steem w/ Bonus
#### Details
Implement Night Mode toggle accessible from the Profile drop down menu in the upper right corner.  Might Mode changes the white background with black text to a black background with white text. 

I will send 100 Steem to the accepted Pull Request's creator. 

BONUS 100 Steem if you also implement the ability link to a page on Busy.org with link syntax which sets the user's them to Night Mode if not currently logged into an account. 

So a total of 200 Steem if you implement both features. 

 
#### Components
This will effect the website's UI likely requiring CSS modifications. 
 
#### Deadline
Wednesday July 18th
 
#### Communication
Email me at mail@cahlen.io if you have questions.

mktcode:
Fixes #2065 

### Changes

* moved nightmode class from <Layout> to <body>
* changed nightmode classname from `.dark` to `.nightmode`
* added a lot of `<Component>-nightmode.less` files
* all wrapped by the body's `.nightmode` class
* overwrote necessary CSS
* imported files at bottom of related `<Component>.less`

### Test plan

* [ ] Click through aaaaall the possible views of the application with **Nightmode disabled** to check for unintended changes in normal styles
* [ ] Go to settings and **enable Nightmode**
* [ ] Click through aaaaall the possible views of the application to see the new nightmode

### Demo

image

image

image


Sekhmet:
Looks very good. Couple thoughts:
- I would change colour of avatar borders.
- We should change background colour on user profile: https://busy-develop-pr-2094.herokuapp.com/@hellosteem
- Nightmode should be applied on the server (now it flashes white layout after page load).

mktcode:
> We should change background colour on user profile

What exactly do you mean? It's the same background as everywhere else.

> Nightmode should be applied on the server

I think I need some help with that. How can I do that?

mktcode:
image

Done!

For applying nightmode server-side... do you have any hints for me?

Sekhmet:
Thanks. I will take a look later today.

mktcode:
I guess somewhere here directly on the template?
https://github.com/busyorg/busy/blob/develop/templates/index.hbs#L21

But how can I access the state var... it is rendered in the script tag `"nightmode": true` but... well... I'll try to find out myself in the meantime.

mktcode:
@Sekhmet What do you think about my [last commit](https://github.com/busyorg/busy/pull/2094/commits/c3334f51a119dc0a47ae7ed0fcb150ad9e2cfee1)? Works fine for me.

Sekhmet:
Hey.
What I would try to do is not to append classes to the body using `document.classList` and instead try to apply dark theme based on `dark` class on Layout. This way we won't have to modify SSR and avoid manipulating DOM directly.
Have you tried doing something similar?

mktcode:
I agree that the changes in Wrapper.js can now be removed: https://github.com/busyorg/busy/pull/2094/commits/e6800bc092cc0a73a7b9671bec63a08538c53ec1#diff-105a0ee4a79800ce929a0e78481a9286

But what are your concerns with this: https://github.com/busyorg/busy/pull/2094/commits/c3334f51a119dc0a47ae7ed0fcb150ad9e2cfee1

Sekhmet:
You could probably style everything using `dark` class on Layout instead of `nightmode` on body, so those changes in c3334f5 would be unnecessary.

mktcode:
Cmon dude...
> Nightmode should be applied on the server (now it flashes white layout after page load).
That's what https://github.com/busyorg/busy/commit/c3334f51a119dc0a47ae7ed0fcb150ad9e2cfee1 does.

I removed those lines, since onMount will be replace by server-side. https://github.com/busyorg/busy/pull/2094/commits/e6800bc092cc0a73a7b9671bec63a08538c53ec1#diff-105a0ee4a79800ce929a0e78481a9286R149 

But these lines are necessary to react to changing the settings option immediatly. https://github.com/busyorg/busy/pull/2094/commits/e6800bc092cc0a73a7b9671bec63a08538c53ec1#diff-105a0ee4a79800ce929a0e78481a9286R163

And it's better to apply the class to the body because it gives you more control. I first applied it to the Layout component because it seemed to be easier but also has some drawbacks.

Imho the code is now fine.

Sekhmet:
I know what c3334f5 does. I know that it solves this problem. But I'm trying to find a good solution, not any solution.

Anything that is inside of `server` folder should not be concerned by business logic (which nightmode is part of). It is supposed only to render app on the server. This is why I try to avoid modifying this part of code.

mktcode:
Then I don't know how to prevent this flashing of the white layout.

> It is supposed only to render app on the server.

Does nightmode not fall in this realm?

Sekhmet:
Merge this when ready
@cc @gregory-latinier @jm90m

jm90m:
thanks for the PR @mktcode, everything looks fine to me.

mktcode:
Done!

But the back to top button looks ok for me. 
image
I think that's not directly related to nightmode but just not visible on white background.

gregory-latinier:
.BBackTop_button i {
    color: #2088ff!important;
    font-size: 40px;
    line-height: 40px;
    background-color: #fff;
    border-radius: 50%;
// add this
    height: 37px;
    display: block;
    width: 37px;
}
```

if I add this it does the trick for me. It reduces the white background so it's behind the icon.
Does it work for you ?

mktcode:
What browser etc. are you using?

gregory-latinier:
Chrome on windows

mktcode:
Well... your CSS doesn't seem to change anything significant on my side so... if it fixes the windows side.... I'll commit! ;)

mktcode:
Thanks! My pleasure to contribute to the project!

mktcode:
@Sekhmet Can you tell when this will be deployed?

PR #2124, opened 2018-10-12
Update Buttons.js

drov0:
Fixes #

### Changes

VotesRshares can be equal to 0. Which will create a division by 0. 

### Test plan

Effects can be seen on this post : https://busy.org/@petanque/testcomments-xylmwuwi6p if you look at the comment "Commenting is cool" by @howo and the upvotes, you get "Nan $"


Sekhmet:
I've deployed this branch, but it still seems to display NaN:
https://busy-develop-pr-2124.herokuapp.com/@petanque/testcomments-xylmwuwi6p

drov0:
Mmmh that's odd, the division by 0  should be the only thing that causes this issue. Is the ratio calulation also done elsewhere ?

drov0:
Yep indeed the calculation was done elsewhere as well. In my previous commit I only fixed the bug when viewing the votes from the post itself. Now it should be fixed everywhere.

PR #2147, opened 2019-01-05
[Suggestion] show powerdown information in wallet

economicstudio:
### Changes

* Show the powerdown amount in the wallet
* Show the tooltip with next powerdown(withdrawal) date

### Test plan

* [ ] Pick a user powering down to check powerdown information is shown
* [ ] Pick another user not powering down to check powerdown information is not shown.

### Demo

> Currently, **Busy shows nothing!**

> After, Busy shows both powerdown and delegation in a compact form.

### Benefits

Given that most users are not familiar with steemd.com or at least not using it daily basis, widely used UIs should show sufficient powerdown information. There are several benefits.

1. Hacking prevention
There are many cases where hackers start the powerdown and victims even don't recognize it! If the information is properly shown, this hacking can be prevented.

2. Correct vote value
It can also prevent a bug of incorrect vote value. Even if the vote value is shown correctly, e.g., steempeak, users may wonder why their vote value is quite low. Note that around the end of 13 weeks of powerdown, the difference can be huge. In Ned's case, it's almost twice, and much more than twice next week.

3. Stopping powerdown
In addition, users powering down may change their mind to stop the powerdown if the information is clearly shown everyday. Especially when it's almost done (again, for instance, Ned), the powerdown amount is quite big compared to the remaining so the chance of changing their mind might be pretty high :)

### DIFF
--- a/src/client/vendor/steemitHelpers.js
+++ b/src/client/vendor/steemitHelpers.js
@@ -192,6 +192,12 @@ export const calculateTotalDelegatedSP = (user, totalVestingShares, totalVesting
   return receivedSP - delegatedSP;
 };
 
+export const calculatePendingWithdrawalSP = (user, totalVestingShares, totalVestingFundSteem) => {
+  return parseFloat(
+    formatter.vestToSteem(user.vesting_withdraw_rate, totalVestingShares, totalVestingFundSteem),
+  );
+};
+
 export const calculateVotingPower = user => {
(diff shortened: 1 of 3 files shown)


Sekhmet:
Maybe we should display three rows:
- `Steem Power`
- `Delegated Steem Power`
- `Powering down Steem Power`
We could hide last two if they are equals to 0.

PR #2150, opened 2019-01-13
fix-reblogged-by

### LINKED ISSUE #2149: reblogged_by is not shown
#### Expected behavior
Reblogged post in feed should be shown as "reblogged_by {username"}"

#### Actual behavior
Reblogged post in feed is shown the same way as other posts of followings.

#### How to reproduce
Write here every possible detail to reproduce the bug.

* Browser version: Any
* App version: Busy 2.5.6 (current)
* Operating system: Any

#### Recording Of The Bug

> Busy.org: my feed (rosatravels isn't my following)

> Steemit.com shows the reblogged information correctly

#### Cause and Fix
Busy.org uses `post.first_reblogged_by` which is always null. `post.reblogged_by' should be used, as in steemit.com https://github.com/steemit/condenser/blob/7430aca58ae9f86e501f52e3cf8dba5efc9fe3e0/src/app/components/cards/PostSummary.jsx#L70

I made the fix and PR. Here is the screenshot after the fix on my local dev server.

> **After the fix on my local dev server**

economicstudio:
Fixes #2149

### Changes

* Fix reblogged_by

`post.first_reblogged_by` returns null. `post.reblogged_by` should be used as follows:
https://github.com/steemit/condenser/blob/7430aca58ae9f86e501f52e3cf8dba5efc9fe3e0/src/app/components/cards/PostSummary.jsx#L70

### Test plan

* [x] Go to feed
* [x] Compare it with the feed on Steemit.com

### Demo

> Busy.org: my feed (rosatravels isn't my following)

> Steemit.com shows the reblogged information correctly

> **After the fix on my local dev server** 

In addition, to show 'reblogged' in blog (not feed), the current code should be changed. Currently there is no easy way to differentiate if it's feed or blog. But any resteemed post in a blog is easily noticeable by author's name, so not important as reblogged_by in feed.

### DIFF
--- a/src/client/components/Story/Story.js
+++ b/src/client/components/Story/Story.js
@@ -275,7 +275,7 @@ class Story extends React.Component {
 
     let rebloggedUI = null;
 
-    if (post.first_reblogged_by) {
+    if (post.reblogged_by) {
       rebloggedUI = (
         <div className="Story__reblog">
           <i className="iconfont icon-share1" />
@@ -284,8 +284,8 @@ class Story extends React.Component {
             defaultMessage="{username} reblogged"
             values={{
               username: (
-                <Link to={`/@${post.first_reblogged_by}`}>
-                  <span className="username">{post.first_reblogged_by}</span>
+                <Link to={`/@${post.reblogged_by[0]}`}>
+                  <span className="username">{post.reblogged_by[0]}</span>
                 </Link>
               ),
             }}


Sekhmet:
@economicstudio 
You have to check this checkbox on the right to let us edit this branch:
image

economicstudio:
Thanks for your approval and I made it editable.

PR #2174, opened 2019-01-29
Show zero payout

### LINKED ISSUE #2173: [suggestion] payout information for zero payout
##### Current payout information
Currently, Busy payout information has the following features:
- Show nothing when pending payout is less than $0.005 (i.e., $0.01 by rounding)

> https://github.com/busyorg/busy/blob/27dd2383806eda8daf46748cbbbb26739d08ced4/src/client/components/StoryFooter/Payout.js#L16

- Show nothing when paid out amount is zero

While this might be an aesthetical choice, this creates the following inconvenience.

##### Problems of the current payout information
- For a payout declined post, there is no way (unless you double check, e.g., with steemit or steemd) to tell if it's declined or not in the beginning. Imagine that you gave someone $10 voting, but it turns out that the post was payout-declined, so you now see 
  ## ~~$10.00~~

You may not want this :( **Remember that voting power can't be recovered even if you unvote.**

- When a post is paid out, there is no easy way to tell if zero payout is due to decline or downvoting.

##### Proposed payout information
- Busy should also show the payout information for zero pending or zero payout.

See Mockups / Examples for the details.

#### Mockups / Examples
Let's first see the Steemit.com's UI.

> Steemit.com shows it as ~~$0.00~~, so we can tell that Ned's post wasn't severely downvoted but he just just declined the payout.

Now, let's talk about Busy's UI.

##### Before

> Currently, Busy shows nothing when payout is zero. So it's hard to tell whether it's due to downvoting or decline.

As explained in the proposal, this might be more serious when a post has just been posted. Before voting, no way to tell whether the post is payout-declined or not.

##### After

> Busy should show it as ~~$0.00~~ for both posts just posted and paid out.

#### Benefits
- For a payout-declined post, you can easily tell if it's declined or not in the beginning
- When a post is paid out with $0, you can easily tell if it was due to decline or downvoting. (But maybe due to both :)
- Easy to understand for newbies. Newbie may not even understand why some post is paid out with zero.
- Consistency with Steemit.com. I don't think that everything should be consistent with Steemit, but for this, it's much better if it's consistent.

economicstudio:
Fixes #2173 

### Changes

* Show ~~$0.00~~ for payout-declined post all the time
* Show $0.00 or ~~$0.00~~ depending on the declined status even for posts with pending payout lower than $0.005.

### Test plan

* [ ] https://busy.org/@ned/the-guiding-mission-vision-and-values-of-steemit-inc should show ~~$0.00~~ instead of nothing
* [ ] Create any post. It should show either $0.00 or ~~$0.00~~ depending on the declined status from the beginning.

##### Before
https://busy.org/@ned/the-guiding-mission-vision-and-values-of-steemit-inc

> Currently, Busy shows nothing when payout is zero. So it's hard to tell whether it's due to downvoting or decline.

As explained in the proposal, this might be more serious when a post has just been posted. Before voting, no way to tell whether the post is payout-declined or not.

##### After

> Busy should show it as ~~$0.00~~ for both posts just posted and paid out.

### DIFF
--- a/src/client/components/StoryFooter/Payout.js
+++ b/src/client/components/StoryFooter/Payout.js
@@ -13,29 +13,27 @@ const Payout = ({ intl, post }) => {
   const payoutValue = payout.cashoutInTime ? payout.potentialPayout : payout.pastPayouts;
 
   return (
-    payoutValue >= 0.005 && (
-      <span className="Payout">
-        <BTooltip title={<PayoutDetail post={post} />}>
-          <span
-            className={classNames({
-              'Payout--rejected': payout.isPayoutDeclined,
-            })}
-          >
-            <USDDisplay value={payoutValue} />
-          </span>
+    <span className="Payout">
+      <BTooltip title={<PayoutDetail post={post} />}>
+        <span
+          className={classNames({
+            'Payout--rejected': payout.isPayoutDeclined,
+          })}
+        >
+          <USDDisplay value={payoutValue} />
+        </span>
+      </BTooltip>
+      {post.percent_steem_dollars === 0 && (
+        <BTooltip
+          title={intl.formatMessage({
+            id: 'reward_option_100',
+            defaultMessage: '100% Steem Power',
+          })}
+        >
+          <i className="iconfont icon-flashlight" />
         </BTooltip>
-        {post.percent_steem_dollars === 0 && (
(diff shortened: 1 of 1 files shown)


economicstudio:
Thanks for merging :)

PR #2176, opened 2019-01-29
Fix 2175: 3-digit precision

### LINKED ISSUE #2175: [suggestion] 3-digit precision for less than $0.02
Currently, Busy shows only up to 2-digit precision, while SBD has 3-digit precision.

This might be an aesthetical choice. While many people like 3-digit on Steempeak, for instance, I still prefer 2-digit when the amount is large enough. But currently $0.01 needs a full voting with about 500 SP, which means even $0.005 needs 250 SP. As you know, most newbies are much below than that. Their voting always marks as $0.00 **which can't be a good user experience**.

#### Proposed payout information
- For the amount less than $0.02, show up to 3-digit precision.

##### Why $0.02?
This is related to dust payout threshold (STEEM_MIN_PAYOUT_SBD), if payout is less than $0.02, it is not actually paid out. It just disappears! See https://busy.org/@blockchainstudio/effect-of-haircut-early-voting-beneficiary-on-dust-payout

#### Mockups / Examples

> before

> after

> but still 2-digit for >= $0.02

> voting details

#### Benefits
- Prevent dust payout. Pending <= $0.02 can easily be recognized.
- Provide a better user experience for newbies, which I believe very important. Why their votings should be shown as $0.00 all the time?

economicstudio:
Fixes #2175 

### Changes

* 3-digit precision when 0 < STU < $0.02
* 2-digit precision for all negative values
* negative values always have - sign.
* Test codes are added and modified for these.

### Test plan

* [ ] Pick any user whose voting value is less than $0.02, it should be shown up to 3-digit. For users whose voting value is at least $0.02, it should still be shown up to 2-digit.
* [ ] Do the same test for vote details for upvote and downvote

### Demo (optional)

> before

> after

> but still 2-digit for >= $0.02

> voting details

### DIFF
--- a/src/client/components/Utils/USDDisplay.js
+++ b/src/client/components/Utils/USDDisplay.js
@@ -6,7 +6,7 @@ const USDDisplay = ({ value }) => {
   const negative = value < 0;
   const absValue = Math.abs(value);
   // 0.02 is dust payout threshold (STEEM_MIN_PAYOUT_SBD)
-  const precision = absValue < 0.02 && value > 0 ? 3 : 2;
+  const precision = absValue < 0.02 && absValue > 0 ? 3 : 2;
   return (
     <span>
       {negative && '-'}
--- a/src/client/components/Utils/USDDisplay.js
+++ b/src/client/components/Utils/USDDisplay.js
@@ -5,11 +5,16 @@ import { FormattedNumber } from 'react-intl';
 const USDDisplay = ({ value }) => {
   const negative = value.toFixed(2) < 0;
   const absValue = Math.abs(value);
+  const precision = absValue < 0.02 && absValue > 0 ? 3 : 2;
   return (
     <span>
       {negative && '-'}
       {'$'}
-      <FormattedNumber value={absValue} minimumFractionDigits={2} maximumFractionDigits={2} />
+      <FormattedNumber
+        value={absValue}
+        minimumFractionDigits={precision}
+        maximumFractionDigits={precision}
+      />
     </span>
   );
 };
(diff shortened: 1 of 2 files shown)


Sekhmet:
Is there a reason to always display 2 digits precision for negative values?

src/client/components/Utils/USDDisplay.js
```diff
@@ -3,13 +3,19 @@ import PropTypes from 'prop-types';
...
 
 const USDDisplay = ({ value }) => {
-  const negative = value.toFixed(2) < 0;
+  const negative = value < 0;
   const absValue = Math.abs(value);
+  // 0.02 is dust payout threshold (STEEM_MIN_PAYOUT_SBD)
+  const precision = absValue < 0.02 && value > 0 ? 3 : 2;
   return (
     <span>
       {negative && '-'}
```
Sekhmet: I think we could get rid of this and use `FormattedNumber` to display a negative sign if needed.

Sekhmet: Right, I forgot about the `$` sign.

It might not be important to display `-$0.013` instead of `-$0.01`, but I'd rather see `-$0.009` instead of `-$0.00` (we display `$0.009` for positive values).

economicstudio: Okay I'll fix that. In another sense of consistency, you're right. I also actually implemented that way in the beginning :)

economicstudio: Done!

More test coded are also added for handling zero correctly. Thanks!

Also huge thanks for busy bot :) I also made a post about it: https://busy.org/@blockchainstudio/en-kr-busy-voting-independence-day-now-it-s-completely-independent-of-steemdb-busy-steemdb

PR #2182, opened 2019-02-04
feat: add downvote power slider with confirmation

### LINKED ISSUE #2126: Downvote button: add confirm + slider
The downvote option has serious issues. I went to click upvote, and hit downvote on my android by accident. A 100% flag was sent, costing 2.5% voting power.

This is really annoying that there hasn't been at least a confirm dialog added since the inception of the downvote. Please add it.

Also, a 100% upvote is limiting as well. Hardly anyone wants.to use that such power on a flag.

th7nder:
Fixes #2126 

### Changes

* Added downvote slider analogically to the upvote's one
* Updated Slider component indicator

### Test plan
* [ ] Make sure you have 'Enable slider' checked in settings or your rank is Dolphin or higher
* [ ] Select a random post and go to the comments' section
* [ ] Click dislike

### Demo (optional)
image
image


Sekhmet:
I think it might be a good idea to always display slider (for upvotes & downvotes) no matter user rank. Otherwise we will have to default to some downvote percent which is not ideal.

Sekhmet (commented):
x

PR #2185, opened 2019-02-11
fix: couldn't reblog with multiple accounts

### LINKED ISSUE #2081: Resteem bugs with multi accounts
#### Expected behavior

Beeing able to resteem an article, even if using busy with several accounts

#### Actual behavior

When logging on a secondary account, resteem button in already selected on the targeted article

#### How to reproduce

log in with account A, choose any article and resteemit
log off
log in with account b
search very same article
and tadaaa
See it ? You cannot resteemit
PS "refreshing does not help

* Browser: all
* Operating system: all

th7nder:
Fixes #2081 

### Changes

* Made reblogs to be stored per user rather than per app

### Test plan

* [ ] Log in with account A
* [ ] Reblog a post
* [ ] Log out
* [ ] Log in with account B
* [ ] Reblog the same post

### DIFF
--- a/src/client/Wrapper.js
+++ b/src/client/Wrapper.js
@@ -140,10 +140,10 @@ export default class Wrapper extends React.PureComponent {
       this.props.getFollowing();
       this.props.getNotifications();
       this.props.busyLogin();
+      this.props.getRebloggedList();
     });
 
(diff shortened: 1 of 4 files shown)


Sekhmet (commented):
x

PR #2187, opened 2019-02-18
feat: remove like post from editor

Sekhmet:
### Changes

Summary of the changes you made.

* Remove "Like this post" from the editor.

### Test plan

* [ ] Go to editor and check that there is no Like this post option.
* [ ] Create a new post to check that post is not upvoted.


jm90m:
works for me tested on - https://busy-develop-pr-2187.herokuapp.com/@hellosteem/7vygkl-test

PR #2191, opened 2019-02-24
fix: skip comment options on edits

### LINKED ISSUE #2188: [Urgent] Cannot edit posts
This happens only on staging. But due to many new features and bug fixes, many people also use staging version.

#### Expected behavior

Users should be able to edit posts.

#### Actual behavior

Users cannot edit posts.

#### How to reproduce

Try to edit any post, then you'll see the following:

This is due to https://github.com/busyorg/busy/pull/2165 Without it, I can still edit.

Beneficiary setting is only possible when a post is created. No modification is allowed afterwards. But currently, by default, the beneficiary setting is on, so when editing, busy is trying to change the beneficiary setting, which is not allowed.

**For editing, beneficiary setting shouldn't be modified.**

Actually I think beneficiary for busy should be off by default if you guys don't mind :) Users will not like it if unintentionally set especially cause it's 10%. I also suggest it could be variable. 10% is too high especially when it's on by default.

* Browser: Any
* Operating system: Any

Sekhmet:
Fixes #2188

### DIFF
--- a/src/client/post/Write/editorActions.js
+++ b/src/client/post/Write/editorActions.js
@@ -89,6 +89,7 @@ const requiredFields = 'parentAuthor,parentPermlink,author,permlink,title,body,j
 
 const broadcastComment = (
   steemConnectAPI,
+  isUpdating,
   parentAuthor,
   parentPermlink,
   author,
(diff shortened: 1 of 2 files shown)


economicstudio:
@Sekhmet Thanks! Now it's working, but still the beneficiary setting is better to be unchecked on editing if the post didn't set beneficiary. Also, I suggest unchecked by default for new posts too.

PR #2206, opened 2019-06-04
fixed background color of highlighted search result in nightmode

mktcode:
### Changes

Just some minor CSS fix...

### Test plan

Search for an exact username. The username will be highlighted in the search results.

### Demo (optional)

Before:
busy-search-before

After:
busy-search-after


Sekhmet:
Thank you!

PR #2225, opened 2019-08-20
sc -> localstorage transition

economicstudio:
Update: Now everything's working

Users only need to login and reload once.

-----

mostly working but there are some known issues
- drafts / bookmarks can't be loaded initially. (maybe due to window instance problem.) but they're working once some modification(add, delete) is done.

settings are loaded well.

### DIFF
--- a/src/client/bookmarks/bookmarksReducer.js
+++ b/src/client/bookmarks/bookmarksReducer.js
@@ -10,11 +10,15 @@ const initialState = {
 const bookmarks = (state = initialState, action) => {
   switch (action.type) {
     case authActions.LOGIN_SUCCESS:
-      if (action.meta && action.meta.refresh) return state;
-      return {
-        ...state,
-        list: action.payload.user_metadata.bookmarks || initialState.list,
-      };
+      try {
+        return {
+          ...state,
+          list: JSON.parse(localStorage.getItem(USER_METADATA_KEY)).bookmarks || state.list,
+        };
+      } catch (error) {
+        // this is due to localStorage hasn't been ready. Can be ignored.
+        return state;
+      }
     case bookmarksActions.TOGGLE_BOOKMARK_START:
       return {
         ...state,
@@ -32,15 +36,7 @@ const bookmarks = (state = initialState, action) => {
         pendingBookmarks: state.pendingBookmarks.filter(bookmark => bookmark !== action.meta.id),
       };
     default:
-      try {
-        return {
-          ...state,
-          list: JSON.parse(localStorage.getItem(USER_METADATA_KEY)).bookmarks || state.list,
-        };
-      } catch (error) {
-        // this is due to localStorage hasn't been ready. Can be ignored.
-        return state;
-      }
+      return state;
   }
 };
 
--- a/src/client/bookmarks/bookmarksReducer.js
+++ b/src/client/bookmarks/bookmarksReducer.js
@@ -1,5 +1,6 @@
 import * as authActions from '../auth/authActions';
 import * as bookmarksActions from './bookmarksActions';
+import { USER_METADATA_KEY } from '../helpers/constants';
 
 const initialState = {
   list: {},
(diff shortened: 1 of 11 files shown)


src/client/helpers/metadata.js
```diff
@@ -1,75 +1,85 @@
 import omit from 'lodash/omit';
 import SteemConnect from '../steemConnectAPI';
 
-const getMetadata = () => SteemConnect.me().then(resp => resp.user_metadata);
+// sc2 metadata support discontinuation preparation
+// const getMetadata = () => SteemConnect.me().then(resp => resp.user_metadata);
+export const getMetadata = () =>
+  SteemConnect.me().then(resp => {
```
bonustrack: Request to steemconnect API should only be sent if there isn't localStorage with user metadata available.

src/client/helpers/metadata.js
```diff
@@ -1,75 +1,85 @@
...
+      localStorage.setItem('user_metadata', JSON.stringify(resp.user_metadata));
+      return resp.user_metadata;
+    } catch (err) {
+      return Promise.reject(new Error(err));
+    }
+  });
+
+export const updateUserMetadata = metadata => {
+  localStorage.setItem('user_metadata', JSON.stringify(metadata));
+  return Promise.resolve(metadata);
```
bonustrack, line 29: You dont need a promise anymore for this method

src/client/helpers/metadata.js
economicstudio: great point!

src/client/helpers/metadata.js
economicstudio: I know but one problem is all the other codes later on expect promise, so i had to choose this dirty trick rather than rewriting so many parts.

Thanks a lot for your review!

ps. Any idea on the known issues that I wrote?

src/client/bookmarks/bookmarksReducer.js
```diff
@@ -31,7 +32,15 @@ const bookmarks = (state = initialState, action) => {
         pendingBookmarks: state.pendingBookmarks.filter(bookmark => bookmark !== action.meta.id),
       };
     default:
-      return state;
+      try {
+        return {
+          ...state,
+          list: JSON.parse(localStorage.getItem(USER_METADATA_KEY)).bookmarks || state.list,
+        };
+      } catch (error) {
```
Sekhmet: Redux state shouldn't change without action.
If you want to load some state on startup from localStorage you need to do this:

1. Call `getUpdatedSCUserMetadata` action there: https://github.com/busyorg/busy/blob/27dd2383806eda8daf46748cbbbb26739d08ced4/src/client/Wrapper.js#L134-L145
2. Add new case there: `case types.UPDATE_SC2_USER_METADATA.SUCCESS:` that will restore the list using the payload from the action.
3. Add a similar case to `editorReducer` to restore drafts from the payload.

This way once webpage is loaded drafts and bookmarks will be restored from localStorage.

economicstudio: Hi @Sekhmet Thanks for your comment. I've actually already optimized that part (only load in LOGIN_SUCCESS case.) For somehow it didn't work that way last night, but it seems that i forgot to include parse :) so it was always null :( in both states. But now only loading in LOGIN_SUCCESS is enough.

As you know since the code will be refactored anyway, I implemented a way with a minimal code change. In this way, we don't need to add a new case.

Now all initial loading (bookmarks & drafts) is working. Could you review it again?

economicstudio: step 1 seems needed anyway (users need to do something to call getmetadata  otherwise), i'll do it, so wait for the review.

economicstudio: @Sekhmet and @bonustrack 

Now everything's working. Users need to reload just once for their initial login. Other than that, there's no problem.

Check it out and feel free to merge and disconnect sc metadata.

src/client/helpers/metadata.js
economicstudio: done!

Sekhmet (approved):
uACK.

